PR #1441, opened 2015-01-06
fixes CC-663

wwwtyro:
I was able to kick off serviced and deploy a service template while not connected to the internet. Already had all images locally.

### DIFF
--- a/cli/api/utils.go
+++ b/cli/api/utils.go
@@ -22,8 +22,8 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/zenoss/glog"
 	"github.com/control-center/serviced/utils"
+	"github.com/zenoss/glog"
 )
 
 const (
--- a/cli/cmd/cmd.go
+++ b/cli/cmd/cmd.go
@@ -95,7 +95,7 @@ func getLocalAgentIP() string {
 	ip := configEnv("OUTBOUND_IP", "")
 	if ip != "" {
 		ip = ip + ":4979"
-	}else{
+	} else {
 		ip = api.GetAgentIP()
 	}
 	return ip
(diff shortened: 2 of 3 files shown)


cli/api/utils.go
```diff
@@ -22,8 +22,8 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/zenoss/glog"
 	"github.com/control-center/serviced/utils"
+	"github.com/zenoss/glog"
```
wwwtyro: just linting

cli/cmd/cmd.go
```diff
@@ -95,7 +95,7 @@ func getLocalAgentIP() string {
 	ip := configEnv("OUTBOUND_IP", "")
 	if ip != "" {
 		ip = ip + ":4979"
-	}else{
+	} else {
```
wwwtyro: just formatting

commons/docker/api.go
```diff
@@ -100,7 +100,9 @@ func NewContainer(cd *ContainerDefinition, start bool, timeout time.Duration, on
 	if useRegistry {
 		if err := PullImage(iid.String()); err != nil {
 			glog.V(2).Infof("Unable to pull image %s: %v", iid.String(), err)
-			return nil, err
+			if _, err2 := lookupImage(iid.String()); err2 != nil {
+				return nil, err2
+			}
 		}
```
wwwtyro: Don't implode just because we can't pull; only implode if we can't pull AND we don't have the image locally.

wwwtyro:
Thanks JP - yeah, that could be a wrinkle. Perhaps the lack of internet connectivity will not prevent a connection to a local registry on the internal network? I'm testing multihost tomorrow; I'll report back with results.

wwwtyro:
I'm going to close this for now to prevent an unintentional merge. I'll reopen when it's ready.

wwwtyro:
Okay, tested on single and multihost and appears to work.

PR #1445, opened 2015-01-07
CC-650: SVC_USE now works correctly on mulithost environments

alapidas:
This was originally broken on multihost because SVC_USE would re-tag, but would not push to the registry (because the code was running on the client where SERVICED_REGISTRY was set to False, not the server where it is True). 

https://github.com/control-center/serviced/blob/develop/commons/docker/api.go#L563

This change makes the code do an RPC call to the server, so the tagging gets pushed into the registry.  This requires that the operation be physically performed on the master, but we already have this requirement due to snapshotting.

### DIFF
--- a/commons/docker/utils.go
+++ b/commons/docker/utils.go
@@ -23,6 +23,8 @@ import (
 // ServiceUse will tag a new image (imageName) in a given registry for a given tenant
 // to latest, making sure to push changes to the registry
 func ServiceUse(serviceID string, imageName string, registry string, noOp bool) (string, error) {
+	// If noOp is True, then replace the 'real' functions that talk to Docker with
+	// no-op functions (for dry run purposes)
 	pullImage := PullImage
 	findImage := FindImage
 	tagImage := TagImage
--- a/commons/docker/utils.go
+++ b/commons/docker/utils.go
@@ -34,6 +34,7 @@ func ServiceUse(serviceID string, imageName string, registry string, noOp bool) 
 		tagImage = noOpTagImage
 	}
 
+	// imageName is the new image to pull, eg. "zenoss/resmgr-unstable:1.2.3.4"
 	glog.V(0).Infof("preparing to use image: %s", imageName)
 	imageID, err := commons.ParseImageID(imageName)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


commons/docker/utils.go
```diff
@@ -0,0 +1,80 @@
...
+// to latest, making sure to push changes to the registry
+func ServiceUse(serviceID string, imageName string, registry string, noOp bool) (string, error) {
+	pullImage := PullImage
+	findImage := FindImage
+	tagImage := TagImage
+	if noOp {
+		pullImage = noOpPullImage
+		findImage = noOpFindImage
+		tagImage = noOpTagImage
+	}
```
smousa: need comment here...I don't understand this block

commons/docker/utils.go
```diff
@@ -0,0 +1,80 @@
...
+	return nil, nil
+}
+
+func noOpFindImage(repotag string, pull bool) (*Image, error) {
+	return nil, nil
+}
+
+func noOpPullImage(repotag string) error {
+	return nil
+}
```
smousa: Is noOp a test scenario?

alapidas: Yes

commons/imageid.go
```diff
@@ -62,6 +66,18 @@ func init() {
 	underscore, _ = utf8.DecodeRune([]byte("_"))
 }
 
+func RenameImageID(dockerRegistry, tenantId string, imgID string, tag string) (*ImageID, error) {
```
smousa: Need comment here: I don't understand what this function does

smousa: Need test cases

script/runner.go
```diff
@@ -123,6 +125,7 @@ func newRunner(config *Config, pctx *parseContext) *runner {
 		pullImage:       docker.PullImage,
 		execCommand:     defaultExec,
 		tagImage:        defaultTagImage,
```
jplouis: Can tag, pull and find image be removed?

alapidas: Yes, I forgot to do so.  Doing it now.

alapidas:
This should be good to go.

PR #1450, opened 2015-01-08
Feature/cc 577

tgjonestx:
See [CC-577](https://jira.zenoss.com/browse/CC-577) for changes to CLI usage.

**Questions:**

1. dfs/backup_test.go
The backup tests require sudo, which I assume we don't have in the CI build, so I introduced makefile changes so that 'make test' runs all of the other dfs unit-tests.  Anyone who wants to execute backup_test.go would have to do so manually (which is the case today). My solution is a smell to me, but I couldn't think of another alternative. 
Is there a better approach?

1. Where to put mocks?
I created several mocks to test changes in dfs/snapshot.go and I made all of those mocks package-private to dfs. In theory these mocks could be reused for tests in other packages. Should I move them to their respective 'parent' package? For instance, move dfs/mockDockerClient.go to commons/docker/MockDockerClient.go.
If I do, will that create other problems trying to build/run the tests because of package interdependencies, like you can't unit-test dfs without first unit-testing commons/docker?

1. Other calls that create snapshots
Besides the CLI, there are several other places in serviced where a snapshot is created. For the moment, I've specified an empty string as the description, but not sure what's the best approach for each of those cases:
  * dao/elasticsearch/filesystem.go::AsyncSnapshot()
    * Backing up through references in dao/interface.go and node/cp_client.go, it looks like it's unused to me. Should I remove it?  * dfs/backup.go::saveSnapshots()
  * dfs/docker.go::Commit()
  * scheduler/leader.go::TakeSnapshot()
  * script/eval.go::evalSnapshot()
  * web/resources.go::restSnapshotService()
    * is this called/documented anywhere?
    * if it's not used, should we remove it?

1. zzk Snapshots
I didn't change anything in zzk/snapshot because the compiler didn't complain. 
Does anything need to be changed here?

1. Missing unit tests from serviced/makefile
The toplevel makefile doesn't run unit-tests for the script and volume packages. The tests in script work on my box, but the tests in volume have atrophied to the point that they don't even build
Should I enable the unit-tests for script?

1. Side issue
On my box, I'm getting intermittent unit-test failures for common/docker.
Seems like it occassionally has problems connecting to the docker registry, but beyond that I stumped.
Just curious if this is some kind of known issue?


jplouis:
1)  Using sudo should be addressed as part of a larger change of separating integration tests from unit tests.  Also as part of test restructuring we should not need make files in every package to run unit tests.
2)  Leave them where they are for now.  Again should do a larger test restructure and put mocks in their own packages.
3) I don't know if snapshots in backups are visible or usable on their own.  As for the other uses of snapshots, at least script/eval.go::evalSnapshot() should be able to use a description if provided and default to something meaningful.
4) Ask Summer
5) Part of test restructuring
6) I think the build machines have seen this intermittent issue.

tgjonestx:
@smousa - do you have any comments to add to the questions in the PR? Particularly, 3 and 4

cli/cmd/service.go
```diff
@@ -1315,19 +1318,25 @@ func (c *ServicedCli) cmdServiceListSnapshots(ctx *cli.Context) {
...
-	if len(ctx.Args()) < 1 {
+	nArgs := len(ctx.Args())
+	if nArgs < 1 {
 		fmt.Printf("Incorrect Usage.\n\n")
 		cli.ShowCommandHelp(ctx, "snapshot")
 		return
 	}
 
+	description := ""
+	if nArgs <= 3 {
```
smousa: why are you checking this?

commons/docker/api.go
```diff
@@ -82,7 +82,7 @@ func NewContainer(cd *ContainerDefinition, start bool, timeout time.Duration, on
 	}{&cd.CreateContainerOptions, &cd.HostConfig, start, oncreate, onstart}
 
 	timeoutc := time.After(timeout)
-	dc, err := dockerclient.NewClient(dockerep)
```
smousa: Does this file belong here?

commons/docker/client.go
```diff
@@ -0,0 +1,167 @@
+// Copyright 2014 The Serviced Authors.
```
smousa: this file doesn't seem to make sense with the rest of the PR

commons/docker/kernel.go
```diff
@@ -113,7 +113,7 @@ func init() {
 		}
```
smousa: again, not related it seems?

dfs/volume.go
```diff
@@ -37,11 +37,15 @@ func (dfs *DistributedFilesystem) GetVolume(serviceID string) (*volume.Volume, e
 }
 
 // GetSubvolume gets the path of the *local* volume on the host
-func GetSubvolume(fsType, varpath, serviceID string) (*volume.Volume, error) {
+func GetSubvolume(fsType, varpath, serviceID string, mountVolume VolumeMounter) (volume.Volume, error) {
```
smousa: Instead of passing in a function, we should instead create a fsType called "mockfs" or something, I think.

cli/cmd/service.go
tgjonestx: I wanted to treat the new arguments for `--description "your description here"` as optional. In other words, don't call `ctx.String("description")` unless those arguments were added.
Maybe the check should have been `if nargs == 3`?

dfs/volume.go
tgjonestx: Hmm. I'm not sure how that would work from a unit-testing perspective. 
For example, see `TestSnapshot_Snapshot_VolumeNotFound()` in dfs/snapshot_test.go. I needed a way to return a mock instance of Volume so I that could simulate different kinds of error conditions in the unit-test.  To do that, I injected a the VolumeMounter function in the tests which returns a mock of volume.Volume.

PR #1452, opened 2015-01-12
Delete extra services on rollback/restore

jafcrocker:
Fixes https://jira.zenoss.com/browse/ZEN-15903
Enable code to delete services which are not part of the backup
Removed dead code

### DIFF
--- a/dfs/backup.go
+++ b/dfs/backup.go
@@ -532,8 +532,11 @@ func (dfs *DistributedFilesystem) loadSnapshots(tenantID, infile string) error {
 		glog.Errorf("Could not load services from %s: %s", label, err)
 		return err
 	}
+
 	// Restore the service data
-	if err := dfs.restoreServices(svcs); err != nil {
+	// This is necessary to restore services for tenants which do not exist - in that
+	//   case dfs.Rollback will fail if we don't restore first.
+	if err := dfs.restoreServices(tenantID, svcs); err != nil {
 		glog.Errorf("Could not restore services from %s: %s", label, err)
 		return err
 	}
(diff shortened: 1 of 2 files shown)


dfs/backup.go
```diff
@@ -532,8 +532,10 @@ func (dfs *DistributedFilesystem) loadSnapshots(tenantID, infile string) error {
 		glog.Errorf("Could not load services from %s: %s", label, err)
 		return err
 	}
+
 	// Restore the service data
```
smousa: We don't need this anymore, right?

dfs/backup.go
```diff
@@ -547,68 +549,6 @@ func (dfs *DistributedFilesystem) loadSnapshots(tenantID, infile string) error {
 	return nil
 }
 
-func (dfs *DistributedFilesystem) importSnapshots(filename string) error {
```
smousa: Why did you delete this function?

dfs/snapshot.go
```diff
@@ -299,7 +299,7 @@ func (dfs *DistributedFilesystem) DeleteSnapshots(tenantID string) error {
 	return nil
 }
 
-func (dfs *DistributedFilesystem) restoreServices(svcs []*service.Service) error {
+func (dfs *DistributedFilesystem) restoreServices(tenantID string, svcs []*service.Service) error {
```
smousa: Does it make sense to figure out the tenant ID from the list of passed-in services?

dfs/snapshot.go
```diff
@@ -365,28 +365,11 @@ func (dfs *DistributedFilesystem) restoreServices(svcs []*service.Service) error
 					glog.Errorf("Could not add service %s: %s", serviceID, err)
 					return err
 				}
+			}
 
-				/*
-					// TODO: enable this to generate a new service ID, instead of recycling
-					// the old one
```
smousa: Should we keep this TODO in case we decide we want to do this in the future?

smousa:
Don't forget to create another pull request into develop

dfs/backup.go
jafcrocker: It was not used and called dfs.Rollback.  I cleaned it up to help simplify my code analysis.

dfs/snapshot.go
jafcrocker: Possibly. given that we know what tenant we are trying to restore it made sense to me to use that information rather than try to infer it.  One could make a case for asserting that the tenant ID inferred from the snapshot matches the expected

PR #1461, opened 2015-01-14
fixes CC-702

wwwtyro:


alapidas:
is that formatting correct?

PR #1462, opened 2015-01-14
adds GOBUILD_FLAGS option to makefile

wwwtyro:
So if you want to build with race checking, for example, you can:

`control-center/serviced$ GOBUILD_FLAGS=-race make`


wwwtyro:
test this

PR #1472, opened 2015-01-15
More passive container cleanup, based on the current age of the oldest container

smousa:
https://jira.zenoss.com/browse/CC-716

### DIFF
--- a/node/agent.go
+++ b/node/agent.go
@@ -262,36 +262,41 @@ func (a *HostAgent) StopService(state *servicestate.ServiceState) error {
 	return ctr.Stop(45 * time.Second)
 }
 
-func reapContainers(maxAge time.Duration) error {
-	containers, err := docker.Containers()
-	if err != nil {
-		return err
-	}
-
-	cutoff := time.Now().Add(-maxAge)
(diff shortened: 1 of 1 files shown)


node/agent.go
```diff
@@ -262,36 +262,41 @@ func (a *HostAgent) StopService(state *servicestate.ServiceState) error {
...
+				// container is still running or hasn't started, skip
+				continue
+			} else if timeToLive := expiration.Sub(finishTime); timeToLive <= 0 {
+				// container has exceeded its expiration date
+				if err := ctr.Delete(true); err != nil {
+					glog.Errorf("Could not delete container %s (%s): %s", ctr.Name, ctr.ID, err)
+					waitTimeout = minInterval
+				}
+			} else if timeToLive < waitTimeout {
+				// set the expiration date of the next oldest container not purged
```
wwwtyro: Is this comment correct? Should it be "Set the time of next purge to the expiration date of oldest unpurged container"?

Also, does this handle the case where a container has been started since this logic was executed and its expiration date falls before timeToLive is hit?

PR #1475, opened 2015-01-16
Feature/controller rename and lint

rDr4g0n:


alapidas:
This code is making files that are owned by root, and the jenkins slave is unable to delete its workspace because of this.

web/package.json
```diff
@@ -1,10 +1,16 @@
...
+  "description": "Create, manage, and scale services in a uniform way",
+  "version": "1.0.0",
+  "repository": "https://github.com/control-center/serviced",
+  "private": true,
   "devDependencies": {
-    "grunt": "~0.4.4",
-    "grunt-contrib-uglify": "~0.4.0",
-    "grunt-contrib-concat": "~0.4.0",
-    "grunt-contrib-watch": "~0.6.1"
+    "gulp": "~3.8.10",
```
tgjonestx: There's a potential problem here - from one build to another, the tool chain used to build the app might change.  Even if we remove the tildes here, the way NPM resolves dependencies if one of these packages (or one of their downstream dependencies) uses permissive versioning, then the new packages might be downloaded, potentially introducing subtle changes in the build results.  I know of two tools that could address this. We should open another ticket to implement one of them (or a comparable solution).
https://www.npmjs.com/package/lockdown
https://docs.npmjs.com/cli/shrinkwrap

@jhanson - whatever solution we choose for CC should be used for RM also.

PR #1481, opened 2015-01-19
Feature/cc 729

tgjonestx:
First, cherry-pick fixes from develop for PR #1466 and then add additional locking for `restLogout()`

Also includes a fix for a deadlock in #1466 which can occur if no one logins for more than 1 minute after serviced starts. In that scenario, the previous impl of `purgeOldsessionTs()` would grab the lock and not release it

### DIFF
--- a/web/session.go
+++ b/web/session.go
@@ -66,11 +66,11 @@ func init() {
 }
 
 func purgeOldsessionTs() {
-	for {
-		time.Sleep(time.Second * 60)
-		sessionsLock.Lock()
+
+	// use a closure to facilitate safe locking regardless of when the purge function returns
+	doPurge := func() {
 		if len(sessions) == 0 {
-			continue
+			return
 		}
 		glog.V(1).Info("Searching for expired sessions")
 		cutoff := time.Now().UTC().Unix() - int64((30 * time.Minute).Seconds())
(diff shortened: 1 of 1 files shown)


web/session.go
```diff
@@ -83,6 +85,14 @@ func purgeOldsessionTs() {
 			delete(sessions, key)
 		}
 	}
+
+	for {
+		time.Sleep(time.Second * 60)
+
+		sessionsLock.Lock()
```
iancmcc: I would suggest you put the Lock() and then a defer Unlock() inside doPurge itself.

web/session.go
```diff
@@ -112,7 +125,9 @@ func restLogout(w *rest.ResponseWriter, r *rest.Request) {
 	if err != nil {
 		glog.V(1).Info("Unable to read session cookie")
 	} else {
+		sessionsLock.Lock()
```
iancmcc: Ditto. Wrap this in a function so you can use defer.

iancmcc: Or actually just use defer.

web/session.go
tgjonestx: I found a bug in the original impl where it failed to Unlock if there were 0 sessions. So if you start, wait > 1 minute, then you can't login. So I could have fixed that bug with an Unlock call before the continue, but I thought that left the routine vulnerable to some future bug if another continue was added.  I can move the Lock/Unlock into the closure if you think that's better.

iancmcc: Yeah, I'd do something like:

if len(sessions) == 0 {
    return
}
sessionsLock.Lock()
defer sessionsLock.Unlock()

PR #1495, opened 2015-01-21
Change rpm reversion code to use new packaging

alapidas:


pkg/reversion/rpm/reversion.sh
```diff
@@ -1,33 +1,37 @@
 #!/bin/sh
 
-if [ "$#" -ne 3 ]
+if [ "$#" -ne 5 ]
 then
 	echo "USAGE: $0 sourceRepoClass sourceVersion targetVersion"
```
tgjonestx: The usage and examples still have just 3 arguments instead of 5.

PR #1498, opened 2015-01-21
CC-740: Halt Ready in zzk if base path does not exist

smousa:
https://jira.zenoss.com/browse/CC-740


wplu:
SHIP!   waiting for build test

PR #1535, opened 2015-01-27
Run 'npm install' in the serviced-build container to precache NPM packag...

tgjonestx:
@jplouis - the intent is to speed up builds for developers who haven't taken extra steps to install node. I can explain the problem and solution better in person than in prose.


tgjonestx:
This PR is a deadend

PR #1537, opened 2015-01-28
facade call to findChildService requires an existing parent service, whi...

smousa:
...le Add/Update uses the DeploymentID

https://jira.zenoss.com/browse/CC-763

### DIFF
--- a/facade/service.go
+++ b/facade/service.go
@@ -43,7 +43,7 @@ func (f *Facade) AddService(ctx datastore.Context, svc service.Service) error {
 	}
 
 	// verify the service with parent ID does not exist with the given name
-	if s, err := store.FindChildService(ctx, svc.ParentServiceID, svc.Name); err != nil {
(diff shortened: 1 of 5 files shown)


facade/service.go
```diff
@@ -324,7 +324,16 @@ func (f *Facade) GetServiceEndpoints(ctx datastore.Context, serviceId string) (m
 func (f *Facade) FindChildService(ctx datastore.Context, serviceId string, childName string) (*service.Service, error) {
 	glog.V(3).Infof("Facade.FindChildService")
 	store := f.serviceStore
-	return store.FindChildService(ctx, serviceId, childName)
+	parentService, err := store.Get(ctx, serviceId)
```
kindkid: Is it the parentService or the topLevelService?

smousa: parentService, otherwise it would be called the tenant

kindkid: is parentService the parent of the service with ID == serviceId? Or is it the exact service with ID == serviceId?

wwwtyro:
test this

PR #1546, opened 2015-01-29
fixes CC-757: seed dotfile to indicate initialized service volume

wplu:
https://jira.zenoss.com/browse/CC-757

I would prefer the 2 optimizations here: https://github.com/control-center/serviced/pull/1547

### DIFF
--- a/node/utils.go
+++ b/node/utils.go
@@ -29,6 +29,7 @@ import (
 	"os"
 	"os/exec"
 	"os/user"
+	"path"
 	"path/filepath"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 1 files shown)


node/utils.go
```diff
@@ -310,13 +311,22 @@ func createVolumeDir(hostPath, containerSpec, imageSpec, userSpec, permissionSpe
 	createVolumeDirMutex.Lock()
```
jplouis: This whole method will be serial because of the mutex. On initial startup it will only run one at a time w/ at least the 5 second sleep later on.  Can this lock be more granular on say the something related to the hostPath?

wplu: i suppose we could create a lock per hostpath, but that seems somewhat risky this close to release.  With the other two optimizations that are less risky, we have already reduced all service container startup times from 10 minutes down to 1m30s.

jplouis: How fast is initial startup?

wplu: initial startup is now 1 minute 30 for initial start of zenoss.resmgr on AWS.  Mariadb averages 30 seconds  :(

wplu: there are various initial startup times in the comments in the ticket

wplu: sorry, 1m30s is for this https://github.com/control-center/serviced/pull/1547, not https://github.com/control-center/serviced/pull/1546

jplouis: Awesome fix

wplu: thank you - i was also thinking along your lines - hostpath based map of locks - but deferred to the simpler/easier solutions

wplu: it's still very tempting to try the hostpath based map of locks since we'll probably get to under 45 seconds of initial startup for everything in a mostly safe manner

PR #1552, opened 2015-01-30
Change Version number to reflect CR3

jhuie-zenoss:


wplu:
that's a good very short term

wplu:
we should make a runtime call to call rpm or dpkg to get the package version

PR #1555, opened 2015-01-30
fix bad path

rDr4g0n:


wplu:
ship - once build test finishes

PR #1556, opened 2015-01-30
CC-770: Delete services hierarchically during rollback/restore

smousa:
changed order of deleting services during rollback and restore
https://jira.zenoss.com/browse/CC-770


jafcrocker:
Looks good to me...

PR #1562, opened 2015-02-02
Feature/angular reorg

rDr4g0n:
- Move angularjs source files around (group by feature, not by type). IE: all service related stuff (angular controllers, html, etc) are in src/Services
- Delete old angular library
- Delete angular docs directory as it's way too big
- update gulpfile, karm.conf, etc to use new paths


tgjonestx:
:+1:

PR #1569, opened 2015-02-03
Feature/cc 544

morr-zenoss:
Partial fix for CC-544 - serviced was not cleaning up its registry of processes, so when it posted metrics it posted the last metric it had for every process, whether the process was alive or not. Now, we're cleaning up the list, so metrics are only posted for current processes.


alapidas:
Looks good to me, are we waiting to merge this for some reason?

PR #1576, opened 2015-02-04
fixes CC-801

wwwtyro:


wplu:
ship when build test passes  :)

PR #1578, opened 2015-02-05
Feature/hosts factory

rDr4g0n:
Step 2 of big CC UI plumbing refactor. Makes all angular controllers and providers use hostsFactory for any host related requests.


web/ui/src/DeployWiz/DeployWizard.js
```diff
@@ -253,18 +254,6 @@
...
-        };
-
-        $scope.addHostFinish = function() {
-            $scope.newHost.Name = $scope.newHost.IPAddr;
-            $scope.newHost.ID = 'fakefakefake';
-            $scope.newHost.selected = true;
-            $scope.detected_hosts.push($scope.newHost);
-            $scope.step_page = $scope.steps[step].content;
-        };
-
```
rDr4g0n: these functions are no longer used/referenced

web/ui/src/Hosts/hostsFactory.js
```diff
@@ -0,0 +1,136 @@
...
+            this.name = host.Name;
+            this.id = host.ID;
+
+            // TODO - determine full pool path from PoolID
+            this.fullPath = "";
+
+            this.host = Object.freeze(host);
+        },
+
+        getInstances: function(){
```
tgjonestx: Seems like the method name should say something about services to clarify what kind of instance is being retrieved. Maybe a better name would be `getActiveServices()`

tgjonestx:
How about some unit tests for the controllers? At least for HostDetails and/or HostMap?

rDr4g0n:
I think ill hold off on the writing tests as I suspect I will be making another pass at the service, host, pool, and serviceInstance types to further distill them into a base prototype.

PR #1585, opened 2015-02-10
speed up elastic tests by using a single elastic server

wwwtyro:


tgjonestx:
Do we already have setup/teardown for each test case and/or test suite which uses ES to make sure that ES data from one test case/suite doesn't interfere with another test case/suite?

makefile
```diff
@@ -435,9 +435,19 @@ export GOTEST_FLAGS:=$(GOTEST_FLAGS) -race
...
+ES_URL=https://download.elasticsearch.org/elasticsearch/elasticsearch/elasticsearch-$(ES_VER).tar.gz
+ES_TMP:=$(shell mktemp -d)
+ES_DIR=$(ES_TMP)/elasticsearch-$(ES_VER)
+
 .PHONY: test
 test: build docker_ok
-	go test -p 1 ./... $(GOTEST_FLAGS)
+	cd $(ES_TMP) && curl -s $(ES_URL) | tar -xz
+	echo "cluster.name: zero" > $(ES_DIR)/config/elasticsearch.yml
+	$(ES_DIR)/bin/elasticsearch -f -Des.http.port=9202 > $(ES_TMP)/elastic.log & echo $$!>$(ES_TMP)/elastic.pid
```
tgjonestx: Can 2 serviced builds execute on the same Jenkins agent at the same time? If so, we should parameterize the ES port number so that different Jenkins jobs can choose different ports to avoid collision.

jplouis:
@tgjonestx Yes, the setup deletes the index being used for the test.  https://github.com/control-center/serviced/blob/develop/datastore/elastic/testutils.go#L120

wwwtyro:
test this

makefile
wwwtyro: Only one runs on a slave at a time, so we're safe there.

PR #1588, opened 2015-02-11
Better management of external elastic process during testing.

wwwtyro:


wwwtyro:
test this

wwwtyro:
test this

wwwtyro:
test this

isvcs/es.go
```diff
@@ -121,7 +121,6 @@ func elasticsearchHealthCheck(port int) func() error {
 					return err
 				} else {
 					http.Post(baseUrl+"/controlplane", "application/json", buffer)
-					buffer.Close()
 				}
```
wwwtyro: Post will close buffer. Might solve a race we're seeing.

makefile
```diff
@@ -423,31 +423,10 @@ docker_buildandpackage: docker_ok
...
-ifeq (,$(findstring history_size,$(GORACE)))
-export GORACE:=$(GORACE) history_size=7
-endif
-ifeq (,$(findstring halt_on_error,$(GORACE)))
-export GORACE:=$(GORACE) halt_on_error=1
-endif
-ifeq (,$(findstring -race,$(GOTEST_FLAGS)))
-export GOTEST_FLAGS:=$(GOTEST_FLAGS) -race
-endif
-endif
```
tgjonestx: I suggest keeping these variables, and changing run-test.sh to something like
```
go test ${GOTEST_FLAGS} -p 1 ./...
```
The idea is two-fold:
1. The make user can run `make NORACETEST=1` to execute the tests without race checks
2. The make user can run `make GO_TEST_FLAGS="my custom flags"` to pass additional flags to go test.

run-tests.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+ES_VER=0.90.13
+ES_TMP=/tmp/serviced_elastic
+ES_DIR=$ES_TMP/elasticsearch-$ES_VER
+
+if [ -e $ES_TMP/pid ]; then \
```
tgjonestx: The `\` at the end of lines 7, 8 are not necessary

run-tests.sh
```diff
@@ -0,0 +1,29 @@
...
+	kill `cat $ES_TMP/pid`; \
+fi
+rm -rf $ES_TMP
+
+mkdir $ES_TMP
+if [ ! -e /tmp/elasticsearch-$ES_VER.tar.gz ]; then \
```
tgjonestx: Ditto last comment for lines 13 & 14

run-tests.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+ES_VER=0.90.13
+ES_TMP=/tmp/serviced_elastic
+ES_DIR=$ES_TMP/elasticsearch-$ES_VER
+
+if [ -e $ES_TMP/pid ]; then \
+	kill `cat $ES_TMP/pid`; \
+fi
+rm -rf $ES_TMP
```
tgjonestx: lines 7-10 can be wrapped into a function like `stop_elasticsearch`, then you can call that function before/after starting ES instead of repeating the code at lines 24-27

tgjonestx:
FYI - I did a local checkout, build and exec with the latest changes (ff7ab8bb81f718661dc23757323be9e0331acd89), but I couldn't reproduce the error in build [3272](http://jenkins.zendev.org/view/Control%20Center/job/serviced-pullrequests/3272/) :-(

I ran the following: 
```
make clean test DOCKERCFG=""
make
make smoketest DOCKERCFG=""
```

zzk/service/lock_test.go
```diff
@@ -118,7 +118,24 @@ func (t *ZZKTest) TestEnsureServiceLock(c *C) {
 
 	// Release lock
 	close(exit)
-	runtime.Gosched()
+
```
tgjonestx: Sanity check - Is this necessary?

Just asking because it seems unrelated to the other changes.

makefile
wwwtyro: I was thinking that, but since the user can as easily run go test ./... with whatever flags they want, I leaned towards making this script simpler.

zzk/service/lock_test.go
wwwtyro: The code below is supposed to be a more robust replacement for that. We should probably get John Crocker to take a look at it and make sure it's okay; it's his test.

run-tests.sh
wwwtyro: Agreed; will do!

makefile
tgjonestx: Keeping the flags in make will keep the test execution and the build execution more consistent, plus it allows someone to experiment with changing the default options in way that's consistent with how Jenkins would run the build.

wwwtyro:
test this

run-tests.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+ES_VER=0.90.13
+ES_TMP=/tmp/serviced_elastic
+ES_DIR=$ES_TMP/elasticsearch-$ES_VER
+
+if [[ -z $GORACE ]]; then
+	export GORACE="history_size=7,halt_on_error=1";
```
tgjonestx: Does comma separator work?  According to the [doc](https://golang.org/doc/articles/race_detector.html#Options), separate options are supposed to be separated by spaces, which is what the makefile in an earlier revision

zzk/service/lock_test.go
tgjonestx: @jafcrocker - can you look at these changes to `zzk/service/lock_test.go`?

tgjonestx:
:+1:  Will merge as soon as build is green.

wwwtyro:
test this

wwwtyro:
test this

PR #1591, opened 2015-02-12
Feature/govet

tgjonestx:
Please review aa746d2 and cc77548 very carefully

### DIFF
--- a/container/controller.go
+++ b/container/controller.go
@@ -670,6 +670,8 @@ func (c *Controller) checkPrereqs(prereqsPassed chan bool, rpcDead chan struct{}
 		return nil
 	}
 	healthCheckInterval := time.Tick(1 * time.Second)
+
+WaitingForPrereqs:
 	for {
 		select {
 		case <-rpcDead:
@@ -693,7 +695,7 @@ func (c *Controller) checkPrereqs(prereqsPassed chan bool, rpcDead chan struct{}
 			if !failedAny {
 				glog.Infof("Passed all prereqs.")
 				prereqsPassed <- true
-				return nil
+				break WaitingForPrereqs
 			}
 		}
 	}
--- a/container/controller.go
+++ b/container/controller.go
@@ -539,7 +539,7 @@ func (c *Controller) Run() (err error) {
 
 	rpcDead, err := c.rpcHealthCheck()
 	if err != nil {
-		glog.Error("Could not setup RPC ping check: %s", err)
+		glog.Errorf("Could not setup RPC ping check: %s", err)
 		return err
 	}
 
--- a/container/controller.go
+++ b/container/controller.go
@@ -670,8 +670,6 @@ func (c *Controller) checkPrereqs(prereqsPassed chan bool, rpcDead chan struct{}
 		return nil
 	}
 	healthCheckInterval := time.Tick(1 * time.Second)
-
-WaitingForPrereqs:
 	for {
 		select {
 		case <-rpcDead:
@@ -695,7 +693,7 @@ WaitingForPrereqs:
 			if !failedAny {
 				glog.Infof("Passed all prereqs.")
 				prereqsPassed <- true
-				break WaitingForPrereqs
+				return nil
 			}
 		}
 	}
--- a/container/endpoint.go
+++ b/container/endpoint.go
@@ -101,16 +101,23 @@ func getServiceState(conn coordclient.Connection, serviceID, instanceIDStr strin
 	}
 	instanceID := int(tmpID)
 
+	var serviceState *servicestate.ServiceState
+	err = fmt.Errorf("unable to retrieve service state")
+
+LookingForServiceState:
 	for {
-		serviceStates, err := getServiceStates(conn, serviceID)
-		if err != nil {
-			glog.Errorf("Unable to retrieve running service (%s) states: %v", serviceID, err)
-			return nil, fmt.Errorf("endpoint.go getServiceState zzk.GetServiceStates failed: %v", err)
+		serviceStates, e := getServiceStates(conn, serviceID)
+		if e != nil {
+			glog.Errorf("Unable to retrieve running service (%s) states: %v", serviceID, e)
+			err = fmt.Errorf("endpoint.go getServiceState zzk.GetServiceStates failed: %v", e)
+			break LookingForServiceState
 		}
 
 		for ii, ss := range serviceStates {
 			if ss.InstanceID == instanceID && ss.PrivateIP != "" {
-				return &serviceStates[ii], nil
+				serviceState = &serviceStates[ii]
+				err = nil
+				break LookingForServiceState
 			}
 		}
 
@@ -118,7 +125,7 @@ func getServiceState(conn coordclient.Connection, serviceID, instanceIDStr strin
 		time.Sleep(1 * time.Second)
 	}
 
-	return nil, fmt.Errorf("unable to retrieve service state")
+	return serviceState, err
 }
 
 // getEndpoints builds exportedEndpoints and importedEndpoints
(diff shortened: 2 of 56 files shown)


tgjonestx:
please test this

tgjonestx:
retest this

tgjonestx:
retest this

tgjonestx:
retest this

container/controller.go
```diff
@@ -693,7 +695,7 @@ func (c *Controller) checkPrereqs(prereqsPassed chan bool, rpcDead chan struct{}
 			if !failedAny {
 				glog.Infof("Passed all prereqs.")
 				prereqsPassed <- true
-				return nil
+				break WaitingForPrereqs
```
wwwtyro: Our initial design for the prereqs is that they are only checked the first time we kick off the service, so this change would need some discussion.

wwwtyro: Wait - is the purpose of this to break out of the for? Hmm. I guess I'm not a fan of this style - if I were reading it, it would take me longer to parse the break than the return.

container/endpoint.go
```diff
@@ -101,16 +101,23 @@ func getServiceState(conn coordclient.Connection, serviceID, instanceIDStr strin
...
+			err = fmt.Errorf("endpoint.go getServiceState zzk.GetServiceStates failed: %v", e)
+			break LookingForServiceState
 		}
 
 		for ii, ss := range serviceStates {
 			if ss.InstanceID == instanceID && ss.PrivateIP != "" {
-				return &serviceStates[ii], nil
+				serviceState = &serviceStates[ii]
+				err = nil
+				break LookingForServiceState
```
wwwtyro: The previous form reads easier to me.

domain/host/host.go
```diff
@@ -156,8 +156,6 @@ func Build(ip string, rpcport string, poolid string, ipAddrs ...string) (*Host,
 	host.ServiceD.Date = servicedversion.Date
 	host.ServiceD.Buildtag = servicedversion.Buildtag
 
-	*host = *host
-
```
wwwtyro: What in the world? I'd like to double check with someone to make sure nothing fancy is going on here before we take it out.

facade/service.go
```diff
@@ -225,9 +225,9 @@ func (f *Facade) GetServices(ctx datastore.Context, request dao.EntityRequest) (
 
 		return services, nil
 	default:
+		services = nil
 		err := fmt.Errorf("Bad request type %v: %+v", v, request)
 		glog.V(2).Info("Facade.GetTaggedServices: err=", err)
-		return nil, err
```
wwwtyro: Can you explain this change?

makefile
```diff
@@ -155,6 +155,11 @@ build_js:
 $(GOSRC)/$(godep_SRC):
 	go get $(godep_SRC)
 
+#
+# FIXME: drop -composites=false to get full coverage
+govet:
```
wwwtyro: Can I vote for "make vet"? :D

utils/docker.go
```diff
@@ -45,15 +46,13 @@ func DockerLogin(username, password, email string) (string, error) {
...
+		status, err = registry.Login(&authconfig, registry.HTTPRequestFactory(nil))
 		if err != nil {
 			return "", err
 		}
-		return status, nil
 	}
 
-	return "", fmt.Errorf("Auth params don't make sense.")
-
+	return status, nil
```
wwwtyro: So, this code needs some fixing - that last return of mine is pretty laughable. But what about this? (winging it, this might be broken)

```
func DockerLogin(username, password, email string) (string, error) {
	var status string
	var err error

	if username == "" && password == "" && email == "" {
		// Attempt login with .dockercfg file.
		configFile, err := registry.LoadConfig(os.Getenv("HOME"))
		if err != nil {
			return "", err
		}
		authconfig, ok := configFile.Configs[registry.IndexServerAddress()]
		if !ok {
			return "", fmt.Errorf("Error: Unable to login, no data for index server.")
		}
		status, err = registry.Login(&authconfig, registry.HTTPRequestFactory(nil))
		if err != nil {
			return "", err
		}
		return status, nil
	} 
	// Attempt login with this function's auth params.
	authconfig := registry.AuthConfig{
		Username:      username,
		Email:         email,
		Password:      password,
		ServerAddress: registry.IndexServerAddress(),
	}
	status, err = registry.Login(&authconfig, registry.HTTPRequestFactory(nil))
	if err != nil {
		return "", err
	}
	return status, nil
}
```

Then if we're looking at a traceback, it's more obvious which path was taken. I'm very open to discussion.

web/cpserver.go
```diff
@@ -381,7 +382,7 @@ func (sc *ServiceConfig) syncAllVhosts(shutdown <-chan interface{}) error {
 		select {
 		case <-shutdown:
 			close(cancelChan)
-			return nil
+			break WatchingChildren
```
wwwtyro: Same opinion here. Is go vet recommending this? If so, I'm gonna cry. :D

zzk/registry/registry.go
```diff
@@ -191,7 +190,7 @@ func (r *registryType) ensureKey(conn client.Connection, key string) error {
 		default:
 		}
 	}
-	return fmt.Errorf("could not create key: %s", key, err)
+	return nil
```
wwwtyro: I'm not sure this works - looks like we want to return an error if the timeout is reached.

zzk/registry/registry.go
```diff
@@ -208,7 +207,7 @@ func (r *registryType) ensureDir(conn client.Connection, path string) error {
 		default:
 		}
 	}
-	return fmt.Errorf("could not create dir: %s", path, err)
+	return nil
```
wwwtyro: Same.

zzk/registry/registry.go
```diff
@@ -253,7 +254,7 @@ func watch(conn client.Connection, path string, cancel <-chan bool, processChild
 			glog.V(1).Infof("watch event %+v at path: %s", ev, path)
 		case <-cancel:
 			glog.V(1).Infof("watch cancel at path: %s", path)
-			return nil
+			break WatchingChildren
```
wwwtyro: Same opinion.

zzk/registry/registry.go
```diff
@@ -282,7 +285,7 @@ func (r *registryType) watchItem(conn client.Connection, path string, nodeType c
 		case ev := <-event:
 			glog.V(2).Infof("watch event %+v at path: %s", ev, path)
 		case <-cancel:
-			return nil
+			break WatchingItem
```
wwwtyro: Same opinion.

zzk/snapshot/snapshot_test.go
```diff
@@ -104,9 +104,9 @@ func (t *ZZKTest) TestSnapshotListener_Listen(c *C) {
 	c.Log("Sending failure snapshot")
 	serviceID = "service-id-failure"
 	if nodeID, err := Send(conn, serviceID); err != nil {
-		c.Errorf("Could not send failure snapshot: ", err)
+		c.Errorf("Could not send failure snapshot: %s", err)
 	} else if err := Recv(conn, nodeID, &actual); err != nil {
-		c.Errorf("Could not receive failure snapshot: ", err)
+		c.Errorf("Could not receive failure snapshot: %s", err)
```
wwwtyro: Hat's off for plowing through this. I'm super excited to get go vet going. (Get it? GOing? I kill me.) Let's discuss the break label pattern you've been using on Tuesday. My main concerns are that it seems to make things a little more complex to read and debug, with no clear benefit (apparent to me). If you teach me something and convince me otherwise, I will mergify all the things. :)

tgjonestx:
@wwwtyro - *bump*

PR #1596, opened 2015-02-17
merge snapshot ttl from release

wwwtyro:


wwwtyro:
Self-merging port from release.

PR #1602, opened 2015-02-23
Feature/factories clean

rDr4g0n:


web/ui/src/Hosts/HostsController.js
```diff
@@ -129,6 +129,8 @@
         // update hosts
         update();
 
+        hostsFactory.activate();
```
tgjonestx: When do we deactivate the hostsFactory?  Does angular have a hook for leaving a page so we can deactivate a factory when the corresponding view is no longer active?

web/ui/src/common/baseFactory.js
```diff
@@ -0,0 +1,107 @@
...
+                            }
+                        }
+                    }
+
+                    deferred.resolve();
+                });
+            return deferred.promise;
+        },
+
+        activate: function(){
```
tgjonestx: Can this be simplified to something like this just to avoid stopping/starting a new timer when the promise is basically the same each time?
```
activate: function() {
    if (!this.updatePromise) {
        this.updatePromise = $interval(this.update.bind(this), UPDATE_FREQUENCY);
   }
}
```

web/ui/src/Instances/instancesFactory.js
```diff
@@ -3,106 +3,46 @@
...
+        newFactory.instanceMap = newFactory.objMap;
 
-            getByServiceId: (id) => {
+        angular.extend(newFactory, {
+            getByServiceId: function(id){
                 let results = [];
-                for(let i in instanceMap){
-                    if(instanceMap[i].model.ServiceID === id){
-                        results.push(instanceMap[i]);
+                for(let i in this.objMap){
```
tgjonestx: Why replace `instanceMap` with `objMap`? 

Within a given file, I think it would be better to use the same naming convention.  Also, if I look for `instanceMap`, the only reference I find is line 21, which seems a little crufty

web/ui/src/Services/servicesFactory.js
```diff
@@ -5,211 +5,192 @@
...
+                            });
 
-                            // new
-                            } else {
-                                serviceMap[serviceDef.ID] = new Service(serviceDef);
-                                addServiceToTree(serviceMap[serviceDef.ID]);
+                            // generate service tree
+                            for(var serviceId in this.objMap){
+                                // TODO - check for service
+                                service = this.objMap[serviceId];
```
tgjonestx: Ditto previous comment about replacing `objMap` with `serviceMap`

web/ui/src/Hosts/HostDetailsController.js
```diff
@@ -79,6 +79,9 @@
         update();
 
         function update(){
+            // start polling
+            hostsFactory.activate();
```
tgjonestx: Needs to be deactivated on when the modal closes.

web/ui/src/Pools/PoolDetailsController.js
```diff
@@ -110,6 +110,9 @@
             $location.path('/hosts/' + hostId);
         };
 
+        // start polling
+        poolsFactory.activate();
```
tgjonestx: Needs to be deactivated at some point

web/ui/src/Pools/PoolDetailsController.js
```diff
@@ -117,14 +120,19 @@
                 if ($scope.currentPool) {
                     $scope.breadcrumbs.push({label: $scope.currentPool.id, itemClass: 'active'});
 
+                    // start polling
+                    hostsFactory.activate();
```
tgjonestx: needs to be deactivated at some point

web/ui/src/Pools/PoolsController.js
```diff
@@ -105,6 +105,9 @@
                 });
         };
 
+        // start polling
+        poolsFactory.activate();
```
tgjonestx: needs to be deactivated

web/ui/src/Services/ServiceDetailsController.js
```diff
@@ -600,15 +600,10 @@
             }, $scope.update);
         });
 
+        // start polling
+        hostsFactory.activate();
```
tgjonestx: needs to be deactivated

tgjonestx:
retest this

PR #1608, opened 2015-02-24
Feature/deployed apps clean

rDr4g0n:


tgjonestx:
retest this

web/ui/src/common/miscUtils.js
```diff
@@ -128,12 +131,43 @@
...
+
             // call fn b after fn a
             after: function(a, b, context){
                 return function(){
                     a.apply(context, arguments);
                     b.call(context);
                 };
+            },
+
+            mapToArr: function(map){
```
tgjonestx: Maybe use http://api.jquery.com/jquery.makearray/ instead?

web/ui/src/common/miscUtils.js
```diff
@@ -128,12 +131,43 @@
...
+                for(let key in map){
+                    arr.push(map[key]);
+                }
+                return arr;
+            },
+
+            // cache function results based on hash function.
+            // NOTE: unlike regular memoize, the caching is entirely
+            // based on hash function, not on arguments
+            memoize: function(fn, hash){
```
tgjonestx: unit test?

web/ui/src/common/miscUtils.js
```diff
@@ -128,12 +131,43 @@
                 $translate.use(ln);
             },
 
+            capitalizeFirst: function(str){
```
tgjonestx: unit-test?

web/ui/src/DeployWiz/DeployWizard.js
```diff
@@ -328,61 +283,46 @@
...
-                                }
+                                $status.html('<strong>' + $translate.instant(parts[0]) + '</strong>');
                             }
+                        }
 
-                            getStatus();
-                        });
-                    }
-                };
+                        getStatus();
```
tgjonestx: I don't see any kind of delay or timeout between polls here. Am I missing something?

Also kind of confusing - get_deployed_templates checks on the deployment status, but this callback doesn't clear `checkStatus`. When does the polling stop?

web/ui/src/Services/AppsController.js
```diff
@@ -6,78 +6,41 @@
...
-
-        $scope.click_app = function(id) {
+        // redirect to specific service details
+        $scope.routeToService = function(id) {
             $location.path('/services/' + id);
         };
 
-        $scope.click_pool = function(id) {
+        // redirect to specific pool
+        $scope.routeToPool = function(id) {
```
tgjonestx: There are several places where the code builds URLs, and it's all kind of scattered around - some in resourcesFactory, some in the resource-specific file like PoolsController and some in other places like here and HostController.

Another good cleanup item (maybe another PR?) would to consolidate all of methods like `routeToPool` in one place resourcesFactory perhaps?   It's unlikely to happen for 1.1, but we've talked about reorganizing the backend URL namespace or standing up a separate node server. When that happens, it will be easier to update the UI code if the URL manipulation code is in fewer places

web/ui/src/Services/AppsController.js
```diff
@@ -229,40 +180,126 @@
...
+                }
+                lastPollResults = $scope.deployingServices.length;
             });
         }
 
-        // Get a list of templates
-        refreshTemplates();
+        function removeService(service) {
+            resourcesFactory.remove_service(service.id, function(){
+                // TODO - once the backend updates deleted
```
tgjonestx: This callback is good example of some unit-testable code within a controller.

web/ui/src/Services/AppsController.js
```diff
@@ -229,40 +180,126 @@
...
+        }
 
-            // if only isvcs are deployed, and this is the first time
-            // running deploy wizard, show the deploy apps modal
-            if(!$cookies.autoRunWizardHasRun && $scope.services.data.length === 1){
-                $scope.modalAddApp();
+        // init stuff for this controller
+        function init(){
+            // check is location.hostname is an IP
+            var re = /\b(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\b/;
```
tgjonestx: This same check is ServiceDetailsController.js - both controllers should use the same implementation somehow.
Maybe a new method on resourcesFactory?

web/ui/src/Services/view-apps.html
```diff
@@ -26,87 +27,98 @@ <h3 class="pull-left" translate>applications</h3>
...
         <thead>
-        <tr>
-            <th ng-repeat="header in templates.headers">
-                <span translate>{{header.name}}</span>
-                <button ng-click="templates.set_order(header.id, templates)"
-                        ng-class="templates.get_order_class(header.id, templates)"></button>
-            </th>
-            <th translate>running_tbl_actions</th>
-        </tr>
+            <tr>
```
tgjonestx: nit - there are tabs instead of spaces here

PR #1613, opened 2015-02-26
Feature/cli migration

tgjonestx:
An initial implement of service migration. Still very simple at this stage. This PR just addresses the first two tasks of the Rally story [\[Service Migration\] Invoke migration script](https://rally1.rallydev.com/#/30873679006d/detail/userstory/31069383245)

Please look at 8226df1 very closely. I refactored out a function to validate a service definition so that 'migrate --dry-run' shares the same validation as a regular add or edit operation, but I had to rearrange a few things that could break add/edit if a I made a mistake.

### DIFF
--- a/cli/cmd/service.go
+++ b/cli/cmd/service.go
@@ -91,6 +91,9 @@ func (c *ServicedCli) initService() {
 				Description:  "serviced service migrate SERVICEID PATH_TO_SCRIPT",
 				BashComplete: c.printServicesAll,
 				Action:       c.cmdServiceMigrate,
+				Flags: []cli.Flag{
+					cli.BoolFlag{"dry-run", "Executes the migration and validation without updateing anything"},
+				},
 			}, {
 				Name:         "remove",
 				ShortName:    "rm",
@@ -713,7 +716,7 @@ func (c *ServicedCli) cmdServiceMigrate(ctx *cli.Context) {
 		input = os.Stdin
 	}
 
-	if migratedSvc, err := c.driver.MigrateService(svc.ID, input); err != nil {
+	if migratedSvc, err := c.driver.MigrateService(svc.ID, input, ctx.Bool("dry-run")); err != nil {
 		fmt.Fprintf(os.Stderr, "%s: %s\n", svc.ID, err)
 	} else {
 		fmt.Println(migratedSvc.ID)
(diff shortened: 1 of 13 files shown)


tgjonestx:
retest this

tgjonestx:
My bad - I merged by mistake (too much context switching). Please review anyway, and I'll fix whatever needs to be addressed ASAP

jafcrocker:
Looking...

jafcrocker:
As coded this migrates a single service; I think we need to be able to migrate an entire service tree (i.e., a tenant), since that is the unit of upgrade.  I think the idea of migrating the whole service tree applies to services being migrated as part of ZenPack upgrades as well.

jafcrocker:
As coded the actual migration is running on the master host.  This implies that both the SDK and its environment (e.g., Python2.7) are installed on the host.  In the long run I believe we want it to run in a container; is the current implementation just temporary PoC?  If not, we need to make sure that the temporary directory is cleaned up on completion.

jafcrocker:
Purely for my education: is there a reason that the file handling code differs between createServiceMigrationInputFile and createServiceMigrationScriptFile?  Is one preferable?

jafcrocker:
We need to think about whether we need a lock to prevent multiple service migrations from running simultaneuously.  I am concerned about the multi-master case.  I *think* it's ok but not quite sure.

cli/cmd/service.go
```diff
@@ -678,6 +688,41 @@ func (c *ServicedCli) cmdServiceAdd(ctx *cli.Context) {
...
+	}
+
+	svc, err := c.searchForService(args[0])
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		return
+	}
+
+	var input *os.File
+	if len(args) >= 2 {
```
jafcrocker: Minor nit, but shouldn't we throw an error if there are more than two args?

cli/cmd/service.go
```diff
@@ -85,6 +85,16 @@ func (c *ServicedCli) initService() {
...
 				},
 			}, {
+				Name:         "migrate",
+				ShortName:    "mig",
+				Usage:        "Migrate an existing service",
+				Description:  "serviced service migrate SERVICEID PATH_TO_SCRIPT",
+				BashComplete: c.printServicesAll,
+				Action:       c.cmdServiceMigrate,
+				Flags: []cli.Flag{
+					cli.BoolFlag{"dry-run", "Executes the migration and validation without updateing anything"},
```
jafcrocker: Spelling: "updating" I wouldn't care except that it's user facing

tgjonestx:
Yes - the migration for the entire service tree is an outstanding issue that will be addressed in a future PR.

Will discuss the locking issue with some others

PR #1618, opened 2015-02-26
Feature/resources factory clean

rDr4g0n:


web/ui/src/common/baseFactory.js
```diff
@@ -65,6 +65,9 @@
                     }
 
                     deferred.resolve();
+                })
+                .error((data, status) => {
+                    console.error("Unable to update factory", data);
```
tgjonestx: Should probably log the `status` as well.

Will the polling continue after an error? If so, we should open a new issue to look into how we want to handle that. The topic is a out of scope for this PR, but if the server returns an error, repeating the same request is probably not going to improve the situation

web/ui/src/common/resourcesFactory.js
```diff
@@ -1,488 +1,248 @@
...
+        }
+
+        // generate methods from methodConfigs
+        var methods = {};
+        for(var name in methodConfigs){
+            methods[name] = generateMethod(methodConfigs[name]);
+        }
+
+        return {
+            assign_ip: methods.assignIP,
```
tgjonestx: Very nice :+1:

tgjonestx:
As we discussed, there was also the 'length not defined' error at line 300 in AppsController.js

web/ui/src/common/baseFactory.js
rDr4g0n: Good point. As discussed, we could try exponential backup after n errors, and display a notification in the UI that there is a problem with the endpoint. This probably should go into a separate PR/ticket

PR #1621, opened 2015-02-27
restart watch of endpoint when node goes away

jplouis:


container/endpoint.go
```diff
@@ -361,6 +361,17 @@ func (c *Controller) watchRemotePorts() {
 		return
 	}
 
+	//translate closing call to endpoint cancel
+	cancelEndpointWatch := make(chan bool)
+	go func() {
+		select {
+		case <-c.closing:
+			glog.Infof("Closing endpoint watchers")
+			close(endpointsWatchCanceller)
```
smousa: Is this a race condition?  You are writing to this channel below...

container/endpoint.go
```diff
@@ -380,17 +391,31 @@ func (c *Controller) watchRemotePorts() {
 			}
 			if !missingWatchers {
 				glog.V(2).Infof("all imports are being watched - cancelling watcher on /endpoints")
-				endpointsWatchCanceller <- true
-				return
+				select{
+				case endpointsWatchCanceller <- true:
```
smousa: You write to this channel here

container/endpoint.go
jplouis: the default should take care of it.

container/endpoint.go
```diff
@@ -380,17 +391,31 @@ func (c *Controller) watchRemotePorts() {
...
+					return
+				default:
```
jplouis: Default

container/controller.go
```diff
@@ -630,6 +630,7 @@ func (c *Controller) Run() (err error) {
 		return nil
 	}
 	defer client.Close()
+	c.Close()
```
jplouis: This was added to let things know it is shutting down.

PR #1624, opened 2015-03-02
Add serviced migration framework.

wwwtyro:


tgjonestx:
The file mode on `jenkins-build.sh` changed to `0644` - was that intentional?

migration/servicedmigration/version.py
```diff
@@ -0,0 +1,27 @@
+
+API_VERSION = "1.0.0"
+
+required = False
+
+def require(version):
```
tgjonestx: unit tests?

migration/servicedmigration/service.py
```diff
@@ -0,0 +1,28 @@
+
+import sys
+import json
+
+from version import versioned
+
+
+class Service:
```
tgjonestx: We probably want some kind of doc for the SDK in general.  Can you add some pydoc comments for these methods as you go?

migration/servicedmigration/service.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    Manages loading, altering, and commiting services.
+    """
+
+    @versioned
+    def __init__(self):
```
jafcrocker: I think __init__ should take a string as input parameter, rather than reading the filename directly from sys.argv.  Better yet, the string should be the json formatted string.  Perhaps even better would be __init__ take no parameters, but have a separate layer that deserializes a json string and construct the Service object.

I anticipate that the json file will eventually [contain the entire service tree](https://github.com/control-center/serviced/pull/1613#issuecomment-76758816), not just one service.  The proposed refactoring would ease that transition.

migration/servicedmigration/version.py
```diff
@@ -0,0 +1,33 @@
...
+    else:
+        required = True
+
+def versioned(func):
+    """
+    Decorator for all functions depending on API compatibility.
+    """
+    def func_wrapper(*args):
+        if not required:
+            raise RuntimeError("You must first call require(version_number) before using this funciton.")
```
jafcrocker: Typo in user-facing string

migration/servicedmigration/service.py
wwwtyro: Good idea. Any objection to a wrapper function that handles making a Service from argv[] for the user-developer? I'd like to avoid boilerplate.

wwwtyro:
test this

wwwtyro:
test this

PR #1637, opened 2015-03-05
Feature/minimal migration functionality

wwwtyro:


migration/makefile
```diff
@@ -0,0 +1,6 @@
+test:
```
tgjonestx: The toplevel make should invoke this test so that the CI build will run these unit-tests

tgjonestx:
:+1:  from my perspective, but I'm a python noob. 
@jafcrocker - can you take a look too?

migration/makefile
wwwtyro: Done!

jafcrocker:
*Filters:*
* filter should be a function that returns a boolean
* if the map is a useful convention, create an object that uses the map
* it would be useful to chain filters; i.e., find me services tagged as daemons whose parent are tagged as collectors and whose parents are tagged as hubs

*getServices()*
* I think that you should probably drop the parent and child filter arguments -instead have routines for iterating over children and parents of a given service, and the user can filter the results of that iterator.

*versioned:*
* why check each function?  Maybe just check once when the library initialized?
* removeRun/addRun are not versioned
	
*getServices/_reloadServiceList/_serviceList*
* I don't think this should be a singleton-style function.  Instead I think the API should look something like this:
```
	ctx = ServiceContext()
	ctx.load(open(filename, 'r')) 
	service = ctx.getService(filter)
	service.setDescription("foobar")
	ctx.commit(open(filename, 'w'))
```
* The crucial point is that the there can be multiple service contexts, that they can have different sources, be loaded and saved independently, etc.  That would get the test and argument-handling code out of service.py.  
* The ServiceContext is a natural place to cache parent/child pointers.
	
	
*Supporting Objects*
We should not be exposing raw json to user - otherwise we cannot support future schema versions. Need objects for attributes:  
* Endpoint
* Healthcheck
* Volumes
* Run
* LogConfig

wwwtyro:
Closing for now until we have a discussion about the API.

PR #1638, opened 2015-03-05
Preserve docker image history when performing a backup/restore

smousa:
https://jira.zenoss.com/browse/CC-659
https://github.com/zenoss/go-dockerclient/pull/23


smousa:
https://github.com/zenoss/go-dockerclient/pull/23

PR #1643, opened 2015-03-06
Feature/base factory documentation

rDr4g0n:


web/ui/karma.conf.js
```diff
@@ -96,10 +95,10 @@ module.exports = function(config) {
     // Code coverage results below these thresholds will trigger a build failure
     // For more info, see https://www.npmjs.com/package/karma-threshold-reporter
     thresholdReporter: {
-      statements: 5,
-      branches: 0,
-      functions: 5,
-      lines: 5
+      statements: 8,
```
tgjonestx: Looks like these can be tightened even further to something like:
```
   statements: 17,
   branches: 10,
   functions: 14,
   lines: 17
``` 

In practice, we might want the thresholds to stay within -2 of the latest high-water mark rather than -1 like I have above.  On one hand, we should strive to increase coverage slowing over time, but minor additions shouldn't trigger failures.  In the other words, the values I suggested might be too tight, but there's certainly room to increase them over what's there now.

tgjonestx:
Side note - we should work towards refactoring (reimplementing) tests from `web/ui/test/obsoleteSpec.js` into new test specs when/where it makes sense so that ultimately, we can get rid of obsoleteSpec.js

web/ui/karma.conf.js
```diff
@@ -87,7 +87,6 @@ module.exports = function(config) {
       // specify a common output directory
       dir: 'test/results/coverage',
       reporters: [
-        { type: 'html', subdir: 'html' },   // use for local dev review
```
tgjonestx: FYI - If you change the `test` task to use Chrome or Firefox for one of the `browsers` then you can get the interactive report which show you coverage detail.  The coverage for baseFactory is super good (generally >90). The coverage for servicesFactory is so so  (in the 60s). The big outliers are `mergeArray`, and the getters for "hosts" and "addresses"

PR #1656, opened 2015-03-13
add support for turning off the nfs client on an agent

jplouis:


kindkid:
:+1: just waiting for jenkins

kindkid:
Is there a separate JIRA ticket for preventing services that use DFS from being scheduled on agents that have the nfs client turned off?

jplouis:
There is not a ticket for preventing the services being scheduled.  I will open one.

PR #1661, opened 2015-03-19
Feature/grid directive ngtable

rDr4g0n:


tgjonestx:
retest this

tgjonestx:
retest this

tgjonestx:
I fetched this PR, rebuilt, and got the following error after logging in.  Looks like a race of some sort in the AppsController.
```
TypeError: Cannot read property 'length' of undefined
    at controlplane.js:2523
```
where line 2523 is the if statement below
```
            // start polling for apps
            servicesFactory.activate();
            servicesFactory.update().then(function () {
                // if only isvcs are deployed, and this is the first time
                // running deploy wizard, show the deploy apps modal
                if (!$cookies.autoRunWizardHasRun && $scope.apps.length === 1) {
                    $scope.modalAddApp();
                }
```

tgjonestx:
UX nit - some tables show the default sort. but some don't.  Sorting by the first column (name) is pretty obvious, but would be kind of nice if we showed that.

Tables that show default sort
 * Applications
 * Resource Pools
 * Virtual IPs
 * Hosts for a resource pool

Tables that do not show default sort
 * Application templates
 * Hosts
 * Backup/restore

PR #1669, opened 2015-03-24
Override the service migration SDK version

tgjonestx:
Addresses Rally user stories US12315, US12316 and US12317

### DIFF
--- a/facade/service.go
+++ b/facade/service.go
@@ -1530,14 +1530,14 @@ func executeMigrationScript(serviceID, tmpDir, scriptFilePath, inputFilePath str
 	if tagName != "" {
 		glog.V(2).Infof("Facade:executeMigrationScript: using docker tag=%q", tagName)
 	}
-	dokckerImage := fmt.Sprintf("%s:%s", SERVICE_MIGRATION_IMAGE_NAME, tagName)
+	dockerImage := fmt.Sprintf("%s:%s", SERVICE_MIGRATION_IMAGE_NAME, tagName)
 
 	mountPath := fmt.Sprintf("%s:/migration", tmpDir)
 	cmd := exec.Command("docker",
 		"run", "--rm", "-t",
 		"--name", "service-migration",
 		"-v", mountPath,
-		dokckerImage,
+		dockerImage,
 		"python", containerScript, containerInputFile, containerOutputFile)
 
 	glog.V(2).Infof("Facade:executeMigrationScript: service ID %+v: cmd: %v", serviceID, cmd)
--- a/facade/service.go
+++ b/facade/service.go
@@ -107,10 +107,10 @@ func (f *Facade) UpdateService(ctx datastore.Context, svc service.Service) error
 }
 
 // TODO: Should we use a lock to serialize migration for a given service? ditto for Add and UpdateService?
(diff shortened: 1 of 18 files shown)


facade/service.go
```diff
@@ -1517,15 +1518,26 @@ func executeMigrationScript(serviceID, tmpDir, scriptFilePath, inputFilePath str
...
+	if sdkVersion != "" {
+		tagName = sdkVersion
+	} else if tagOverride := os.Getenv("SERVICED_SERVICE_MIGRATION_TAG"); tagOverride != "" {
+		tagName = tagOverride
+	}
+
+	if tagName != "" {
+		glog.V(2).Infof("Facade:executeMigrationScript: using docker tag=%q", tagName)
+	}
+	dokckerImage := fmt.Sprintf("%s:%s", SERVICE_MIGRATION_IMAGE_NAME, tagName)
```
wplu: dockerImage typo?

pkg/serviced.default
```diff
@@ -126,7 +126,8 @@ SERVICED_FS_TYPE=btrfs
 # Default: 1    (enable)
 # SERVICED_MONITOR_DFS_MASTER_RESTART=1
 
+# Overrides the default for the service migration image.
+# SERVICED_SERVICE_MIGRATION_TAG=1.0.2
```
wplu: I'm not seeing this being used anywhere

tgjonestx: As I understand it, all of those settings are commented out by default, and the user can uncomment if they want to enable it.   The variable itself is referenced at line 1526 in facade/service.go
```
	} else if tagOverride := os.Getenv("SERVICED_SERVICE_MIGRATION_TAG"); tagOverride != "" {
		tagName = tagOverride
	}
```

facade/service.go
```diff
@@ -1517,15 +1518,26 @@ func executeMigrationScript(serviceID, tmpDir, scriptFilePath, inputFilePath str
...
+
 	mountPath := fmt.Sprintf("%s:/migration", tmpDir)
 	cmd := exec.Command("docker",
 		"run", "--rm", "-t",
 		"--name", "service-migration",
 		"-v", mountPath,
-		SERVICE_MIGRATION_IMAGE_NAME,
+		dokckerImage,
```
wwwtyro: Here too

script/nodes.go
```diff
@@ -114,6 +114,32 @@ func buildNode(ctx *parseContext, cmd string, args []string) (node, error) {
...
+	//try to match
+	const SDK_VERSION_PATTERN = `^SDK=([a-zA-Z0-9.\-_]+)$`
+	sdkVerRegex := regexp.MustCompile(SDK_VERSION_PATTERN)
+	matches := sdkVerRegex.FindStringSubmatch(arg)
+	if len(matches) != 2 || matches[1] == "" {
+		return "", fmt.Errorf("arg %s did not match %s", arg, SDK_VERSION_PATTERN)
+	}
+	return matches[1], nil
+}
+
```
wwwtyro: Is there any reason we wouldn't want an exact match, here? i.e., just test "abc" == "abc" ?

wwwtyro:
test this

pkg/serviced.default
wplu: ahhh - must have been different commit

PR #1672, opened 2015-03-27
Feature/cc 878

morr-zenoss:


facade/servicetemplate.go
```diff
@@ -446,3 +437,26 @@ func reloadLogstashContainerImpl(ctx datastore.Context, f FacadeInterface) error
...
+			return err
+		}
+		tag := imageID.Tag
+		if tag == "" {
+			tag = "latest"
+		}
+		image := fmt.Sprintf("%s:%s", imageID.BaseName(), tag)
+		glog.Infof("Pulling image %s", image)
+		if err := docker.PullImage(image); err != nil {
+			glog.Warningf("Unable to pull image %s", image)
```
alapidas: You should maybe return the err in here instead

morr-zenoss: Are you talking about the warning in 458? If so, that was in the original code (I just extracted it to another method to avoid duplication). I think they used a warning there so that we would pull as many docker images as possible, even if one failed.

PR #1676, opened 2015-04-02
back-end performance metrics api

smousa:


metrics/memoryusage.go
```diff
@@ -0,0 +1,168 @@
...
+
+			if last = dp.Value.Value; last > max {
+				max = last
+			}
+			sum += dp.Value.Value
+		}
+		mems[i].Last = int64(last)
+		mems[i].Max = int64(max)
+		mems[i].Average = int64(sum / float64(len(result.Datapoints)))
+	}
```
wwwtyro: I guess we'd want to use some proper rounding function here, but probably doesn't matter for single byte values.

metrics/memoryusage.go
```diff
@@ -0,0 +1,168 @@
...
+
+	metrics := make([]MetricOptions, len(instances))
+	for i, instance := range instances {
+		metrics[i] = MetricOptions{
+			Metric: "cgroup.memory.totalrss",
+			Name:   fmt.Sprintf("%s.%d", instance.ServiceID, instance.InstanceID),
+			Tags: map[string][]string{
+				"controlplane_service_id":  []string{instance.ServiceID},
+				"controlplane_instance_id": []string{fmt.Sprintf("%d", instance.InstanceID)},
+			},
```
wwwtyro: What is the default aggregator, since none is defined here?

PR #1682, opened 2015-04-08
Feature/cc 882

morr-zenoss:


smousa:
waiting for tests to pass

PR #1687, opened 2015-04-14
service migration clone support

wwwtyro:


facade/service.go
```diff
@@ -176,32 +176,90 @@ func (f *Facade) MigrateService(ctx datastore.Context, request dao.ServiceMigrat
...
-				glog.V(2).Infof("Facade:MigrateService: dry-run of migration script complete for serviceID %+v", svc.ID)
+	// Validate the modified services.
+	for _, svc := range svcsMap["modified"] {
+		if err = f.verifyServiceForUpdate(ctx, svc, nil); err != nil {
+			return err
+		}
+	}
+
+	// Validate the cloned services.
+	existing, err := f.serviceStore.GetServices(ctx)
```
tgjonestx: This function is getting really long. Can you refactor lines 192-213 into a private method, something like validateClonedServices?

facade/service.go
```diff
@@ -176,32 +176,90 @@ func (f *Facade) MigrateService(ctx datastore.Context, request dao.ServiceMigrat
...
+		if err = f.stopServiceForUpdate(ctx, *svc); err != nil {
+			return err
+		}
+		if err = f.migrateService(ctx, svc); err != nil {
+			return err
 		}
 	}
 
-	return err
+	// Wipe out the cloned service's endpoint AddressAssignment data.
```
tgjonestx: Seems like this should be part of validateClonedServices. Is there any reason for these updates to Endpoints to occur at this particular point in the execution path?

facade/service.go
```diff
@@ -176,32 +176,90 @@ func (f *Facade) MigrateService(ctx datastore.Context, request dao.ServiceMigrat
...
+	for _, svc := range svcsMap["cloned"] {
+		for _, ep := range svc.Endpoints {
+			ep.AddressAssignment = addressassignment.AddressAssignment{}
+		}
+	}
+
+	// Add the cloned services.
+	for _, svc := range svcsMap["cloned"] {
```
tgjonestx: We should verify this with someone who knows the domain better than I do, but I'm guessing that cloned services need to be added BEFORE modified services are migrated. Specifically, if I create a clone and then reparent an existing service to that clone, then I'm guessing the update for the existing service will fail if the clone does not already exist when we migrate the existing service.

tgjonestx:
The last 2 commits address my earlier comments, but I think we need (at least) 4 more tests cases in controlplanedao_test.go to cover this new logic:

1. A successful clone 
1. A clone that fails validation because of a duplicate service name
1. A clone that fails validation because of a duplicate endpoint
1. A clone that fails AddService; for instance, the value of DesiredState is invalid

PR #1690, opened 2015-04-17
fixes CC-889

wwwtyro:


dao/elasticsearch/controlplanedao.go
```diff
@@ -124,16 +125,22 @@ func NewControlPlaneDao(hostName string, port int) (*ControlPlaneDao, error) {
...
 		hostName: hostName,
 		port:     port,
+		rpcPort:  rpcPort,
 	}
 
 	return dao, nil
 }
 
-func NewControlSvc(hostName string, port int, facade *facade.Facade, varpath, fsType string, maxdfstimeout time.Duration, dockerRegistry string) (*ControlPlaneDao, error) {
+func NewControlSvc(hostName string, port int, facade *facade.Facade, varpath, fsType string, rpcPort string, maxdfstimeout time.Duration, dockerRegistry string) (*ControlPlaneDao, error) {
```
tgjonestx: Kind of weird to have two different arguments for the same kind of thing, but they have different types.  I think `rpcPort` should be an int.  To get really picky, the 'correct' type should be `unit16`

PR #1692, opened 2015-04-21
Configuration parser for default file information

smousa:
https://jira.zenoss.com/browse/CC-927

### DIFF
--- a/cli/cmd/parser.go
+++ b/cli/cmd/parser.go
@@ -1,0 +1,167 @@
+// Copyright 2015 The Serviced Authors.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"strconv"
+	"strings"
+)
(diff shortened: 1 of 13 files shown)


main.go
```diff
@@ -35,5 +39,28 @@ func main() {
...
+func getConfigs(args []string) (*cmd.EnvironConfigReader, error) {
+	filename := conffile
+
+	for i, arg := range args {
+		if strings.HasPrefix(arg, "--config-file") {
+			if idx := strings.IndexByte(arg, '='); idx >= 0 {
+				filename = arg[idx+1:]
+			} else if i+1 < len(args) {
+				filename = args[i+1]
+			}
```
tgjonestx: Isn't `--config-file` without a file name an error?

cli/cmd/parser.go
```diff
@@ -0,0 +1,167 @@
...
+	if err := r.parse(file); err != nil {
+		return r, err
+	}
+	return r, nil
+}
+
+// parse is a really dumb reader parser.  It maps only key values in the form
+// of key=value and strips whitespaces surrounding either field.  If the format
+// does not match, then and error will return.
+func (p *EnvironConfigReader) parse(reader io.Reader) error {
```
tgjonestx: I know this works, but I think it's a little harder to read for correctness when it's parsing character by character, instead of reading line-by-line.  How about something like this?
```
func (p *EnvironConfigReader) parse(reader io.Reader) error {
	var (
		line string
		err  error
	)

	bufReader := bufio.NewReader(reader)
	for err != io.EOF {
		line, err = bufReader.ReadString('\n')
		if err != nil && err != io.EOF {
			return err
		}

		line = strings.TrimSpace(strings.Split(line, "#")[0])
		if err := keyvalue([]byte(line)); err != nil {
			return err
		}
	}

	if err := keyvalue([]byte(line)); err != nil {
		return err
	}

	return nil
}
```

main.go
smousa: yes, but the command line parser should capture that I think

PR #1697, opened 2015-04-23
isvc healthcheck cleanup

tgjonestx:
1. Remove timeouts inside the healthchecks because we now have timeouts in isvcs/container.go
2. Remove cluster.Health() from ES healthchecks because it incorrectly reports all ES instances
failed, when just one of them are failed


tgjonestx:
retest this

PR #1701, opened 2015-04-28
refine isvcs healthchecks

tgjonestx:
Refinements to periodic health checking of iservices:
 1.  At startup, repeat attempts at healthcheck for up to 2 minutes. Fail serviced startup if the health check is still negative after 2 minutes. 
 1. After startup, if a periodic health check fails, then don't restart; just report a failed status.
 1. After startup, if an isvc exits, then restart it, UNLESS it has failed to start and run for more than 1 minute 3 times in a row (i.e. if it's starting/exiting repeatedly, then don't continue trying to restart it).

### DIFF
--- a/isvcs/container.go
+++ b/isvcs/container.go
@@ -86,7 +86,7 @@ type actionrequest struct {
 }
 
 // HealthCheckFunction- A function to verify the service is healthy
-type HealthCheckFunction func() error
+type HealthCheckFunction func(halt <-chan struct{}) error
 
 type healthCheckDefinition struct {
 	healthCheck HealthCheckFunction
@@ -559,8 +559,10 @@ func (svc *IService) startupHealthcheck() <-chan error {
 
 func (svc *IService) runCheckOrTimeout(checkDefinition healthCheckDefinition) error {
 	finished := make(chan error, 1)
+	halt := make(chan struct{}, 1)
+
 	go func() {
-		finished <- checkDefinition.healthCheck()
+		finished <- checkDefinition.healthCheck(halt)
 	}()
 
 	var result error
@@ -570,6 +572,7 @@ func (svc *IService) runCheckOrTimeout(checkDefinition healthCheckDefinition) er
 	case <-time.After(checkDefinition.Timeout):
 		glog.Errorf("healthcheck timed out for %s", svc.name())
 		result = fmt.Errorf("healthcheck timed out")
+		halt <- struct{}{}
 	}
 	return result
 }
--- a/isvcs/container.go
+++ b/isvcs/container.go
@@ -113,6 +113,8 @@ type IService struct {
 	exited         <-chan int
 	root           string
 	actions        chan actionrequest
+	startTime      time.Time
+	restartCount   int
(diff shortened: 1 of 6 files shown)


isvcs/container.go
```diff
@@ -459,17 +469,61 @@ func (svc *IService) run() {
...
+				}
+				glog.Errorf("isvc %s not restarted; failed too many times", svc.Name)
+				svc.stop()
+				svc.exited = nil
 			}
 		}
 	}
 }
 
+// Return true if the service is flapping (starting/stopping repeatedly).
```
wplu: perhaps "isFlapping returns true if the service is flapping (starting/stopping repeatedly)."

isvcs/container.go
```diff
@@ -504,20 +558,40 @@ func (svc *IService) checkvolumes(ctr *docker.Container) bool {
 	return true
 }
 
-func (svc *IService) healthcheck(currentTime int64) <-chan error {
+// Run the default healthcheck (if any) and the return the result.
```
wplu: perhaps "startupHealthcheck runs the default" ...

isvcs/container.go
```diff
@@ -504,20 +558,40 @@ func (svc *IService) checkvolumes(ctr *docker.Container) bool {
...
+// If the healthcheck fails, then this method will sleep 1 second, and then
+// repeat the healthcheck, continuing that sleep/retry pattern until
+// the healthcheck succeeds or 2 minutes has elapsed.
+//
+// An error is returned if the no healtchecks succeed in the 2 minute interval,
```
wplu: healthchecks

isvcs/manager.go
```diff
@@ -147,20 +147,16 @@ func (m *Manager) GetHealthStatus(name string) (dao.IServiceHealthResult, error)
...
-	//        not found, or should we return an instance of IServiceHealthResult
-	//        with HealthStatuses["running"] == a-failed-instance?
-	ctr, err := docker.FindContainer(svc.name())
-	if err != nil {
-		glog.Errorf("Could not find container for isvc %s: %s", svc.Name, err)
-		return dao.IServiceHealthResult{}, err
+	if ctr, err := docker.FindContainer(svc.name()); err == nil {
+		result.ContainerID = ctr.ID
+	} else {
+		result.ContainerID = "<none>"
```
wplu: perhaps empty string instead of "<none>"

isvcs/es.go
```diff
@@ -116,14 +116,20 @@ func elasticsearchHealthCheck(port int) func() error {
...
 			}
 
 			if time.Since(lastError) > minUptime {
 				break
 			}
-			time.Sleep(time.Millisecond * 1000)
+
+			select {
+			case <-halt:
+				glog.V(1).Infof("Quit healthcheck for elasticsearch at %s", baseUrl)
```
wplu: please print the port number

tgjonestx: The string `baseUrl` already contains the port number

wplu: oh cool!

PR #1703, opened 2015-04-28
Feature/service migration endpoint

wwwtyro:
Related to:
https://github.com/zenoss/zenoss-prodbin/pull/850
https://github.com/control-center/service-migration/pull/18

### DIFF
--- a/facade/service.go
+++ b/facade/service.go
@@ -134,7 +134,7 @@ func (f *Facade) MigrateService(ctx datastore.Context, request dao.ServiceMigrat
 		return err
 	}
 
-	svcs, err2 := f.getServiceList(ctx, svc.ID)
+	svcs, err2 := f.GetServiceList(ctx, svc.ID)
 	if err2 != nil {
 		return err2
 	}
@@ -1465,7 +1465,7 @@ func (f *Facade) updateServiceConfigs(ctx datastore.Context, oldSvc, newSvc *ser
 	return nil
 }
 
-func (f *Facade) getServiceList(ctx datastore.Context, serviceID string) ([]*service.Service, error) {
+func (f *Facade) GetServiceList(ctx datastore.Context, serviceID string) ([]*service.Service, error) {
 	svcs := make([]*service.Service, 0, 1)
 
 	err := f.walkServices(ctx, serviceID, true, func(childService *service.Service) error {
--- a/facade/service.go
+++ b/facade/service.go
@@ -181,6 +181,8 @@ func (f *Facade) RunMigrationScript(ctx datastore.Context, request dao.RunMigrat
 		return err
 	}
 
+	migrationRequest.DryRun = request.DryRun
+
 	err = f.MigrateServices(ctx, *migrationRequest)
 
 	return err
--- a/facade/service.go
+++ b/facade/service.go
@@ -1717,10 +1717,9 @@ func readServiceMigrationRequestFromFile(outputFileName string) (*dao.ServiceMig
 		return nil, fmt.Errorf("could not read new service definition: %s", err)
 	}
 
-	// svcs := make(map[string][]*service.Service)
-	var smr dao.ServiceMigrationRequest
-	if err = json.Unmarshal(data, &smr); err != nil {
+	var request dao.ServiceMigrationRequest
+	if err = json.Unmarshal(data, &request); err != nil {
 		return nil, fmt.Errorf("could not unmarshall new service definition: %s", err)
 	}
-	return &smr, nil
+	return &request, nil
 }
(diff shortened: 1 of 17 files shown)


web/resources.go
```diff
@@ -235,6 +235,37 @@ func getIRS() []dao.RunningService {
 	return services
 }
 
+func restGetServicesForMigration(w *rest.ResponseWriter, r *rest.Request, client *node.ControlClient) {
```
tgjonestx: Should be renamed simple `restGetServiceList`

web/routes.go
```diff
@@ -77,6 +77,10 @@ func (sc *ServiceConfig) getRoutes() []rest.Route {
 		rest.Route{"PUT", "/services/:serviceId/startService", gz(sc.authorizedClient(restStartService))},
 		rest.Route{"PUT", "/services/:serviceId/stopService", gz(sc.authorizedClient(restStopService))},
 
+		// Services (migration)
+		rest.Route{"GET", "/servicemigration/:serviceId", sc.authorizedClient(restGetServicesForMigration)},
```
tgjonestx: After some more thought, I don't think it makes sense to create a new endpoint just for this.
I'm paraphrasing slightly, but 
`GET /servicemigration/:serviceId` should be replaced with `GET /services/:servicedId?includeChildren=true`.
The point is that returning a list of a service plus it's children is generic operation for services in general, so we should be able to incorporate that function into the existing `/services` namespace.

web/routes.go
```diff
@@ -77,6 +77,10 @@ func (sc *ServiceConfig) getRoutes() []rest.Route {
...
+		rest.Route{"POST", "/servicemigration", sc.authorizedClient(restPostServicesForMigration)},
```
tgjonestx: The endpoint should be `POST /services/:serviceId/migrate`

web/resources.go
```diff
@@ -235,6 +235,37 @@ func getIRS() []dao.RunningService {
...
+	}
+	w.WriteJson(&services)
+}
+
+func restPostServicesForMigration(w *rest.ResponseWriter, r *rest.Request, client *node.ControlClient) {
+	var smr dao.ServiceMigrationRequest
+	err := r.DecodeJsonPayload(&smr)
+	if err != nil {
+		glog.Errorf("Could not decode services for migration: %v", err)
+		restServerError(w, err)
```
tgjonestx: Just an FYI - this is pattern (used throughout serviced) will send a RESTafarian to screaming. The proper HTTP response to malformed JSON is 400, not 500. No need to change in this case, but it's a bigger change that should be made throughout the serviced REST API.

web/resources.go
```diff
@@ -235,6 +235,37 @@ func getIRS() []dao.RunningService {
...
+		return
+	}
+	var unused int
+	if err = client.MigrateServices(smr, &unused); err != nil {
+		restServerError(w, err)
```
tgjonestx: Just FYI - both restBadRequest() and restServerError() return an HTTP 500 which is totally wrong.  500 is only correct for situations like the server segfaults.  But fixing all the REST methods to return the proper HTTP status codes is outside the scope of this PR

facade/service.go
```diff
@@ -176,20 +176,31 @@ func (f *Facade) MigrateService(ctx datastore.Context, request dao.ServiceMigrat
...
+	return err
+
+}
+
+func (f *Facade) MigrateServices(ctx datastore.Context, request dao.ServiceMigrationRequest) error {
+	var err error
+
 	// Validate the modified services.
-	for _, svc := range svcsMap["modified"] {
+	for _, svc := range request.Modified {
```
tgjonestx: Better :-)

facade/service.go
```diff
@@ -1698,16 +1711,16 @@ func executeMigrationScript(serviceID string, serviceContainer *docker.Container
...
+func readServiceMigrationRequestFromFile(outputFileName string) (*dao.ServiceMigrationRequest, error) {
+	data, err := ioutil.ReadFile(outputFileName)
 	if err != nil {
 		return nil, fmt.Errorf("could not read new service definition: %s", err)
 	}
 
-	svcs := make(map[string][]*service.Service)
-	err = json.Unmarshal(newServiceDefinition, &svcs)
-	if err != nil {
+	// svcs := make(map[string][]*service.Service)
```
tgjonestx: this is dead code that can be deleted

facade/service.go
```diff
@@ -1698,16 +1711,16 @@ func executeMigrationScript(serviceID string, serviceContainer *docker.Container
...
+	var smr dao.ServiceMigrationRequest
```
tgjonestx: `var request dao.ServiceMigrationRequest`

PR #1705, opened 2015-04-28
cleanup old tests, fix pools not appearing in IP Assignments table

rDr4g0n:


jhanson:
Please separate out the bug fix from the refactor/cleanup

PR #1729, opened 2015-05-21
Add support for deploying services from service definitions to the mi…

wwwtyro:
…gration process.

### DIFF
--- a/facade/service.go
+++ b/facade/service.go
@@ -199,8 +199,8 @@ func (f *Facade) MigrateServices(ctx datastore.Context, request dao.ServiceMigra
 		}
 	}
 
-	// Make required mutations to the cloned services.
-	for _, svc := range request.Cloned {
+	// Make required mutations to the added services.
+	for _, svc := range request.Added {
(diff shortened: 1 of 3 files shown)


facade/service.go
```diff
@@ -213,15 +213,19 @@ func (f *Facade) MigrateServices(ctx datastore.Context, request dao.ServiceMigra
...
+	if err = f.validateServiceDeploymentRequests(ctx, request.Deploy); err != nil {
 		return err
 	}
 
 	// If this isn't a dry run, make the changes.
 	if !request.DryRun {
 
-		// Add the cloned services.
-		for _, svc := range request.Cloned {
+		// Add the added services.
```
iancmcc: This code matches the order from before, but I don't understand: why don't we want to apply our existing state BEFORE we add the new services? Or rather, why is it safe to do it in this order?

wwwtyro:
Test this

wwwtyro:
retest this

wwwtyro:
retest this

PR #1732, opened 2015-05-27
CC-990: Runs should exit with the container command's exit code

alapidas:


cli/api/interfaces.go
```diff
@@ -75,7 +75,7 @@ type API interface {
 
 	// Shell
 	StartShell(ShellConfig) error
-	RunShell(ShellConfig) error
+	RunShell(ShellConfig) (int, error)
```
wplu: nice

PR #1742, opened 2015-06-05
Handle killing a non-iteractive run

alapidas:


smoke.sh
```diff
@@ -155,10 +155,21 @@ test_service_shell() {
...
 
 test_service_run() {
-    ${SERVICED} service run s2 exit0; local rc="$?"
+    local rc=""
+    ${SERVICED} service run s2 exit0; rc="$?"
     [ "$rc" -eq 0 ] || return "$rc"
     ${SERVICED} service run s2 exit1; rc="$?"
     [ "$rc" -eq 42 ] || return "255"
+    # make sure kills to 'runs' are working OK
+    for signal in 2 15; do
```
wplu: why not INT and TERM instead of 2 15

wplu:
waiting for build test

PR #1745, opened 2015-06-08
Install npm tools for UI builds in zenoss/serviced-build

tgjonestx:
DO NOT MERGE.  If the changes in this PR are acceptable, I need to push a new copy of the zenoss/serviced-build docker image to docker hub before the PR build will pass.

This PR improves build times for developers who do not have Node.js installed by:
 * NOT building the zenoss/serviced-build docker image on each build, and
 * NOT running `npm install` on each build

However, it requires a small process change w.r.t. the serviced-build image and UI dev. Specifically, anytime the contents of the serviced-build are changed, someone must explicitly push them to Docker Hub (see the `pushDockerImage` make target). Also, anytime a UI developer changes `package.json' to modify the set of NPM packages used for building the UI, they must build and push a new copy of the Docker image.

The basic idea is to run npm-install when the image is built. At runtime when we're building as another user, the `web/uil/makefile` calls `npmLink.sh` to dynamically create soft-link from web/ui/node_modules to a well known location in the docker image (`/npm-cache/serviced/node_modules`).


makefile
```diff
@@ -159,15 +159,22 @@ docker_SRC = github.com/docker/docker
 FORCE:
 
 serviced: $(GODEP)
-serviced: FORCE 
+serviced: FORCE
 	$(GO) build $(GOBUILD_FLAGS) ${LDFLAGS}
 	make govet
 	if [ -n "$(GOBIN)" ]; then cp serviced $(GOBIN)/serviced; fi
 
+buildDockerImage: docker_ok
```
smousa: need a more descriptive name here

makefile
```diff
@@ -159,15 +159,22 @@ docker_SRC = github.com/docker/docker
...
+	cp web/ui/package.json build
+	cp web/ui/npm-shrinkwrap.json build
+	docker build -t zenoss/serviced-build:$(VERSION) build
+
+pushDockerImage: docker_ok
```
smousa: and here

makefile
```diff
@@ -159,15 +159,22 @@ docker_SRC = github.com/docker/docker
...
+	docker push zenoss/serviced-build:$(VERSION)
```
smousa: Does this command happen every time *anyone* does a build?

tgjonestx: No. Both of the targets for `buildDockerImage` and `pushDockerImage` are intended for manual execution (e.g. `make buildDockerImage pushDockerImage`). They only need to be executed when someone needs to refresh the contents of the `zenoss/serviced-build` image.   The regular builds for developers, and the different Jenkins builds do not use these targets and they no longer include a `docker build -t zenoss/serviced-build`

smousa:
looks good to me

tgjonestx:
Retest this

smousa:
retest this

smousa:
retest this

PR #1747, opened 2015-06-10
remove realm from scheduler

smousa:
* Divided scheduler into a group of managers.  Each manager must be a leader in order to run its process.  This is especially effective in multi-master such that tasks can be distributed across masters

Managers:
Storage: sets up the storage leader and keeps storage clients in sync
Resource: schedules services across hosts per pool.
TTL: manages ttl operations like local sync, snapshot, and eventually remote sync.

### DIFF
--- a/coordinator/storage/client.go
+++ b/coordinator/storage/client.go
@@ -14,8 +14,8 @@
 package storage
 
 import (
-	"fmt"
 	"os"
+	"path"
 	"sync"
 	"time"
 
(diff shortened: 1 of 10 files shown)


coordinator/storage/client.go
```diff
@@ -98,7 +98,7 @@ func (c *Client) loop() {
 
 	var doneC chan<- string
 	var leader client.Leader
-	nodePath := fmt.Sprintf("/storage/clients/%s", node.IPAddr)
+	nodePath := path.Join(storageClientsPath, node.IPAddr)
```
wplu: +1

dfs/snapshotttl.go
```diff
@@ -48,6 +53,7 @@ func RunSnapshotTTL(client SnapshotTTLInterface, cancel <-chan interface{}, min,
 // Implements utils.TTL
 func (ttl *SnapshotTTL) Purge(age time.Duration) (time.Duration, error) {
 	expire := time.Now().Add(-age)
+	glog.Infof("Purging snapshots created before %s", expire)
```
wplu: +1

dfs/snapshotttl.go
```diff
@@ -69,6 +75,7 @@ func (ttl *SnapshotTTL) Purge(age time.Duration) (time.Duration, error) {
 				if parts := strings.Split(s.SnapshotID, "_"); len(parts) == 2 {
 					if ts, err := time.Parse(timeFormat, parts[1]); err == nil {
 						if timeToLive := ts.Sub(expire); timeToLive <= 0 {
+							glog.Infof("Deleting snapshot %s; created at %s", s.SnapshotID, ts)
```
wplu: +1

PR #1748, opened 2015-06-10
Feature/cc 945

romanbakaleyko:


wplu:
ship when test build succeeds

PR #1751, opened 2015-06-15
Feature/governed pool

smousa:


smousa:
retest this

smousa:
retest this

smousa:
retest this

smousa:
retest this

tgjonestx:
I don't think we're going to need utils/codec.go. Below is the pseudo code for what I think the downstream will do to post a request to the upstream. The `authorizer.AddToken()` will read the request body for all of the info it needs to encode. 

```
import "github/control-center/serviced/auth/delegateauth"

	authorizer, err := NewDelegateAuthorizer()

	// url - the URL to upstream endpoint, something like:
	//       "https://control-center:<upstreamPort>/[uriPefix/]pool/<remotePoolID>/services"
	//        - upstreamPort is the port the upstream serviced is listening on
	//        - uriPrefix is an optional value from the serviced configuration to deal with situations where
	//            the customer's upstream endpoint on the public internet has some extra prefix on the URL
	//        - remotePoolID is the remote pool id
	// body - the JSON body of the request containing whatever info we need to send upstream
	request, err := http.NewRequest("POST", url, body)
	
	err = authorizer.AddToken(remotePoolID, request, uriPrefix)

	<< send the request >>
```

tgjonestx:
Disregard my last comment - the encoding/decoding for sending the packet of key/pool info OOB (which is in this PR) is a separate issue from the JWT stuff I mentioned in my last comment

PR #1754, opened 2015-06-16
Add libraries to create/verify JWT authentication tokens

tgjonestx:
Initial implementation, still WIP.  The new packages in this PR are:

 * `auth/jwt` - provides generic JWT handling per Zenoss-specific conventions. Shields the rest of the application from the 3rd-party lib for JWT.  SHOULD NOT be called by application-level  code. SHOULD be called by JWT implementations which are specific to particular authentication scheme (such as authenticating a downstream CC with an upstream CC)
 * `auth/delegateauth` - provides a Zenoss-specific JWT implementation for authenticating a downstream CC with an upstream CC. Uses `auth/jwt`. This is the package that SHOULD be used by application-level code to add JWT tokens to an outbound request, and verify Authorization headers on in bound requests. 

Remaining work for future PRs:
 * Update `delegateAuth.getKeyLookup()` to lookup HMAC encryption key based on poolID
 * Integration `DelegateAuthorizer` with web tier for new endpoint that listens for requests from downstream CC delegates


auth/jwt/signer.go
```diff
@@ -0,0 +1,41 @@
+// You may obtain a copy of the License at
```
smousa: Missing half of your boilerplate

cli/api/api.go
```diff
@@ -46,7 +46,6 @@ type Options struct {
 	VarPath              string
 	ResourcePath         string
 	Zookeepers           []string
-	RemoteZookeepers     []string
```
smousa: Should this be called JWTTTL?  I feel like it may make sense to genericize it further by calling it AuthTTL or TokenTTL

smousa:
I had to rebase your branch, because I rebased the feature branch, please pull latest if you are making changes

cli/cmd/cmd.go
```diff
@@ -94,6 +94,7 @@ func New(driver api.API, config ConfigReader) *ServicedCli {
 		cli.IntFlag{"max-rpc-clients", defaultOps.MaxRPCClients, "max number of rpc clients to an endpoint"},
 		cli.IntFlag{"rpc-dial-timeout", defaultOps.RPCDialTimeout, "timeout for creating rpc connections"},
 		cli.IntFlag{"snapshot-ttl", defaultOps.SnapshotTTL, "snapshot TTL in hours, 0 to disable"},
+		cli.IntFlag{"jwt-ttl", defaultOps.JWTTTL, "JWT TTL in seconds past issued-at-time"},
```
smousa: rename to auth-expiration

tgjonestx:
Closing to replace with new PR after rebaseing

PR #1763, opened 2015-06-22
CC-1034: Add maintenance scripts to the install and have them executed by cron

iancmcc:
https://jira.zenoss.com/browse/CC-1034


pkg/serviced-container-cleanup
```diff
@@ -0,0 +1,31 @@
...
+
+if [ $# -eq 0 ]; then
+    DOCKER_LIB="/var/lib/docker"
+else
+    DOCKER_LIB="$1"
+fi
+
+if [ -n "$(df --type=btrfs $DOCKER_LIB 2>/dev/null)" ]; then
+    for LAYER in $(comm -23 <(ls $DOCKER_LIB/btrfs/subvolumes | grep -v '\init' | sort | uniq) <(cat <(docker images --no-trunc -qa) <(docker ps --no-trunc -qa) | sort | uniq)); do
+        btrfs subvolume delete $DOCKER_LIB/btrfs/subvolumes/$LAYER &>/dev/null
```
tgjonestx: I don't know what kind of range of values we get for `LAYER`.  If it might contain spaces or slashes or other control characters, maybe it would be safer to use `"$LAYER"`

iancmcc: It's an alphanumeric string, but no harm quoting it.

PR #1773, opened 2015-06-24
CC-996: Add new unit test, add missing function comment

kwalkerz:


wplu:
ship with test build success

PR #1777, opened 2015-06-26
Break service migration script execution out into its own package.

wwwtyro:


migrate/migrate.go
```diff
@@ -0,0 +1,251 @@
...
+)
+
+const (
+	// The mount point in the service migration docker image
+	MIGRATION_MOUNT_POINT = "/migration"
+	// The well-known path within the service's docker image of the directory which contains the service's migration script
+	EMBEDDED_MIGRATION_DIRECTORY = "/opt/serviced/migration"
+)
+
+func RunMigrationScript(ctx datastore.Context, fcd *facade.Facade, request dao.RunMigrationScriptRequest) error {
```
smousa: For ease of testability:

Instead of passing in the facade, create an interface that does only the requests you need.  For example, see https://github.com/control-center/serviced/blob/feature/hllb-release/scheduler/localsync.go

tgjonestx: +1 for unit-tests and injecting mocks

PR #1778, opened 2015-06-26
Feature/ui tests

racheloei:
Sorting the hosts by memory doesn't work in the UI, so that test is failing.


acceptance/runUIAcceptance.sh
```diff
@@ -15,6 +15,7 @@ debug=false
 runAsRoot=false
 DRIVER_NAME=selenium
 TIMEOUT=10
+# CUCUMBER_OPTS="--tags @run"
```
tgjonestx: Please remove this line.

acceptance/ui/features/hosts.feature
```diff
@@ -18,15 +17,204 @@ Feature: Host Management
...
-      And I fill in the RAM Commitment field with "100%"
+      And I fill in the RAM Commitment field with "50%"
       And I click "Add Host"
-    Then I should see "gjones-dev"
-      And I should see "default"
-      And I should see "Delete"
+    Then I should see "Error"
+      And I should see "Internal Server Error: host already exists"
+
+  @run @login-required @defaultHostPage
```
tgjonestx: Please remove `@run`

acceptance/ui/features/support/hooks.rb
```diff
@@ -7,3 +7,18 @@
   fillInDefaultPassword()
   clickSignInButton()
 end
+
+Before('@emptyHostsPage') do
```
tgjonestx: Functionally, this use of tags totally works, but I think the tests would read a little better if we converted these hooks to Given steps.  Also, using steps in cases like this would keep more of the feature-related code in the same file (e.g. steps/hosts.rb), making it a little easier for maintenance.

For instance, instead of tagging some Scenarios with @emptyHostsPage, we could start those Scenarios with something like `Given there are no hosts defined`.  Similarly, @defaultHostPage, could be replaced with a step like `Given the default host is defined`

acceptance/ui/features/hosts.feature
```diff
@@ -18,15 +17,204 @@ Feature: Host Management
...
+  @login-required @emptyHostsPage
+  Scenario: Add an invalid host with an invalid name
+    When I am on the hosts page
+      And I click the Add-Host button
+      And I fill in the Host Name field with "bogushost"
+      And I fill in the Resource Pool field with the default resource pool
+      And I fill in the RAM Commitment field with the default RAM commitment
+      And I click "Add Host"
+    Then I should see "Error"
+      And I should see "Bad Request"
```
tgjonestx: Wow - our error feedback is terrible, but that's not a problem with this test :-)

tgjonestx: Please add another step here, and in other tests for negative input, to verify that we highlight the bad field with a red-border. The step should be something like `And the Host Name field should be flagged as invalid` (or whatever wording you think makes the most sense).  The red-border is created by the CSS class `ng-invalid`

acceptance/ui/features/hosts.feature
```diff
@@ -18,15 +17,204 @@ Feature: Host Management
...
+      And I should see "No Data Found"
```
tgjonestx: Do you think the test would be more readable if instead of checking for "No Data Found" and "Showing 0 Results", we had a step like `And I should see an empty Hosts page`

acceptance/ui/features/hosts.feature
```diff
@@ -18,15 +17,204 @@ Feature: Host Management
...
+
+  @login-required @emptyHostsPage
+  Scenario: Add an invalid host with an invalid Resource Pool field
+    When I am on the hosts page
+      And I click the Add-Host button
+      And I fill in the Host Name field with the default host name
+      And I fill in the RAM Commitment field with the default RAM commitment
+      And I click "Add Host"
+    Then I should see "Error"
+      And I should see "Bad Request"
```
tgjonestx: And verify that the resource pool field is highlighted as an error

acceptance/ui/features/hosts.feature
```diff
@@ -18,15 +17,204 @@ Feature: Host Management
...
+  @login-required @emptyHostsPage
+  Scenario: Add an invalid host with an invalid RAM Commitment field
+    When I am on the hosts page
+      And I click the Add-Host button
+      And I fill in the Host Name field with the default host name
+      And I fill in the Resource Pool field with the default resource pool
+      And I fill in the RAM Commitment field with "invalidentry"
+      And I click "Add Host"
+    Then I should see "Error"
+      And I should see "Bad Request"
```
tgjonestx: And verify the RAM commitment field is highlighted as an error

acceptance/ui/features/hosts.feature
```diff
@@ -18,15 +17,204 @@ Feature: Host Management
...
+      And I fill in the Resource Pool field with the default resource pool
+      And I fill in the RAM Commitment field with the default RAM commitment
+      And I click "Add Host"
+    Then I should see "Error"
+      And I should see "Bad Request"
+      And I should see "No Data Found"
+      And I should see "Showing 0 Results"
+
+  @login-required @emptyHostsPage
+  Scenario: Add an invalid host with an invalid Resource Pool field
```
tgjonestx: should be "Add a valid host ..."

acceptance/ui/features/hosts.feature
```diff
@@ -18,15 +17,204 @@ Feature: Host Management
...
+      And I should see "Host and port"
+      And I should see the Host and port field
       And I should see "Resource Pool ID"
-    When I fill in the Host Name field with "gjones-dev:4979"
+      And I should see the Resource Pool ID field
+      And I should see "RAM Commitment"
+      And I should see the RAM Commitment field
+
+  @login-required @emptyHostsPage
+  Scenario: Add an invalid host with an invalid name
```
tgjonestx: Please add another Scenario for a similar, but slightly different case, and that is a valid hostname (something that's in the host table) and port number, but the address/port combination does not match a running instance of CC.  This case will also fail, but the error message is different.   For instance, in your tests, vagrant:4979 is a valid CC host, so try something like vagrant:9999 - in this case, the hostname "vagrant" is a valid host name (it's in /etc/hosts), but there is no CC instance running on that host at port 9999

acceptance/ui/features/steps/hosts_steps.rb
```diff
@@ -2,6 +2,18 @@
     visitHostsPage()
 end
 
+When(/^I fill in the Host Name field with the default host name$/) do
+    fillInDefaultHostAndPort()
```
tgjonestx: Let's refactor this a bit to apply the [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself) principle w.r.t. the page elements, because that will make maintenance a little easier if we ever rearrange the add/edit host dialog.

So instead of `fillInDefaultHostAndPort()` call a more general function and pass it the value you want to use for default - something like `fillInHostAndPort("172.17.42.1:4979")`, that line 18 use  `fillInHostAndPort(hostName)`. The point is to encapsulate the call to `@hosts_page.hostName_input.set` in a single method.

acceptance/ui/features/steps/hosts_steps.rb
```diff
@@ -2,6 +2,18 @@
...
+end
+
+When(/^I fill in the Resource Pool field with the default resource pool$/) do
+    fillInDefaultResourcePool()
+end
+
+When(/^I fill in the RAM Commitment field with the default RAM commitment$/) do
+    fillInDefaultRAMCommitment()
```
tgjonestx: Similar to last comment, use `fillInRAMCommitment("50%")` instead

acceptance/ui/features/steps/hosts_steps.rb
```diff
@@ -2,6 +2,18 @@
     visitHostsPage()
 end
 
+When(/^I fill in the Host Name field with the default host name$/) do
+    fillInDefaultHostAndPort()
+end
+
+When(/^I fill in the Resource Pool field with the default resource pool$/) do
+    fillInDefaultResourcePool()
```
tgjonestx: Similar to last comment, use `fillInResourcePool("default")` instead, and refactor line 22 to call `fillInResourcePool(resourcePool)`

acceptance/ui/features/steps/hosts_steps.rb
```diff
@@ -31,6 +72,27 @@ def visitHostsPage()
...
+    @hosts_page.ramCommitment_input.set "50%"
+end
+
+def removeAllHosts()
+    Capybara.match=:first
+    while page.all("[ng-repeat='host in $data']").size != 0 do
+      click_link_or_button("Delete")
+      click_link_or_button("Remove Host")
+    end
+    Capybara.match=:smart
```
tgjonestx: A downside of using third-party libraries is getting burned when their behavior changes with a new release. So we want to insulate ourselves as much as possible from assumptions that might change. In this case it would be safer to not assume the default value of Capybara.match is :smart. We can do that by saving the current value before we do anything, and restoring the saved value when we're done:

```
    defaultMatch = Capybara.match
    Capybara.match = :first
    while page.all("[ng-repeat='host in $data']").size != 0 do
      click_link_or_button("Delete")
      click_link_or_button("Remove Host")
    end
    Capybara.match = defaultMatch
```

acceptance/ui/features/hosts.feature
```diff
@@ -18,15 +17,204 @@ Feature: Host Management
...
+      And I fill in the Host Name field with "172.17.42.1:4979"
       And I fill in the Resource Pool field with "default"
-      And I fill in the RAM Commitment field with "100%"
+      And I fill in the RAM Commitment field with "50%"
       And I click "Add Host"
-    Then I should see "gjones-dev"
-      And I should see "default"
-      And I should see "Delete"
+    Then I should see "Error"
+      And I should see "Internal Server Error: host already exists"
```
tgjonestx: Please add a verification step like `And I should see an empty Hosts page` as mentioned above.

acceptance/ui/features/steps/hosts_steps.rb
```diff
@@ -15,9 +27,38 @@
 end
 
 When /^I click the Add-Host button$/ do
-    clickAddHostsButton()
+    clickAddHostButton()
+end
+
+Then (/^the "Active" column should be sorted with active hosts on (top|the bottom)$/) do |order|
+  list = page.all("[ng-if$='host.active']")
```
tgjonestx: Did you try this with hosts that were NOT active?  I suspect it won't work. If you take a look at [view-hosts.html](../../../../web/ui/src/Hosts/view-hosts.html), you'll see that there are 2 separate `<div>`s for active vs inactive.  My guess is that selector expression here will only match the active ones.  Off the top of my head, I'm not sure how to fix this, though I'm sure with some experimentation there's a way to do it.  My Ruby knowledge is pretty slim, but maybe something like this:
```
   list = page.all("[ng-if$='host.active']")
   list += page.all("[ng-if$='!host.active']")
```
If this turns out to be hard to fix, don't get stuck on it. We can leave it as a known issue and circle back to fix it later.

acceptance/ui/features/steps/generic_steps.rb
```diff
@@ -1,15 +1,70 @@
...
+  find(element).visible?
+end
+
+Then (/^I should see "(.*?)" in the "([^"]*)" column$/) do |text, column|
+  # attribute that includes name of column of all table cells
+  list = page.all("td[data-title-text='#{column}']")
+  list.include?(:text => text)
+end
+
+Then (/^the "([^"]*)" column should be sorted in ([^"]*) order$/) do |category, order|
```
tgjonestx: Nice

acceptance/ui/features/pools.feature
```diff
@@ -0,0 +1,113 @@
...
+      And I should see the Description field
+
+  @login-required
+  Scenario: Add a resource pool with a duplicate name
+    When I am on the resource pool page
+      And I click the add Resource Pool button
+      And I fill in the Resource Pool name field with "default"
+      And I fill in the Description field with "none"
+      And I click "Add Resource Pool"
+    Then I should see "Adding pool failed"
```
tgjonestx: Please add
```
And I should see "resource pool exists"
```

acceptance/ui/features/pools.feature
```diff
@@ -0,0 +1,113 @@
...
+    When I am on the resource pool page
+      And I click the add Resource Pool button
+    Then I should see "Add Resource Pool"
+      And I should see "Resource Pool: "
+      And I should see the Resource Pool name field
+      And I should see "Description: "
+      And I should see the Description field
+
+  @login-required
+  Scenario: Add a resource pool with a duplicate name
```
tgjonestx: Please add a test case for adding a pool with no name specified

acceptance/ui/features/pools.feature
```diff
@@ -0,0 +1,113 @@
...
+      And I should see "Added resource pool"
+      And I should not see "Description:"
+
+  @login-required
+  Scenario: Delete a resource pool
+    When I am on the resource pool page
+      And I remove "test"
+    Then I should see "This action will permanently delete the resource pool"
+    When I click "Remove Pool"
+    Then I should see "Removed Pool"
```
tgjonestx: Please add some additional steps to verify that the pool is really deleted

acceptance/ui/features/pools.feature
```diff
@@ -0,0 +1,113 @@
...
+  @login-required
+  Scenario: Add a resource pool
+    When I am on the resource pool page
+      And I click the add Resource Pool button
+      And I fill in the Resource Pool name field with "test"
+      And I fill in the Description field with "test"
+      And I click "Add Resource Pool"
+    Then I should see "Added new Pool"
+      And I should see "Added resource pool"
+      And I should not see "Description:"
```
tgjonestx: Please add some steps to verify that the new pool is in the table

acceptance/ui/features/steps/user_steps.rb
```diff
@@ -0,0 +1,39 @@
...
+end
+
+When (/^I clear my messages$/) do
+  page.find("[ng-click='clearMessages()']").click()
+end
+
+When (/^I click on an unread message$/) do    # clicks on the first unread message
+  Capybara.match=:first
+  page.find("[class='message unreadMessage ng-scope']").click()
+  Capybara.match=:smart
```
tgjonestx: Ditto comments for [host_steps.rb](host_steps.rb) re:Capybara.match

acceptance/ui/features/steps/user_steps.rb
```diff
@@ -0,0 +1,39 @@
...
+
+Then /^I should see my messages$/ do
+  page.assert_selector("[ng-repeat='message in messages.messages track by message.id']")
+end
+
+Then /^I should not see any messages$/ do
+  page.assert_no_selector("[ng-repeat='message in messages.messages track by message.id']")
+end
+
+Then /^I should see a checkmark$/ do
```
tgjonestx: This step definition reads like something really generic, but the implementation is very specific to this page.
How about something like `Then I should see that the message is marked as read`

acceptance/ui/features/user.feature
```diff
@@ -0,0 +1,45 @@
...
+      And I clear my messages
+    Then I should not see any messages
+
+  @login-required
+  Scenario: Switch language to Spanish
+    When I view user details
+      And I switch the language to Spanish
+    Then I should see "Usuario"
+      And I should see "Mensajes"
+      And I should see "Servidores"
```
tgjonestx: Please add `And I should see "Borrar"`

acceptance/ui/features/user.feature
```diff
@@ -0,0 +1,45 @@
...
+
+  @login-required
+  Scenario: Switch language to English
+    When I view user details
+      And I switch the language to English
+    Then I should see "Username"
+      And I should see "Messages"
+      And I should see "Hosts"
```
tgjonestx: Please add `And I should see "Delete"`

acceptance/ui/features/user.feature
```diff
@@ -0,0 +1,45 @@
...
+  @login-required @defaultHostPage
+  Scenario: View all messages
+    When I view user details
+    Then I should see my messages
+
+  @login-required @defaultHostPage
+  Scenario: Mark messages as read
+    When I view user details
+      And I click on an unread message
+    Then I should see a checkmark
```
tgjonestx: Let's discuss this one. Ideally, the test would verify that the message clicked on was the message marked as read, but given we don't know what messages will be there, that's a little hard to verify

acceptance/ui/features/user.feature
```diff
@@ -0,0 +1,45 @@
...
+
+  @login-required
+  Scenario: View user details
+    When I view user details
+    Then I should see "User Details"
+      And I should see "Username: zenoss"
+      And I should see "Messages"
+      And I should see "Clear"
+
+  @login-required @defaultHostPage
```
tgjonestx: Please remove @defaultHostPage

acceptance/ui/features/applications.feature
```diff
@@ -0,0 +1,95 @@
+@apps
+Feature: Application Management
+  In order to use Control Center
+  As a CC admin user
+  I want to manage applications
+
+  @login-required
+  Scenario: View default applications page
+    When I am on the applications page
+    Then I should be on the applications page
```
tgjonestx: I think this scenario should be similar to the first scenarios for Hosts and Pools where there are a set of "I should see" steps to verify the basic table contents.

acceptance/ui/features/applications.feature
```diff
@@ -0,0 +1,95 @@
...
+
+  @login-required
+  Scenario: View application template dialog
+    When I am on the applications page
+      And I click the Add-Application Template button
+    Then I should see "Add Application Template"
+      And I should see "Application Template File:"
```
tgjonestx:
```
And I should see "No file chosen"
```

acceptance/ui/features/applications.feature
```diff
@@ -0,0 +1,95 @@
...
+
+  @login-required
+  Scenario: View services map
+    When I am on the applications page
+      And I click the Services Map button
+    Then I should not see "Deployment ID"
```
tgjonestx: We should be able to include some kind of positive verification for viewing the service map. At a minimum, every system should have a service map that contains a node named "Internal Services".  Instead of looking for just any label containing "Internal Services", we need to look for one inside the services map.  The services map is rooted under an `svg` element.

So maybe a step that says something like:
```
And I should see "Internal Services" in the Services Map
```

Conceptually, the implementation of that step would be something like find the `tspan` element which is a child of an `svg` element and the contents of the `tspan` element are "Internal Services" (or whatever label is specified by the caller of the step)

tgjonestx:
I'm done with my review of the first 5 commits.  This is really great first step towards filling out our automation.

acceptance/ui/features/hosts.feature
racheloei: The only field that gets highlighted as invalid is the host name field if a port isn't provided.

acceptance/ui/features/steps/hosts_steps.rb
racheloei: The selector matches active and inactive hosts, but I haven't tested any bad (red exclamation mark) hosts.

acceptance/ui/features/hosts.feature
tgjonestx: The dialog is supposed to have two behaviors for invalid input: (1) display an error message, and (2) highlight the input field that caused the error. I'd like the test case to verify both behaviors, which means we need another step here for something like `And the Host Name field should be flagged as invalid`

acceptance/ui/features/steps/hosts_steps.rb
tgjonestx: OK - we can revisit that issue later, once we're able to add several more hosts

acceptance/ui/features/applications.feature
```diff
@@ -0,0 +1,161 @@
...
+  Scenario: View default applications page
+    When I am on the applications page
+    Then I should see "Application Template"
+      And I should see "Deployment ID"
+      And I should see "Virtual Host Names"
+      And I should see "Internal Services"
+      And I should see "Description"
+      And I should see "Services Map"
+
+  @run @login-required
```
tgjonestx: Please remove @run since it's not used

acceptance/ui/features/applications.feature
```diff
@@ -0,0 +1,161 @@
...
+    When I select "default"
+      And I click "Next"
+    Then I should see "Zenoss.core has been configured for resource pool default"
+    When I fill in the Deployment ID field with "Hbase"
+      And I click "Deploy"
+    Then I should see "Pulling images"
+      And I should see "this may take awhile"
+      And I should see "Application deployed"
+
+  @login-required @emptyHostsPage
```
tgjonestx: Please remove `@emptyHostsPage` since it's not used

tgjonestx:
I'm getting the following messages when I run the tests on my local. Are you seeing these? It looks like there are some steps which are undefined.
```
69 scenarios (9 failed, 1 undefined, 59 passed)
378 steps (9 failed, 38 skipped, 3 undefined, 328 passed)
10m18.796s

You can implement step definitions for undefined steps with these snippets:

When(/^I fill in the Host and port field with "(.*?)"$/) do |arg1|
  pending # express the regexp above with the code you wish you had
end

When(/^I fill in the Resource Pool ID field with "(.*?)"$/) do |arg1|
  pending # express the regexp above with the code you wish you had
end

Then(/^I should be on the applications page$/) do
  pending # express the regexp above with the code you wish you had
end
```

acceptance/ui/features/applications.feature
```diff
@@ -0,0 +1,161 @@
...
+      And I click the Services Map button
+    Then I should see "Internal Services" in the Services Map
+
+  @login-required
+  Scenario: View application template dialog
+    When I am on the applications page
+      And I click the Add-Application Template button
+    Then I should see "Add Application Template"
+      And I should see "Application Template File:"
+      And I should see "No file chosen"
```
tgjonestx: When I ran this test on my machine, the dialog contained "No file selected" instead of "No file chosen". 
Did does it display on your test?

racheloei: "No file selected" comes from Firefox, "No file chosen" comes from Chrome. I'll remove that line.

acceptance/ui/features/applications.feature
racheloei: The Add Application tests don't work since Cucumber times out before the app is added. I'll remove them for now.

racheloei:
I fixed the "I should be on the application page" problem, but I don't see the other two messages. Those steps don't appear in the scenarios anymore.

tgjonestx:
FYI - The other two undefined steps where resolved with the changes to `applications.feature` in 377fddb

tgjonestx:
:+1:  - will merge as soon as we get a green build

tgjonestx:
Retest this

PR #1786, opened 2015-07-02
navbar references in SitePrism section

racheloei:


tgjonestx:
Thanks for separating the code-style changes from the functional changes - that really helps clarify things in the code review

PR #1787, opened 2015-07-02
CC-811: service endpoint validation

smousa:
https://jira.zenoss.com/browse/CC-811

### DIFF
--- a/dao/elasticsearch/controlplanedao_test.go
+++ b/dao/elasticsearch/controlplanedao_test.go
@@ -161,7 +161,8 @@ func (dt *DaoTest) TestDao_ValidateEndpoints(t *C) {
 		},
 	}
 	err := dt.Dao.AddService(svc, &id)
-	t.Assert(err, NotNil)
+	t.Check(err, NotNil)
+	t.Check(strings.Contains(err.Error(), "found duplicate endpoint name"), Equals, true)
 
 	// test 2: success
 	svc.Endpoints = []service.ServiceEndpoint{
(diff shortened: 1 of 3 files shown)


facade/service.go
```diff
@@ -1335,11 +1389,6 @@ func (f *Facade) verifyServiceForUpdate(ctx datastore.Context, svc *service.Serv
 		return err
 	}
 
-	err = svc.ValidEntity()
```
tgjonestx: Seems like it makes more sense to keep the call to `svc.ValidEntity()` here and remove the one at lines 1410-14142, because if the basic attributes of the service are bad, why do more extensive validation like child service, endpoints, etc?

smousa: sure

dao/elasticsearch/controlplanedao_test.go
```diff
@@ -145,6 +145,94 @@ func (dt *DaoTest) TearDownSuite(c *C) {
...
+		ParentServiceID: svc.ID,
+		PoolID:          svc.PoolID,
+		Launch:          svc.Launch,
+		DeploymentID:    svc.DeploymentID,
+		Endpoints: []service.ServiceEndpoint{
+			{EndpointDefinition: servicedefinition.EndpointDefinition{Name: "test_ep_1", Application: "test_ep_1", Purpose: "export"}},
+		},
+	}
+	err = dt.Dao.AddService(svc2, &id)
+	t.Assert(err, NotNil)
```
tgjonestx: Should we test the value of `err` to verify that the test case is failing the right reason?

smousa: that is a bit harder to verify :-\

PR #1796, opened 2015-07-08
Stop NFS during application rollback or delete

smousa:
https://jira.zenoss.com/browse/CC-1069
https://jira.zenoss.com/browse/CC-1070

### DIFF
--- a/dfs/snapshot.go
+++ b/dfs/snapshot.go
@@ -185,10 +185,19 @@ func (dfs *DistributedFilesystem) Rollback(snapshotID string, forceRestart bool)
 		return err
 	}
 
-	// rollback the dfs
-	glog.V(0).Infof("Performing rollback for %s (%s) using %s", tenant.Name, tenant.ID, snapshotID)
-	if err := snapshotVolume.Rollback(snapshotID); err != nil {
-		glog.Errorf("Error while trying to roll back to %s: %s", snapshotID, err)
+	err = func() error {
+		dfs.networkDriver.Stop()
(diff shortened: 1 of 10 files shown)


dfs/snapshot.go
```diff
@@ -185,10 +185,19 @@ func (dfs *DistributedFilesystem) Rollback(snapshotID string, forceRestart bool)
 		return err
 	}
 
-	// rollback the dfs
-	glog.V(0).Infof("Performing rollback for %s (%s) using %s", tenant.Name, tenant.ID, snapshotID)
-	if err := snapshotVolume.Rollback(snapshotID); err != nil {
-		glog.Errorf("Error while trying to roll back to %s: %s", snapshotID, err)
+	err = func() error {
+		dfs.networkDriver.Stop()
```
tgjonestx: We should check the error code from Stop(). If we can't stop NFS, we should either skip the rollback, or at least report that error.

dfs/snapshot.go
```diff
@@ -308,8 +317,17 @@ func (dfs *DistributedFilesystem) DeleteSnapshots(tenantID string) error {
 		glog.Errorf("Could not find the volume for service %s: %s", tenantID, err)
 		return err
 	}
-	if err := snapshotVolume.Unmount(); err != nil {
-		glog.Errorf("Could not unmount volume for service %s: %s", tenantID, err)
+
+	err = func() error {
+		dfs.networkDriver.Stop()
```
tgjonestx: ditto previous comment re: error checking return from Stop()

PR #1798, opened 2015-07-08
Externalize inputs into JSON files

racheloei:


acceptance/ui/features/steps/generic_steps.rb
```diff
@@ -122,4 +135,38 @@ def sortColumn(category, sortOrder)
     while categoryLink[:class] != order do
         categoryLink.click()
     end
+end
+
+def removeAllEntries()
```
tgjonestx: Nice generalization

acceptance/ui/features/steps/generic_steps.rb
```diff
@@ -122,4 +135,38 @@ def sortColumn(category, sortOrder)
...
+    end
+end
+
+def getTableValue(valueOrTableUrl)
+    if valueOrTableUrl.start_with?("table://") == false
+        return valueOrTableUrl
+    end
+    parsedUrl = valueOrTableUrl.split(/\W+/)
+    tableType = parsedUrl[1]
+    fileName = File.join(ENV["DATASET_FILES"], tableType + ".json")
```
tgjonestx: Instead of reading/parsing the files on every call to getTableValue(), can you refactor this a bit so that at startup (somewhere in env.rb), you read/parse DATASET_FILES/*.json into a some kind of singleton for `table`? 

This will save a little on I/O since we call getTableValue() so many times.
If you can not read or parse one of the files, then the test should exit with a fatal error.

acceptance/ui/features/steps/generic_steps.rb
```diff
@@ -122,4 +135,38 @@ def sortColumn(category, sortOrder)
...
+        end
+        click_link_or_button("Remove")
+    end
+end
+
+def getTableValue(valueOrTableUrl)
+    if valueOrTableUrl.start_with?("table://") == false
+        return valueOrTableUrl
+    end
+    parsedUrl = valueOrTableUrl.split(/\W+/)
```
tgjonestx: To be defensive here, please add a check for the length of parsedUrl. For now, a valid URL must have exactly 4 values - anything else is invalid.

acceptance/ui/features/steps/hosts_steps.rb
```diff
@@ -107,35 +92,30 @@ def visitHostsPage()
...
     clickAddHostButton()
     fillInHostAndPort(name)
     fillInResourcePool(pool)
     fillInRAMCommitment(commitment)
     click_link_or_button("Add Host")
+end
+
+def addDefaultHost()
+    addHost("table://hosts/defaultHost/nameAndPort", \
+        "table://hosts/defaultHost/pool", "table://hosts/defaultHost/commitment")
```
tgjonestx: style nit pick - please put "table://hosts/defaultHost/commitment" on a line by itself.  There's many schools of aesthetic-thought here, but the general convention here at Zenoss is to put additional arguments one/line if they don't all fit on the first line

PR #1800, opened 2015-07-09
mismatch of parameters vs develop

smousa:


jplouis:
Did the smoke test not catch this bug?

smousa:
maintenance build does not do smoke tests evidently

PR #1803, opened 2015-07-09
CC-994: Allow restart of individual service instance

kwalkerz:


smousa:
waiting for build to pass

PR #1805, opened 2015-07-10
CC-1050: Retry InitHostRegistry in case of blocking.

iancmcc:


zzk/service/hostregistry.go
```diff
@@ -77,11 +78,22 @@ type HostRegistryListener struct {
...
+		go func() {
+			done <- conn.CreateDir(hostregpath())
+		}()
+		select {
+		case err := <-done:
+			if err == client.ErrNodeExists {
+				return nil
+			}
+			return err
+		case <-time.After(5 * time.Second):
```
jplouis: What does it mean to retry forever?  what does that do to  zookeeper to keep callking createdir if it isn't returning?  Is there a case where it doesn't return?

zzk/service/hostregistry.go
```diff
@@ -77,11 +78,22 @@ type HostRegistryListener struct {
...
+			glog.Errorf("Could not create host registry directory within 5 seconds. Retrying.")
```
jplouis: What happens to zookeeper and go routines when you keep calling CreateDir?  Is there really a scenario where CreateDir is not returning?

zzk/service/hostregistry.go
```diff
@@ -77,11 +79,20 @@ type HostRegistryListener struct {
...
+		done <- conn.CreateDir(path)
+	}()
+	select {
+	case err := <-done:
+		if err == client.ErrNodeExists {
+			return nil
+		}
+		return err
+	case <-time.After(5 * time.Second):
+		return fmt.Errorf("Could not create %s within 5 seconds", path)
```
jplouis: What happens to zookeeper and go routines if you loop forever?  Is there a scenario where CreateDir is not returning?

zzk/service/hostregistry.go
iancmcc: Yes. A request is somehow lost, and we're blocked waiting on a response that will never come. Doing a serviced reload causes the request to go out again, and things proceed.

Calling CreateDir on a node that exists is a noop. In this case, the node exists (we see it in exhibitor), but CreateDir is definitely blocked (we see it in the goroutine dump).

PR #1816, opened 2015-07-17
Starting and using mock agents

racheloei:


acceptance/runUIAcceptance.sh
```diff
@@ -107,6 +109,8 @@ fi
 HOSTNAME=`hostname -s`
 HOST_IP=`hostname -i`
 
+trap 'docker rm -f ui_acceptance' INT
```
wwwtyro: Can you briefly explain this?

acceptance/ui/features/data/default/hosts.json
```diff
@@ -2,49 +2,54 @@
...
+      "memoryGB": "10.00 GB",
+      "ramGB": "4.80 GB",
       "cores": 4982,
       "kernelVersion": "Kernel Version test 2",
       "kernelRelease": "Kernel Release test 2",
       "ccRelease": "CC Release test 2",
       "pool": "default",
-      "commitment": "25%",
-      "outboundIP": "10.87.110.17"
+      "commitment": "48%",
```
wwwtyro: May I ask what the reasoning is behind the selection of these commitment numbers?

acceptance/ui/features/steps/generic_steps.rb
```diff
@@ -115,7 +115,7 @@ def checkColumn(text, column, present)
     cell = getTableValue(text)
     hasEntry = false
     for i in 0..(list.size - 1)
-        hasEntry = true if list[i].text == cell
+        hasEntry = true if list[i].text == cell.to_s
```
wwwtyro: What is this magic? Convert to string?

acceptance/makeMockAgents.sh
```diff
@@ -0,0 +1,13 @@
...
+trap 'kill $(jobs -p)' EXIT
+
+cd ..
+make mockAgent
+cd acceptance
+echo `mockAgent/mockAgent --config-file ui/features/data/default/hosts.json --host defaultHost` &
+echo `mockAgent/mockAgent --config-file ui/features/data/default/hosts.json --host host2` &
+echo `mockAgent/mockAgent --config-file ui/features/data/default/hosts.json --host host3` &
+
+while true; do echo "CTRL C to stop"; sleep 5m; done
```
wwwtyro: Just curious - what was the trick to getting this to exit properly when the mockAgent exits with an error?

racheloei: It doesn't exit if an mock agent exits with an error. For now, we're trusting the user to not start the test suite if the mock agents fail, since it's now a separate script.

acceptance/ui/features/steps/generic_steps.rb
racheloei: Convert to string, in case it's looking for the number of cores or some other int.

acceptance/ui/features/data/default/hosts.json
racheloei: No reasoning.

acceptance/runUIAcceptance.sh
racheloei: It closes the docker container if you ctrl c out of the test suite. Before, if the test suite didn't end normally, you would have to manually kill and remove the docker container.

acceptance/makeMockAgents.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -e
+trap 'kill $(jobs -p)' EXIT
+
+cd ..
+make mockAgent
```
tgjonestx: I think prefer to rename this script something like startMockAagents.sh and remove this make. Again, the motivation is separation of concerns and I draw an analogy to serviced. Building serviced is separate from starting serviced which is separate from running the test against serviced.   We can instructions for making mockAgent to the README.

tgjonestx:
:+1:  - will merge when builds finish

PR #1818, opened 2015-07-17
reallocate buffers and retry getgrgid_r and getpwdnam_r operations on…

morr-zenoss:
… ERANGE (buffer not large enough)

Fixes CC-1072. See  http://tomlee.co/2012/10/problems-with-large-linux-unix-groups-and-getgrgid_r-getgrnam_r/ for a description of the underlying problem.

### DIFF
--- a/web/cauth.c
+++ b/web/cauth.c
@@ -2,6 +2,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+#include <errno.h>
 #include <grp.h>
 #include <pwd.h>
 #include <security/pam_appl.h>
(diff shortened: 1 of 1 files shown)


morr-zenoss:
I tested this as follows:
* deploy Zenoss on a virtual machine, with OpenLDAP installed.
* Add 1000 users to a group
* Compile the C code on the command-line, as instructed in the comments in main, and copy to test machine
* Run compiled executable, with the large group as the fourth parameter. 

Verified failing with original code, verified fixed with new.

**Final check: **
built serviced with new code, copied to test machine and executed - able to log into cc.

PR #1819, opened 2015-07-17
Add support for timeouts to RPC calls.

wwwtyro:
By default, most RPC calls are set to no timeout. We'll need to nail down proper timeouts for all the things. To help identify RPC calls that are unexpectedly slow, a warning is printed every 10 seconds for outstanding RPC calls indicating how long they've been outstanding.

### DIFF
--- a/rpc/rpcutils/clients.go
+++ b/rpc/rpcutils/clients.go
@@ -71,6 +71,7 @@ func (rc *reconnectingClient) Call(serviceMethod string, args interface{}, reply
 		//release read lock and get write lock
 		rc.RUnlock()
 		rc.Lock()
+		// rc.connectAndSet is idempotent, so no concerns about multiple calls.
 		rpcClient, err = rc.connectAndSet()
 		//release write lock
 		rc.Unlock()
@@ -84,7 +85,6 @@ func (rc *reconnectingClient) Call(serviceMethod string, args interface{}, reply
 		c := make(chan error, 1)
 		go func() {
 			c <- rpcClient.Call(serviceMethod, args, reply)
-			rc.RUnlock()
 		}()
 		select {
 		case err = <-c:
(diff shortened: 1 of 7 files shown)


rpc/rpcutils/clients.go
```diff
@@ -22,7 +23,7 @@ func SetDialTimeout(timeout int) {
 
 type Client interface {
 	Close() error
-	Call(serviceMethod string, args interface{}, reply interface{}) error
+	Call(serviceMethod string, args interface{}, reply interface{}, timeout int64, timewarn bool) error
```
alapidas: Should this maybe be a ```uint``` instead?

rpc/rpcutils/clients.go
```diff
@@ -60,7 +61,7 @@ func (rc *reconnectingClient) Close() error {
 	return nil
 }
 
-func (rc *reconnectingClient) Call(serviceMethod string, args interface{}, reply interface{}) error {
+func (rc *reconnectingClient) Call(serviceMethod string, args interface{}, reply interface{}, timeout int64, timewarn bool) error {
```
alapidas: Same ```uint``` comment

rpc/rpcutils/clients.go
```diff
@@ -79,7 +81,28 @@ func (rc *reconnectingClient) Call(serviceMethod string, args interface{}, reply
...
+	c := make(chan error, 1)
+	go func() {
+		c <- rpcClient.Call(serviceMethod, args, reply)
+	}()
+	start := time.Now()
+Loop:
+	for {
+		select {
+		case err = <-c:
+			break Loop
```
alapidas: Maybe also log V2 here how long the call took?

rpc/rpcutils/clients.go
```diff
@@ -79,7 +81,28 @@ func (rc *reconnectingClient) Call(serviceMethod string, args interface{}, reply
...
+		case <-time.After(time.Duration(timeout) * time.Second):
+			err = fmt.Errorf("RPC call to %s timed out after %d seconds.", serviceMethod, timeout)
+			break Loop
+		case <-time.After(10 * time.Second):
+			if timewarn {
+				glog.Warningf("RPC call to %s has taken more than %ds.", serviceMethod, int(time.Since(start).Seconds()))
```
alapidas: The log message here should maybe have an identifier on it so that interleaved log messages for the same RPC call can be differentiated between

wwwtyro:
test this

wwwtyro:
test this

PR #1820, opened 2015-07-17
background docker push on tag

smousa:
https://jira.zenoss.com/browse/CC-1103

Single host will work immediately, but multi-host will take time because the images will need to push to the registry first.

### DIFF
--- a/facade/servicetemplate.go
+++ b/facade/servicetemplate.go
@@ -384,7 +384,7 @@ func setImageID(registry, tenantID string, svc *service.Service) error {
 		}
 		UpdateDeployTemplateStatus(svc.DeploymentID, "deploy_tagging_image|"+svc.Name)
 		// now tag the image
-		if _, err := image.Tag(imageID); err != nil {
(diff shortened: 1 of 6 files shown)


facade/servicetemplate.go
```diff
@@ -384,10 +384,12 @@ func setImageID(registry, tenantID string, svc *service.Service) error {
 		}
 		UpdateDeployTemplateStatus(svc.DeploymentID, "deploy_tagging_image|"+svc.Name)
 		// now tag the image
-		if _, err := image.Tag(imageID); err != nil {
+		if _, err := image.Tag(imageID, false); err != nil {
 			glog.Errorf("Could not add tag %s to image %s: %s", imageID, svc.ImageID, err)
 			return err
 		}
+		// Push the image in the background
+		go docker.PushImage(imageID)
```
wwwtyro: Any reason not to mark it true above and get rid of this line?

smousa: I want the behavior of tag and push to be consistent with commit and push

PR #1827, opened 2015-07-24
Prevent annoying ES purge errors from being displayed.

wwwtyro:


utils/docker_exec.go
```diff
@@ -45,7 +45,7 @@ func RunDockerExec(containerID string, bashcmd []string) ([]byte, error) {
 	thecmd := exec.Command(command[0], command[1:]...)
 	output, err := thecmd.CombinedOutput()
 	if err != nil {
-		glog.Errorf("Error running command:'%s' output: %s  error: %s\n", command, output, err)
+		glog.V(2).Infof("Error running command:'%s' output: %s  error: %s\n", command, output, err)
```
smousa: Please wrap this message as part of the error that gets sent back to the caller

PR #1838, opened 2015-07-30
Fix/cc 1122

wwwtyro:
Upgrades our docker lib to 1.7.1.

Needed to update the docker login check to be compatible with 1.7.1. Removed support for logging in with user/pass/email since we weren't using it.

Added the build/test flag "libdm_no_deferred_remove" so that we can build and test on platforms that can't resolve C.dm_task_deferred_remove.


wwwtyro:
test this

wwwtyro:
test this

PR #1847, opened 2015-08-03
Zenoss core deployment tests

racheloei:


acceptance/ui/features/support/hooks.rb
```diff
@@ -21,7 +21,7 @@
 
 After('@clean_templates') do
     visitApplicationsPage()
-    removeAllEntries("template")
+    #removeAllEntries("template") until templates can be added back
```
tgjonestx: What's the problem here?

racheloei: I don't know how to add application templates

tgjonestx: I think that problem has been fixed. Let's delete that commented line

PR #1848, opened 2015-08-03
Update pkg docker version

wwwtyro:


pkg/makefile
```diff
@@ -164,7 +164,7 @@ deb: stage_deb
 		-d libdevmapper \
 		-d libdevmapper-event \
 		-d nfs-common \
-		-d 'lxc-docker-1.6.0' \
+		-d 'lxc-docker-1.7.1' \
```
tgjonestx: Hmm ... @iancmcc suggested I use https://blog.docker.com/2015/07/new-apt-and-yum-repos/ to update my Ubuntu box. The package from there is called `docker-engine`.

PR #1851, opened 2015-08-04
CC-1130: Protect channel variable with mutex

kwalkerz:
We are setting and clearing the channel object IService.exited multiple
times during the lifetime of an IService.  A build smoke test reported
a "DATA RACE" once where one goproc was setting it to nil while another
goproc was checking whether it was nil.


alapidas:
Looks good

PR #1853, opened 2015-08-04
CC-1130: Use RWMutex to protect channel variable

kwalkerz:
Better pattern to follow (rather than straight Mutex) for slight
performance improvement.


kwalkerz:
retest this

kwalkerz:
retest this

PR #1854, opened 2015-08-05
ZEN-19062: Support an explicit path to initialize a volume from

alapidas:
Tested an upgrade of this, and Zenoss started with no errors RE volumes.

### DIFF
--- a/node/agent.go
+++ b/node/agent.go
@@ -754,7 +754,11 @@ func (a *HostAgent) setupVolume(tenantID string, service *service.Service, volum
 		return "", fmt.Errorf("Could not get zk connection for resource path: %s, %s", resourcePath, err)
 	}
 
-	if err := createVolumeDir(conn, resourcePath, volume.ContainerPath, service.ImageID, volume.Owner, volume.Permission); err != nil {
+	containerPath := volume.InitContainerPath
+	if len(containerPath) == 0 {
+		containerPath = volume.ContainerPath
+	}
+	if err := createVolumeDir(conn, resourcePath, containerPath, service.ImageID, volume.Owner, volume.Permission); err != nil {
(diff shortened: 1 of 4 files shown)


node/agent.go
```diff
@@ -754,7 +754,11 @@ func (a *HostAgent) setupVolume(tenantID string, service *service.Service, volum
 		return "", fmt.Errorf("Could not get zk connection for resource path: %s, %s", resourcePath, err)
 	}
 
-	if err := createVolumeDir(conn, resourcePath, volume.ContainerPath, service.ImageID, volume.Owner, volume.Permission); err != nil {
+	containerPath := volume.InitContainerPath
+	if len(containerPath) == 0 {
```
jplouis: check for actual content, not just length 0

node/agent.go
```diff
@@ -754,7 +754,11 @@ func (a *HostAgent) setupVolume(tenantID string, service *service.Service, volum
 		return "", fmt.Errorf("Could not get zk connection for resource path: %s, %s", resourcePath, err)
 	}
 
-	if err := createVolumeDir(conn, resourcePath, volume.ContainerPath, service.ImageID, volume.Owner, volume.Permission); err != nil {
+	containerPath := volume.InitContainerPath
+	if len(strings.TrimSpace(containerPath)) == 0 {
+		containerPath = volume.ContainerPath
+	}
+	if err := createVolumeDir(conn, resourcePath, containerPath, service.ImageID, volume.Owner, volume.Permission); err != nil {
 		glog.Errorf("Error populating resource path: %s with container path: %s, %v", resourcePath, volume.ContainerPath, err)
```
jplouis: volume.ContainerPath is wrong now

PR #1859, opened 2015-08-06
Feature/driver options

smousa:


smousa:
retest this

volume/devicemapper/devicemapper.go
```diff
@@ -19,12 +21,25 @@ const (
 	DriverName = "devicemapper"
 )
 
+var (
+	// DriverOptions are the serviced device mapper options
+	DriverOptions = []string{
+		"SERVICED_DM_THINPOOLDEV",
```
iancmcc: I would consider making these constant typed strings. We may need to reference them in another package, and it would be better below to use SERVICED_DM_BASESIZE rather than "SERVICED_DM_BASESIZE".

volume/devicemapper/devicemapper.go
```diff
@@ -38,9 +53,11 @@ type DeviceMapperVolume struct {
 }
 
 // Init initializes the devicemapper driver
-func Init(root string) (volume.Driver, error) {
+func Init(root string, args []string, options map[string]string) (volume.Driver, error) {
```
iancmcc: I still don't get why we're parsing this here rather than in daemon.go.

smousa: I think that each volume type should support its own argument parser in the least

PR #1863, opened 2015-08-07
Store zookeeper logs in the isvcs container

tgjonestx:
This PR does NOT fix CC-1140. Before this PR, zookeeper log spamming was written to ZK's stdout which was collected by Docker and recorded in `/var/lib/docker`. So an out-of-control ZK could potentially fill up /var/lib/docker.  

With the changes in this PR, the ZK for isvcs will write it's logs to files inside the isvc container. The log4j.properties for ZK is already setup to keep ten 10MB log files, so the log spam will be constrained to no more than 100MB. This is consistent with the other isvcs.

### DIFF
--- a/isvcs/zk.go
+++ b/isvcs/zk.go
@@ -38,10 +38,12 @@ var exhibitorPortBinding = portBinding{
 }
 
 var Zookeeper = IServiceDefinition{
-	Name:         "zookeeper",
-	Repo:         IMAGE_REPO,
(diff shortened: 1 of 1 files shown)


iancmcc:
My only concern here is that we will no longer have access to zk logs without attaching to the running container.

tgjonestx:
OK - I guess it makes more sense to address docker logging constraints in the 1.2 release

PR #1869, opened 2015-08-11
Fix mount path to include '/volumes' suffix

tgjonestx:
Fixes CC-1145

### DIFF
--- a/dfs/snapshot.go
+++ b/dfs/snapshot.go
@@ -17,6 +17,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"path"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


jplouis:
retest this

dfs/snapshot.go
```diff
@@ -324,9 +325,10 @@ func (dfs *DistributedFilesystem) DeleteSnapshot(snapshotID string) error {
 // DeleteSnapshots deletes all snapshots relating to a particular tenantID
 func (dfs *DistributedFilesystem) DeleteSnapshots(tenantID string) error {
 	// delete the snapshot subvolume
```
smousa: We should not be calling GetDriver here.

Instead, it should be dfs.GetVolume
and then if you need the driver you can call volume.Driver()

PR #1882, opened 2015-08-13
user test fixes for Chrome

racheloei:


tgjonestx:
Testing locally with Chrome, I'm getting intermittent test failures for the scenario "Clear Messages".   The screenshot captured by the test failure shows that there are indeed 3 messages still in the log ("Removed host: 4981", "Success: Added host", and "Error: Username/Password is invalid").  However, if I login as the same user, I see there there are zero messages.

PR #1887, opened 2015-08-14
Show error when failing to delete pool.

wwwtyro:


jhanson:
Rye is making a separate ticket to audit all of the createNotification calls

wwwtyro:
Talked to Matos - createNotification's purpose is to display notifications when a modal dialog is being displayed, so it does have a purpose and shouldn't be replaced wholesale with $notification.create. In this case, the modal was closed immediately upon submission, so $notification.create was the proper path.

PR #1890, opened 2015-08-14
Disable defaulted btrfs in /etc/serviced/default.

wwwtyro:


pkg/serviced.default
```diff
@@ -52,7 +52,7 @@
 # SERVICED_CERT_FILE=/etc/....
 
 # Set the driver type on the master for the distributed file system (rsync/btrfs)
-SERVICED_FS_TYPE=btrfs
+# SERVICED_FS_TYPE=btrfs
```
tgjonestx: nit-pick - 

Line 54 - for 1.1, the choices should be expanded to include devicemapper
Line 55 - for 1.1, if we want to start deprecating btrfs in favor of devicemapper, we should change our example to be devicemapper.

PR #1907, opened 2015-08-20
Feature/vhosts disable

jplouis:


web/ui/src/Services/ServiceDetailsController.js
```diff
@@ -246,6 +246,20 @@
...
+            resourcesFactory.enableVHost( vhost.ApplicationId, vhost.ServiceEndpoint, vhost.Name)
+                .error((data, status) => {
+                    $notification.create("Start Vhost failed", data.Detail).error();
+                });
+        };
+
+    $scope.clickVHostDisable = function(vhost){
+        resourcesFactory.disableVHost( vhost.ApplicationId, vhost.ServiceEndpoint, vhost.Name)
+            .error((data, status) => {
+                $notification.create("Stopt Vhost failed", data.Detail).error();
```
tgjonestx: typo - should be `Stop Vhost`

web/routes.go
```diff
@@ -82,6 +82,7 @@ func (sc *ServiceConfig) getRoutes() []rest.Route {
 		rest.Route{"GET", "/services/vhosts", gz(sc.authorizedClient(restGetVirtualHosts))},
 		rest.Route{"PUT", "/services/:serviceId/endpoint/:application/vhosts/*name", gz(sc.authorizedClient(restAddVirtualHost))},
 		rest.Route{"DELETE", "/services/:serviceId/endpoint/:application/vhosts/*name", gz(sc.authorizedClient(restRemoveVirtualHost))},
+		rest.Route{"POST", "/services/:serviceId/endpoint/:application/vhosts/:name/enable", gz(sc.authorizedClient(restVirtualHostEnable))},
```
tgjonestx: this is a nit-pick that we don't necessarily need to address, but strictly speaking we should not allow the user to create vhosts named "enable"

web/virtualhostresource.go
```diff
@@ -100,50 +100,100 @@ func restAddVirtualHost(w *rest.ResponseWriter, r *rest.Request, client *node.Co
...
+		restBadRequest(w, err)
 		return
 	}
+	glog.V(1).Infof("Enable VHOST with %s %#v", r.URL.Path, request)
 
-	err = service.RemoveVirtualHost(application, hostname)
+	service, application, vhostname, err := getVHostContext(r, client)
 	if err != nil {
-		glog.Errorf("Unexpected error removing vhost, %s, from service (%s): %v", hostname, serviceID, err)
+		restServerError(w, err)
```
tgjonestx: Our REST error codes in general are not that great, but this is another example where we're conflating unexpected server-side errors with user-input errors. If the URL is valid, but either the serviceid, application/endpoint name or vhost name is undefined, then that should really be a 404 (though we don't really return 404s today).  If the server encountered some other unexpected error, then a 500 (server-error) is the correct response.

Since the current REST backend treats all not-found conditions as 500, then we should probably leave this as is for now, but see note in EnableVirtualHost()

domain/service/service.go
```diff
@@ -308,6 +332,22 @@ func (s *Service) AddVirtualHost(application, vhostName string) error {
 	return fmt.Errorf("unable to find application %s in service: %s", application, s.Name)
 }
 
+// EnableVirtualHost enable or disable a virtual host for given service
+func (s *Service) EnableVirtualHost(application, vhostName string, enable bool) error {
```
tgjonestx: To help debug cases with invalid input, it would be helpful to return errors that identify undefined application and vhost names

tgjonestx:
Do we need a CLI method for enable/disable?

jplouis:
Most cli service changes are handled via serviced service edit.

PR #1912, opened 2015-08-24
Fix/zen 19139

wwwtyro:


wwwtyro:
test this

domain/service/json.go
```diff
@@ -15,10 +15,38 @@ package service
...
+	if len(s.Commands) > 0 {
+		s.Runs = nil
+		return nil
+	}
+	if len(s.Runs) > 0 {
+		s.Commands = make(map[string]domain.Command)
+		for k, v := range s.Runs {
+			s.Commands[k] = domain.Command{
+				Command:         v,
+				CommitOnSuccess: false,
```
tgjonestx: For backwards compatibility, shouldn't the value be `true`?

The use-case I'm thinking about is as follows - customer has 1.0.x/5.0.x installed, they upgrade to CC 1.1 but don't upgrade RM 5.0.x. The 5.0.x services should work like they did before (which was CommitOnSuccess=true, right?)

For RM 5.1 service definitions, we don't really need to worry about the value in this method because those services will explicitly define the right values on a case-by-case basis. 
Ditto for upgrades from RM 5.0.x to RM 5.1.

domain/service/service.go
```diff
@@ -86,6 +86,7 @@ type Service struct {
 	LogConfigs        []servicedefinition.LogConfig
 	Snapshot          servicedefinition.SnapshotCommands
 	Runs              map[string]string
```
tgjonestx: Should probably add a FIXME or TODO comment here to remind us that Runs should be removed in some future release. Not sure what that release should be, but eventually we should clean it up.

domain/servicedefinition/servicedefinition.go
```diff
@@ -165,6 +166,32 @@ func (p *HostPolicy) UnmarshalText(b []byte) error {
...
+	if len(s.Commands) > 0 {
+		s.Runs = nil
+		return nil
+	}
+	if len(s.Runs) > 0 {
+		s.Commands = make(map[string]domain.Command)
+		for k, v := range s.Runs {
+			s.Commands[k] = domain.Command{
+				Command:         v,
+				CommitOnSuccess: false,
```
tgjonestx: Ditto previous comment about default value

smoke.sh
```diff
@@ -196,6 +196,39 @@ test_service_run() {
     set +x
 }
 
+test_service_run_command() {
```
tgjonestx: :+1:

tgjonestx:
Finished my review. I think we'll need to merge this one first.

smoke.sh
wwwtyro: :)

domain/service/json.go
wwwtyro: Ed brought this up, too. I was under the apparently mistaken impression that we wouldn't have a user on 5.0.x and 1.1, so maybe that's the way to go. I did bring this up with Ian, and his off-the-cuff answer was to default to false, but maybe we should run your point by him. I'll bring it up Monday morning.

iancmcc: Glenn is right. We will certainly have users on 5.0.x and 1.1.

tgjonestx:
:+1:

PR #1917, opened 2015-08-28
Lots of HA fixes

smousa:


tgjonestx:
There are 9 other places in the code already that ignore IsExist errors from MkdirAll. This PR adds 7 more exceptions, leaving ~19.   Instead of doing the same fix in 15-30 places, should we create a utility method, something like this:

```
package utils
...
func MkdirAll(path string, perms int) err {
    if err := os.MkdirAll(path, perms); err != nil && !os.IsExist(err) {
       return err
   }
   return err
}
```

tgjonestx:
:+1:

PR #1921, opened 2015-08-31
porting some minor fixes to the service scheduler back into develop

smousa:
https://jira.zenoss.com/browse/CC-1212


tgjonestx:
Per our discussion, I think it would be good if we could get some unit-tests to verify the changes for CC-1212 considering how severe of an impact it had.

PR #1925, opened 2015-09-01
CC-1217: Services test for stale NFS file handles and commit suicide appropriately.

iancmcc:


container/controller.go
```diff
@@ -473,7 +474,51 @@ func (c *Controller) rpcHealthCheck() (chan struct{}, error) {
...
+	for _, minfo := range mounts {
+		if strings.HasPrefix(minfo.Fstype, "nfs") {
+			nfsMountPoints = append(nfsMountPoints, minfo.Mountpoint)
+		}
+	}
+	if len(nfsMountPoints) > 0 {
+		// Start polling
+		go func() {
+			for {
+				for _, mp := range nfsMountPoints {
```
alapidas: Do we actually need to check every NFS mount point here?

iancmcc: What would be an alternative?

iancmcc: By which I mean: what would be a case where an nfs mount would be stale in
a container and we wouldn't care?

PR #1931, opened 2015-09-02
Feature/cc 1222

smousa:
https://jira.zenoss.com/browse/CC-1222
Do not start the docker kernel when running serviced as a container controller.

### DIFF
--- a/cli/api/daemon.go
+++ b/cli/api/daemon.go
@@ -14,6 +14,7 @@
 package api
 
 import (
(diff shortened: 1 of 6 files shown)


cli/api/daemon.go
```diff
@@ -102,6 +103,10 @@ type daemon struct {
 	networkDriver    storage.StorageDriver
 }
 
+func init() {
```
iancmcc: Can you not put this in init()? I hate having this as a side effect of import.

PR #1933, opened 2015-09-03
Fixes CC-1201: Serviced stopped

kwalkerz:
Looks to be a race condition.  Summer pointed me to a change she had made in the 1.1 code to avoid it.

### DIFF
--- a/zzk/zzk.go
+++ b/zzk/zzk.go
@@ -221,7 +221,7 @@ func Listen(shutdown <-chan interface{}, ready chan<- error, conn client.Connect
 		select {
 		case e := <-event:
 			if e.Type == client.EventNodeDeleted {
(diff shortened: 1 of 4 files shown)


zzk/zzk.go
```diff
@@ -221,7 +221,7 @@ func Listen(shutdown <-chan interface{}, ready chan<- error, conn client.Connect
 		select {
 		case e := <-event:
 			if e.Type == client.EventNodeDeleted {
-				glog.V(1).Infof("Node %s has been removed; shutting down listener", l.GetPath())
+				glog.Infof("Node %s has been removed; shutting down listener", l.GetPath())
```
smousa: This might be too chatty, can you change this back to v1

PR #1934, opened 2015-09-03
Remove service instances when the host is not available

smousa:


kwalkerz:
ship with test build success

PR #1945, opened 2015-09-09
Prevent the use of . and # in pool ids

wwwtyro:


wwwtyro:
test this

facade/pool.go
```diff
@@ -52,6 +54,10 @@ type PoolIPs struct {
 
 // AddResourcePool adds a new resource pool
 func (f *Facade) AddResourcePool(ctx datastore.Context, entity *pool.ResourcePool) error {
+	if strings.ContainsAny(entity.ID, ".#") {
```
smousa: This needs to be in domain/pool/validation.go

PR #1947, opened 2015-09-10
'--es-startup-timeout' CLI flag not respected

johnlhamilton:
Fixes CC-672:
https://jira.zenoss.com/browse/CC-672

Added StartupTimeout member to IServiceDefinition that defaults to the WAIT_FOR_INITIAL_HEALTHCHECK constant.

Modified isvc initialization to apply the ESStartupTimeout option to both the elasticsearch_logstash and elasticsearch_serviced services.

### DIFF
--- a/cli/api/daemon.go
+++ b/cli/api/daemon.go
@@ -142,7 +142,7 @@ func (d *daemon) getEsClusterName(Type string) string {
 }
 
 func (d *daemon) startISVCS() {
-	isvcs.Init()
+	isvcs.Init(options.ESStartupTimeout)
 	isvcs.Mgr.SetVolumesDir(path.Join(options.VarPath, "isvcs"))
 	if err := isvcs.Mgr.SetConfigurationOption("elasticsearch-serviced", "cluster", d.getEsClusterName("elasticsearch-serviced")); err != nil {
 		glog.Fatalf("Could not set es-serviced option: %s", err)
--- a/cli/cmd/cmd.go
+++ b/cli/cmd/cmd.go
@@ -79,7 +79,7 @@ func New(driver api.API, config utils.ConfigReader) *ServicedCli {
 		cli.StringSliceFlag{"mount", convertToStringSlice(defaultOps.Mount), "bind mount: DOCKER_IMAGE,HOST_PATH[,CONTAINER_PATH]"},
 		cli.StringFlag{"fstype", string(defaultOps.FSType), "driver for underlying file system"},
 		cli.StringSliceFlag{"alias", convertToStringSlice(defaultOps.HostAliases), "list of aliases for this host, e.g., localhost"},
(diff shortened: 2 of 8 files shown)


tgjonestx:
@johnlhamilton - looks like some unit tests need updating.  From http://jenkins.zendev.org/view/Control%20Center/job/serviced-pullrequests/4123/console, 
```
Running integration tests ...
<< snip, snip >>
# github.com/control-center/serviced/dao/elasticsearch
dao/elasticsearch/controlplanedao_test.go:110: not enough arguments in call to isvcs.Init
FAIL	github.com/control-center/serviced/dao/elasticsearch [build failed]
```

FYI - you can run the unit and integration tests before pushing with "make test". That takes about 10 minutes. If you want to run tests for a single package, you use the instructions here: https://sites.google.com/a/zenoss.com/engineering/home/faq/run-tests-per-serviced-package

tgjonestx:
Three issues for me: 
 * (original sin) for the end-user, it is not clear what the units should be from the cli help text
 * the default timeout has changed
 * there are 2 different minimum checks

I think these changes effectively change the default wait time because of the way the CLI code handles that argument. If I'm reading the code right, the default has changed from 2 minutes (WAIT_FOR_INITIAL_HEALTHCHECK)  to 10 minutes (the 600 in the config.IntVal() call).

There is one minimum check in options.getDefaultESStartupTimeout() and different one in isvs/container.go; each changes the value to something different (30 seconds in the CLI case, and 120 seconds in the isvcs case).

Suggestions:
 1. We should change the registration text for es-startup-timeout in  cli/cmd/cmd.go to clarify that the units are in seconds.
 1. We should change the 600 literal in `config.IntVal("ES_STARTUP_TIMEOUT", 600)` to be something like isvcs.DEFAULT_ES_STARTUP_TIMEOUT (defined in isvcs/es.go) which is equal to isvcs.WAIT_FOR_INITIAL_HEALTHCHECK
 1. We should change the hard-coded 30 in `getDefaultESStartupTimeout` to isvcs.MIN_ES_STARTUP_TIMEOUT (defined in isvcs/es.go) which is equal to 30
 1. The new check for `if sd.StartupTimeout <= 0` in isvcs/container.go should set the value to MIN_ES_STARTUP_TIMEOUT

johnlhamilton:
@tgjonestx I initially did that, but then it is inconsistent with the timeout constants in container.go:  

WAIT_FOR_INITIAL_HEALTHCHECK
DEFAULT_HEALTHCHECK_INTERVAL
DEFAULT_HEALTHCHECK_TIMEOUT

These are all constants in the isvcs package that are time.Duration values. But I agree it is a little tough to follow this way.  How about I change the two constants in es.go to be values in seconds, and rename them to DEFAULT_ES_STARTUP_TIMEOUT_SECONDS and MIN_ES_STARTUP_TIMEOUT_SECONDS?

tgjonestx:
:+1:  for DEFAULT_ES_STARTUP_TIMEOUT_SECONDS and MIN_ES_STARTUP_TIMEOUT_SECONDS?

PR #1948, opened 2015-09-10
CC-1164: Isvcs recovery

smousa:
If logstash is not working, then we move the bad data into a tgz and then delete the existing directory and create it again.


smousa:
retest this

smousa:
retest this

PR #1949, opened 2015-09-10
Feature/cc 1261

tgjonestx:
Fixes CC-1261

Needed to support upgrades to 5.0.6/5.1 versions of RM which have a new service with a space in the name, "User Interface"


wwwtyro:
Approved by Rye. (tm)

PR #1950, opened 2015-09-10
Feature/cc 840

rDr4g0n:


web/ui/src/Services/servicesFactory.js
```diff
@@ -29,7 +29,7 @@
         angular.extend(newFactory, {
             // TODO - update list by application instead
             // of all services ever?
-            update: function(force){
+            update: function(force, skipUpdateInstances){
```
rDr4g0n: add flag to skip updating instances because for services map we want to make that request directly so we can wait on it.

PR #1953, opened 2015-09-10
Don't collect volume stats on agents.

wwwtyro:


smousa:
waiting for tests to pass

PR #1954, opened 2015-09-11
CC-1268: Pull necessary image if it doesn't already exist when doing shell/run

iancmcc:


smousa:
Waiting for tests to pass

PR #1956, opened 2015-09-11
Helps CC-1248 Panic when no network configured

kwalkerz:
There were several panics in the cli code, and some were early enough to
prevent simple subcommands (like 'help' and 'version') from running. I
changed them to log instead (often Fatal, so serviced would shut down)
so we don't have the stack traces, which are bothersome to customers
(and they didn't seem like they would be helpful for the problems that
were being detected anyway).

I also found that starting the server ignored some errors that were
returned--we now log those as Fatal errors (so we shut down).

### DIFF
--- a/cli/cmd/options.go
+++ b/cli/cmd/options.go
@@ -100,7 +100,6 @@ func getDefaultEndpoint(ip, port string) string {
 	if ip == "" {
 		var err error
 		if ip, err = utils.GetIPAddress(); err != nil {
-			//fmt.Fprintf(os.Stderr, "ERROR: Unable to get default endpoint: %s\n", err)
 			return ""
 		}
 	}
--- a/cli/cmd/options.go
+++ b/cli/cmd/options.go
@@ -100,7 +100,8 @@ func getDefaultEndpoint(ip, port string) string {
 	if ip == "" {
 		var err error
 		if ip, err = utils.GetIPAddress(); err != nil {
-			panic(err)
+			//fmt.Fprintf(os.Stderr, "ERROR: Unable to get default endpoint: %s\n", err)
+			return ""
 		}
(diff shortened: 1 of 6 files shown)


cli/cmd/options.go
```diff
@@ -100,7 +100,8 @@ func getDefaultEndpoint(ip, port string) string {
 	if ip == "" {
 		var err error
 		if ip, err = utils.GetIPAddress(); err != nil {
-			panic(err)
+			//fmt.Fprintf(os.Stderr, "ERROR: Unable to get default endpoint: %s\n", err)
+			return ""
```
smousa: Should you fatal here as well?

cli/cmd/utils.go
```diff
@@ -35,7 +37,7 @@ func remove(index int, list ...interface{}) []interface{} {
 
 	switch {
 	case index < 0 || index > len(list):
-		panic("index out of bounds")
```
smousa: This function isn't called anywhere, can you just delete this?

cli/cmd/options.go
kwalkerz: I did not use Fatal here so simple commands like 'serviced version' and 'serviced help' would run.  This "" gets placed in options.Endpoint, then when the server actually tries to start up and discovers it doesn't have the info it needs, it Fatals out.  (Even printing the error message that I commented out is very distracting and annoying for those simple commands.)

smousa: ok, make sure you clean up your dead code then

cli/cmd/utils.go
kwalkerz: Even better--will do.

smousa:
waiting for tests to pass

PR #1959, opened 2015-09-14
Feature/cc 840

rDr4g0n:


acceptance/ui/features/applications.feature
```diff
@@ -18,10 +18,13 @@ Feature: Application Management
...
       And I should see "Services Map"
 
+  @fun
   Scenario: View services map
+    Given only the default host is added
+      And that the "table://applications/defaultApp/template" application with the "table://applications/defaultApp/id" Deployment ID is added
     When I am on the applications page
       And I click the Services Map button
-    Then I should see "Internal Services" in the Services Map
+    Then I should see "table://applications/defaultApp/template" in the Services Map
```
rDr4g0n: Internal services no longer appears in service map, so look for the test template

acceptance/ui/features/steps/applications_steps.rb
```diff
@@ -88,9 +88,9 @@
     end
 end
 
-Then (/^I should see "([^"]*)" in the Services Map$/) do |node|
+Then (/^I should see "(.*?)" in the Services Map$/) do |node|
     within(@servicesMap_page.map) do
-        assert_text(node)
+        assert_text(getTableValue(node))
```
rDr4g0n: look for variable value from json data

web/ui/src/Services/servicesFactory.js
```diff
@@ -29,7 +29,7 @@
         angular.extend(newFactory, {
             // TODO - update list by application instead
             // of all services ever?
-            update: function(force){
+            update: function(force, skipUpdateInstances){
```
rDr4g0n: updating instances can be pretty costly, so allow manual skip. specifically for service map, we want to manually update the instances so we can be notified when the update completes.

PR #1962, opened 2015-09-15
Feature/cc 1286

rDr4g0n:
DO NOT MERGE

### DIFF
--- a/web/ui/src/Hosts/HostDetailsController.js
+++ b/web/ui/src/Hosts/HostDetailsController.js
@@ -6,8 +6,8 @@
 (function() {
     'use strict';
 
-    controlplane.controller("HostDetailsController", ["$scope", "$routeParams", "$location", "resourcesFactory", "authService", "$modalService", "$translate", "miscUtils", "hostsFactory", "$notification", "instancesFactory", "servicesFactory",
-    function($scope, $routeParams, $location, resourcesFactory, authService, $modalService, $translate, utils, hostsFactory, $notification, instancesFactory, servicesFactory){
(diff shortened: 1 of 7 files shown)


web/ui/src/Hosts/HostDetailsController.js
```diff
@@ -75,8 +75,9 @@
         function init(){
             // start polling
             hostsFactory.activate();
-            servicesFactory.activate();
-            servicesFactory.update();
```
rDr4g0n: servicesFactory isnt needed here, just instances

web/ui/src/Hosts/HostsController.js
```diff
@@ -127,16 +127,19 @@
             // update hosts
             update();
 
-            servicesFactory.activate();
             hostsFactory.activate();
             poolsFactory.activate();
+            // need instances to roll up
+            // RAM commitment per host
+            instancesFactory.activate();
```
rDr4g0n: need instances so that hosts can rollup ram usage. Ideally the server would do this calculation.

web/ui/src/Instances/instancesFactory.js
```diff
@@ -16,7 +16,19 @@
         $notification = _notification;
         utils = _utils;
 
-        var newFactory = new BaseFactory(Instance, resourcesFactory.getRunningServices);
```
rDr4g0n: in order to reduce the size of the running instances request, try to only get running instances for a specific service or a specific host. Otherwise, make the much larger `getRunningServices` request.

PR #1963, opened 2015-09-15
CC-1196: Part I - Changing default paths

smousa:
* TODO: fix zendev
* TODO: need doc update for changes to default file
* Replaced SERVICED_VARPATH with ISVCS_PATH VOLUMES_PATH and BACKUPS_PATH
* Can customize what services to start with the isvcs-manager (if running as an agent)
* DFS changes for updated paths
* Created new option SERVICED_ISVCS_START, that takes a comma delimited string to specify which isvcs can run on the agent-only system.


smousa:
retest this

PR #1964, opened 2015-09-15
Fixes CC-1285: Leaking file descriptors over time

kwalkerz:
Per the documentation for net/http.Get():
  When err is nil, resp always contains a non-nil resp.Body.
  Caller should close resp.Body when done reading from it.
We weren't always doing that.


tgjonestx:
:+1:

PR #1966, opened 2015-09-16
serviced service status -h (but not really)

johnlhamilton:
Addresses CC-107

https://jira.zenoss.com/browse/CC-107

Serviced service status will now include healthcheck info by default.  No additional flag needed.  With no service specified, new field "HC Fail" will have an X for a service that has at least one failing healthcheck.  When a service is specified, two new fields, "Healthcheck" and "Healthcheck Status" will show the status of every healthcheck for the service.

### DIFF
--- a/cli/api/service.go
+++ b/cli/api/service.go
@@ -167,6 +167,34 @@ func (a *api) GetServiceStatus(serviceID string) (map[string]map[string]interfac
 				row["Cur/Max/Avg"] = fmt.Sprintf("--")
 
 				rowmap[fmt.Sprintf("%s/%d", svc.ID, stat.State.InstanceID)] = row
+
+				if(stat.Status == dao.Running && len(stat.HealthCheckStatuses) > 0){ 
+				
+					explicitFailure := false
+
+					for hcName, hcResult := range stat.HealthCheckStatuses{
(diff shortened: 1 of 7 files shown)


alapidas:
Please run this through gofmt

cli/api/service.go
```diff
@@ -167,6 +167,34 @@ func (a *api) GetServiceStatus(serviceID string) (map[string]map[string]interfac
 				row["Cur/Max/Avg"] = fmt.Sprintf("--")
 
 				rowmap[fmt.Sprintf("%s/%d", svc.ID, stat.State.InstanceID)] = row
+
+				if stat.Status == dao.Running && len(stat.HealthCheckStatuses) > 0 {
+
+					explicitFailure := false
+
+					for hcName, hcResult := range stat.HealthCheckStatuses {
+						newrow := make(map[string]interface{})
```
smousa: Why are you doing a map of interfaces here?

johnlhamilton: That is how the other rows were defined.  See lines 128 and 148.  Also, the API function is expected to return a map[string]map[string]interface{}

smousa: Ah, that's unfortunate, but ok

PR #1974, opened 2015-09-18
CC-1317: serviced wait for network to come up

kwalkerz:
In our systemctl unit definition we wait for network-online.target, but
that condition appears to be satisfied when any non-loopback interface
gets an IP address?  What we constantly see in our log is docker0 has an
IP address but the NIC card does not, and our code panics.

Instead of panicing here, we'll loop (every second) waiting to be able
to obtain an outgoing IP address.  Terminate after an arbitrary amount
of time (90 seconds).


kwalkerz:
retest this

kwalkerz:
retest this

cli/api/utils.go
```diff
@@ -41,11 +44,19 @@ func GetAgentIP(defaultRPCPort int) string {
...
 		return options.Endpoint
 	}
-	agentIP, err := utils.GetIPAddress()
-	if err != nil {
-		panic(err)
+
+	var agentIP string
+	var err error
+	for i := 1; i <= networkAccessAttempts; i++ {
+		if agentIP, err = utils.GetIPAddress(); err == nil {
```
wwwtyro: Does this work for offline installs? (Did it ever?)

jplouis: Look at line 44, I believe you need to set options when doing offline installs.

kwalkerz: If SERVICED_OUTBOUND_IP is set, then options.Endpoint will not be empty and offline install will work, right? Or is SERVICED_OUTBOUND_IP for something else?

jplouis: Ken, I think you are correct. My comment was to Rye.

wwwtyro: That's what Summer told me too, so sounds good.

kwalkerz: @JP--I suspected, but feeling very much out of the loop so trying to cover all bases. :-)

PR #1988, opened 2015-09-24
Fixes CC-922: Wait for network for outbound IP

kwalkerz:
When our appliances (possibly other hosts?) are rebooted, serviced often
starts running before IP addresses have been assigned to all interfaces.
So wait (up to 90 seconds) for one of the interfaces of the type we need
to become available.

Also noticed some confusion between the definition and use of ENDPOINT
and OUTBOUND_IP, so straightened those out.  (For Master, ENDPOINT is
optional and can fall back to OUTBOUND_IP or dynamic.  For Agent,
ENDPOINT is required.  When just parsing options, don't report any
error.  Wait to enforce the rules when we actually go to start the
daemon.)

### DIFF
--- a/cli/api/api.go
+++ b/cli/api/api.go
@@ -97,9 +97,14 @@ func LoadOptions(ops Options) {
 	}
 }
 
-// GetOptionsEndpoint returns the serviced RPC endpoint from options
+// GetOptionsRPCEndpoint returns the serviced RPC endpoint from options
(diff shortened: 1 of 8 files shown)


smousa:
Where you fmt.Errorf and propagate the error, I think it is better to glog with more information before propagating

for example:
```go
if err != nil {
    glog.Errorf("Got an error trying to do something: %s", err)
    return err
}
```

PR #2006, opened 2015-09-29
CC-1349: New CPU-aware scheduler

iancmcc:


scheduler/strategy.go
```diff
@@ -0,0 +1,135 @@
...
+
+	// Look up all running services for the hosts
+	glog.V(2).Infof("Looking up instances for hosts: %+v", hostids)
+	svcs, err := facade.GetRunningServicesForHosts(datastore.Get(), hostids...)
+	if err != nil {
+		return nil, err
+	}
+	// Assign the services to the StrategyHosts
+	for _, s := range svcs {
+		h := hostmap[s.HostID]
```
tgjonestx: Seems like there might be a null-pointer risk here if, for whatever reason, s.HostID was not one of hosts[]

scheduler/scheduler.go
```diff
@@ -30,7 +30,7 @@ import (
 	"path"
 )
 
-type leaderFunc func(<-chan interface{}, coordclient.Connection, dao.ControlPlane, string, int)
+type leaderFunc func(<-chan interface{}, coordclient.Connection, dao.ControlPlane, *facade.Facade, string, int)
```
smousa: why do we need connections to the dao and facade?  Is it possible to pick one here?

tgjonestx:
Retest this

scheduler/strategy/pack.go
```diff
@@ -0,0 +1,48 @@
...
+
+package strategy
+
+type PackStrategy struct{}
+
+func (s *PackStrategy) Name() string {
+	return "pack"
+}
+
+func (s *PackStrategy) SelectHost(service ServiceConfig, hosts []Host) (Host, error) {
```
wwwtyro: Does this ever return an error?

scheduler/strategy/score.go
```diff
@@ -0,0 +1,96 @@
...
+			usedMem += svc.RequestedMemoryBytes()
+			// Increment a counter of number of instances, for later strategies to use
+			if svc.GetServiceID() == service.GetServiceID() {
+				scoredHost.NumInstances += 1
+			}
+		}
+
+		// Calculate CPU score as a percentage of used cores on the host with this service deployed
+		if service.RequestedCorePercent() > 0 {
+			cpuScore = (usedCpu + service.RequestedCorePercent()) / totalCpu
```
tgjonestx: Shouldn't we check for totalCpu == 0 just in case?

scheduler/strategy/score.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+
+		// Calculate memory score as a percentage of used memory on the host with this service deployed
+		if service.RequestedMemoryBytes() > 0 {
+			memScore = int((usedMem + service.RequestedMemoryBytes()) * 100 / totalMem)
```
tgjonestx: Shouldn't we check for totalMem == 0 just in case?

scheduler/scheduler.go
iancmcc: We could have used the dao, but, if you remember, you said that was a terrible idea, which is why I added the GetRunningServicesForHosts facade method.

scheduler/strategy/score.go
smousa: will we ever accidently calculate a negative value?

scheduler/strategy/score.go
smousa: will we ever calculate a negative value?

scheduler/scheduler.go
iancmcc: Essentially this code should be moved off of dao to use facade, over time.

scheduler/strategy/score.go
wwwtyro: How accurate is this score, since we're doing integer division?

scheduler/strategy/score.go
wwwtyro: How accurate is this score, since we're doing integer division?

tgjonestx:
Retest this

scheduler/strategy/score.go
iancmcc: It doesn't need to be even close to float division accurate.

iancmcc: See, e.g.: https://github.com/docker/swarm/blob/master/scheduler/strategy/weighted_node.go

scheduler/strategy/require_separate.go
```diff
@@ -0,0 +1,40 @@
...
+		}
+	}
+	if over != nil && len(over) > 0 {
+		for _, h := range over {
+			if h.NumInstances == 0 {
+				return h.Host, nil
+			}
+		}
+	}
+	return nil, nil
```
tgjonestx: As I understand it, the previous impl allowed for the possibility that SelectHost() would return (nil, nil), but did that ever happen in practice?  In other words, a little concerned that this outcome might be easier to trigger some existing, but previously unused, code path in the scheduler.

scheduler/strategy/strategy.go
```diff
@@ -0,0 +1,65 @@
...
+type Strategy interface {
+	// The name of this strategy
+	Name() string
+	// Chooses the best host for svc to run on
+	SelectHost(svc ServiceConfig, hosts []Host) (Host, error)
+}
+
+func Get(name string) (Strategy, error) {
+	for _, strategy := range strategies {
+		if strategy.Name() == strings.ToLower(name) {
```
tgjonestx: It would be a little safe to wrap strategy.Name() in strings.ToLower() in case someone adds a strategy later without realizing what case they should use

scheduler/strategy/score.go
wwwtyro: I'm not so much concerned about being off by a fraction of a percent as I am about truncating to zero. Is there any reason not to use floats here?

tgjonestx:
:+1: - just waiting for the PR build to pass

PR #2022, opened 2015-10-05
Feature/cc 1251.1

smousa:
Part 1 includes:
* Creating domain objects
* Creating facade objects
* Setting up a mockable docker api
* Setting up a registry push listener
* Setting up a registry pull listener
* Setting up a syncing method (to be managed by the scheduler and can also replace service docker sync)

Each commit is independently verifiable and functional, for your convenience


smousa:
retest this

PR #2029, opened 2015-10-07
CC-1370: Incorporated the idea of snapshot tags on the volume

smousa:
This is the back half of the implementation.
Potentially you could do more, like add tags after creating a snapshot or removing a specific tag but preserving the volume, so I will leave that open-ended for now.  I am implementing this because I needed the SnapshotInfo bit for CC-1121 for better parsing of our snapshot labels.

### DIFF
--- a/volume/btrfs/btrfs.go
+++ b/volume/btrfs/btrfs.go
@@ -14,6 +14,7 @@
 package btrfs
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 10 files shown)


smousa:
retests this

smousa:
retest this

volume/btrfs/btrfs.go
```diff
@@ -317,8 +318,41 @@ func (v *BtrfsVolume) isSnapshot(rawLabel string) bool {
 	return strings.HasPrefix(rawLabel, v.getSnapshotPrefix())
 }
 
+// writeSnapshotInfo writes metadata about a snapshot
+func (v *BtrfsVolume) writeSnapshotInfo(label string, info *volume.SnapshotInfo) error {
```
tgjonestx: There's a bit of copy/paste code with the same impls of writeSnapshotInfo() and SnapshotInfo() across the different volumes.  I see that there is a lot of copy/paste code of generic methods like rawSnapshotLabel(), getSnapshotPrefix(), etc.

Seems like there should be some way to consolidate more of that code so it's not duplicated in so much.
In Java or C++, something like an AbstractBaseClass would work, but for GO you would have to make more of the methods public. Another alternative, is to consolidate all of the common code into something like a 'volume helper' class which is injected into each of Volume implementations.

tgjonestx: We discussed decided to defer this in favor of moving on to bigger/better changes in the DFS

PR #2030, opened 2015-10-07
CC-1251: Moving restore functions

smousa:
Originally, this was performed in the DFS--the idea here is to do this exclusively in the facade and either offer an interface to the dfs that calls these restores, or just leave this piece in the facade. (limiting the number of facade calls from the dfs)

### DIFF
--- a/facade/host.go
+++ b/facade/host.go
@@ -118,6 +118,42 @@ func (f *Facade) UpdateHost(ctx datastore.Context, entity *host.Host) error {
 	return err
 }
 
+// RestoreHosts restores a list of hosts, typically from a backup
+func (f *Facade) RestoreHosts(ctx datastore.Context, hosts []host.Host) error {
+	var exists bool
+	var err error
+
+	for _, host := range hosts {
+		host.DatabaseVersion = 0
+		// check all the static ips on this host
+		for _, ip := range host.IPs {
+			if exists, err = f.HasIP(ctx, host.PoolID, ip.IPAddress); err != nil {
+				glog.Errorf("Could no check ip %s in pool %s while restoring host %s: %s", ip.IPAddress, host.PoolID, ip.HostID, err)
+				return err
+			} else if exists {
+				glog.Warningf("Could not restore host %s (%s): ip already exists", ip.HostID, ip.IPAddress)
+				break
+			}
+		}
+		if !exists {
+			// check the primary ip on this host
+			if exists, err := f.HasIP(ctx, host.PoolID, host.IPAddr); err != nil {
+				glog.Errorf("Could not check ip %s in pool %s while restoring host %s: %s", host.IPAddr, host.PoolID, host.ID, err)
+				return err
+			} else if !exists {
+				if err := f.AddHost(ctx, &host); err != nil {
+					glog.Errorf("Could not add host %s to pool: %s", host.ID, host.PoolID, err)
+					return err
+				}
+				glog.Infof("Restored host %s (%s) to pool %s", host.ID, host.IPAddr, host.PoolID)
+			} else {
+				glog.Warningf("Could not restore host %s (%s): ip already exists", host.ID, host.IPAddr)
+			}
+		}
+	}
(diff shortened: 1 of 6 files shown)


smousa:
retest this

facade/host_test.go
```diff
@@ -93,6 +95,107 @@ func (s *FacadeTest) Test_HostCRUD(t *C) {
...
+	c.Assert(err, IsNil)
+	actual, err := s.Facade.GetHosts(s.CTX)
+	c.Assert(err, IsNil)
+	for i := range actual {
+		actual[i].DatabaseVersion = 0
+		actual[i].CreatedAt = time.Time{}
+		actual[i].UpdatedAt = time.Time{}
+	}
+	c.Assert(actual, DeepEquals, hosts1)
+	defer s.Facade.RemoveHost(s.CTX, "deadb11f")
```
tgjonestx: Please move to line 123 just to be safe

facade/host_test.go
```diff
@@ -93,6 +95,107 @@ func (s *FacadeTest) Test_HostCRUD(t *C) {
...
+				}, {
+					HostID:    "deadb11e",
+					IPAddress: "192.168.0.1",
+				},
+			},
+			CreatedAt: time.Time{},
+			UpdatedAt: time.Time{},
+		},
+	}
+	err = s.Facade.RestoreHosts(s.CTX, hosts2)
```
tgjonestx: The result should be a non-nil `err`, right?

facade/host.go
```diff
@@ -118,6 +118,42 @@ func (f *Facade) UpdateHost(ctx datastore.Context, entity *host.Host) error {
 	return err
 }
 
+// RestoreHosts restores a list of hosts, typically from a backup
+func (f *Facade) RestoreHosts(ctx datastore.Context, hosts []host.Host) error {
+	var exists bool
+	var err error
+
+	for _, host := range hosts {
+		host.DatabaseVersion = 0
```
tgjonestx: Seems like it makes more sense to move this before line 144 so we're only changing properties of this host object if we know we're about to restore it.

facade/pool_test.go
```diff
@@ -140,6 +140,79 @@ func (ft *FacadeTest) Test_RemoveResourcePool(t *C) {
...
+	c.Assert(err, IsNil)
+	actual, err := ft.Facade.GetResourcePools(ft.CTX)
+	c.Assert(err, IsNil)
+	for i := range actual {
+		actual[i].DatabaseVersion = 0
+		actual[i].CreatedAt = time.Time{}
+		actual[i].UpdatedAt = time.Time{}
+	}
+	c.Assert(actual, DeepEquals, pools1)
+	defer ft.Facade.RemoveResourcePool(ft.CTX, "testpool-1")
```
tgjonestx: Please move to line 161 just to be safe

facade/pool_test.go
```diff
@@ -140,6 +140,79 @@ func (ft *FacadeTest) Test_RemoveResourcePool(t *C) {
...
+	c.Assert(err, IsNil)
+	actual, err = ft.Facade.GetResourcePools(ft.CTX)
+	c.Assert(err, IsNil)
+	for i := range actual {
+		actual[i].DatabaseVersion = 0
+		actual[i].CreatedAt = time.Time{}
+		actual[i].UpdatedAt = time.Time{}
+	}
+	c.Assert(actual, DeepEquals, pools2)
+	defer ft.Facade.RemoveResourcePool(ft.CTX, "testpool-2")
```
tgjonestx: Please move to line 204 just to be safe

facade/host_test.go
smousa: fixed this below

facade/host.go
smousa: The objects in this array are pass by value, so it wouldn't matter anyway

tgjonestx:
:+1: Changes look good; just waiting for a green build

PR #2039, opened 2015-10-08
CC-1354: Retry sync() if resuming instances fails

kwalkerz:
sync() was not returning false in this case, so the caller was not
setting the retry timer.  Which led to the test case timing out.


zzk/service/service.go
```diff
@@ -267,6 +267,7 @@ func (l *ServiceListener) sync(svc *service.Service, rss []dao.RunningService) b
 		// resumeInstance updates the service state ONLY if it has a PAUSED DesiredState
 		if err := resumeInstance(l.conn, state.HostID, state.ID); err != nil {
 			glog.Warningf("Could not resume paused service instance %s (%s) for service %s on host %s: %s", state.ID, state.Name, state.ServiceID, state.HostID, err)
+			return false
```
tgjonestx: I don't know enough about this method to know if the following suggestion is a good idea or not, but we're in a loop here, so maybe we should try to resume all of the instances, and return false after the loop if one or more could not be resumed?

tgjonestx: Talked to Summer - the changes are fine, but Github says the code needs to be merged first

PR #2052, opened 2015-10-12
Feature/CC-1387

ed-:
Adds a pair of options to serviced to set the logging driver and options for docker containers. This allows us to set the rotation size and file count, for example, to limit use of space in /var/lib/docker

Closes CC-1387

### DIFF
--- a/dao/mocks/ControlPlane.go
+++ b/dao/mocks/ControlPlane.go
@@ -1,0 +1,729 @@
+// Copyright 2015 The Serviced Authors.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build unit integration
+
+package mocks
+
+import "github.com/control-center/serviced/dao"
+import "github.com/stretchr/testify/mock"
+
+import "github.com/control-center/serviced/domain"
+import "github.com/control-center/serviced/domain/addressassignment"
+import "github.com/control-center/serviced/domain/service"
+import "github.com/control-center/serviced/domain/servicestate"
+import "github.com/control-center/serviced/domain/servicetemplate"
+import "github.com/control-center/serviced/domain/user"
+import "github.com/control-center/serviced/metrics"
+import "github.com/control-center/serviced/volume"
+
+type ControlPlane struct {
+	mock.Mock
+}
+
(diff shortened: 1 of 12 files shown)


node/mocks/cp_client.go
```diff
@@ -0,0 +1,300 @@
...
+	"github.com/control-center/serviced/domain/user"
+	"github.com/control-center/serviced/metrics"
+	"github.com/control-center/serviced/volume"
+)
+
+type ControlClient struct {
+	addr      string
+	rpcClient string
+}
+
```
tgjonestx: Please add code like this which will trigger a compile error if the interface should change without an update to this class.
```
// assert the interface
var _ node.ControlClientInterface = &ControlClient{}
```

node/mocks/cp_client.go
```diff
@@ -0,0 +1,300 @@
+// Copyright 2014 The Serviced Authors.
```
tgjonestx: nit pick - should be 2015

node/mocks/cp_client.go
```diff
@@ -0,0 +1,300 @@
...
+	"github.com/control-center/serviced/domain/addressassignment"
+	"github.com/control-center/serviced/domain/service"
+	"github.com/control-center/serviced/domain/servicestate"
+	"github.com/control-center/serviced/domain/servicetemplate"
+	"github.com/control-center/serviced/domain/user"
+	"github.com/control-center/serviced/metrics"
+	"github.com/control-center/serviced/volume"
+)
+
+type ControlClient struct {
```
tgjonestx: Technically speaking you have built a mock here, but it has very limited utility because none of behavior can be controlled for testing purposes.  Please try this tool - https://github.com/vektra/mockery. It will generate a true mock for you - one that can use this [mocking package](https://github.com/stretchr/testify#mock-package). 

From the perspective of the unit-test, everything should work the same. The advantage is that anyone else who tries to use this mock in the future will have a full-feature mock at their disposal.

node/agent.go
```diff
@@ -464,7 +470,7 @@ func updateInstance(state *servicestate.ServiceState, ctr *docker.Container) err
 // configureContainer creates and populates two structures, a docker client Config and a docker client HostConfig structure
 // that are used to create and start a container respectively. The information used to populate the structures is pulled from
 // the service, serviceState, and conn values that are passed into configureContainer.
-func configureContainer(a *HostAgent, client *ControlClient,
+func configureContainer(a *HostAgent, client ControlClientInterface,
```
tgjonestx: I just realized that `node.ControlClient` is itself an implementation of another interface - `dao.ControlPlane`. So instead of creating a new interface in node, you can change this line to be `client dao.ControlPlane` instead of `client *ControlClient`.  And then you can move the mock to `dao/mocks`

dao/mocks/ControlPlane.go
```diff
@@ -0,0 +1,714 @@
+package mocks
```
ed-: I need a header here.

tgjonestx:
:+1: - will merge as soon as the build is good

PR #2055, opened 2015-10-13
Make runUIAcceptance.sh easier to use

tgjonestx:
1. Default to the chrome driver because it's twice as fast as Firefox
2. Add `--tags` command line option to make tag usage a little easier.


tgjonestx:
retest this

PR #2057, opened 2015-10-13
CC-1448: Collect disk statistics from hosts

alapidas:
This is a requirement for a Zenoss feature, and also it's good
information to collect.  control-center/go-procfs was updated to support
pulling data from ```/proc/diskstats``` as part of this


rphillipsz:
retest this

alapidas:
This is invalid - I can explain in more detail

PR #2065, opened 2015-10-15
CC-1438 /var/lib/docker partition ran out of space

ed-:
Port of #2052 to 1.1

Adds a pair of config/env flags to make use of Docker 1.8 logging options
to prevent /var/lib/docker from being overrun by services like zminion that log directly to stdout/stderr

Closes CC-1438

### DIFF
--- a/cli/api/api.go
+++ b/cli/api/api.go
@@ -82,6 +82,8 @@ type Options struct {
 	StartISVCS           []string          // ISVCS to start when running as an agent
 	IsvcsZKID            int               // Zookeeper server id when running as a quorum
 	IsvcsZKQuorum        []string          // Members of the zookeeper quorum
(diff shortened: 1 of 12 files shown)


tgjonestx:
Now I feel like a dummy for missing this in the review of #2052, but we have some redundant code here. dao/mocks/ControlPlane.go and dao/test/mockControlPlane.go are 2 different Mock impls for the same interface. 

AFAIK, dao/test/mockControlPlane.go is only used in cli/api/service_test.go.  Since I built it by hand when I barely knew GO, it's probably less-general than the one Mockery created, so I think we should keep the new one you created and retire  dao/test/mockControlPlane.go

ed-:
Can I convince you that we can defer deduplication to a later review? Rebasing in the meantime

PR #2068, opened 2015-10-20
Reregister the host.

wwwtyro:
An inconsistency in zookeeper state can occur when an agent temporarily loses its connection to zookeeper and ephemeral node and later regains its connection. This change makes the agent watch its ephemeral node, and when dropped, recreates it.

### DIFF
--- a/zzk/service/host.go
+++ b/zzk/service/host.go
@@ -82,6 +82,7 @@ type HostStateListener struct {
 	hostID   string
 	registry string
 	nodelock sync.Mutex
+	done     bool
 }
 
 // NewHostListener instantiates a HostListener object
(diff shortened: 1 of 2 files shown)


zzk/service/host.go
```diff
@@ -101,6 +103,8 @@ func (l *HostStateListener) GetPath(nodes ...string) string {
 
 // Ready adds an ephemeral node to the host registry
 func (l *HostStateListener) Ready() error {
+	l.nodelock.Lock()
```
jplouis: Since you are locking access to this node, do you want to lock the Done method as well?  What about the Read at 162?

wwwtyro: The reason I'm locking Ready() is because it can be called from multiple goroutines running the Spawn() function below (that's new with this change). There was no similar change for Done(), so I think we're safe there. As far as #162 goes, I think a lock should be unnecessary, since the Ready() operation is idempotent.

PR #2082, opened 2015-10-23
CC-1484: Create the backup directory at startup and make sure all errors bubbl…

smousa:
…e to the ui

### DIFF
--- a/cli/api/daemon.go
+++ b/cli/api/daemon.go
@@ -779,6 +779,9 @@ func (d *daemon) initDAO() (dao.ControlPlane, error) {
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


cli/api/daemon.go
```diff
@@ -779,6 +779,9 @@ func (d *daemon) initDAO() (dao.ControlPlane, error) {
 	if err != nil {
 		return nil, err
 	}
+	if err := os.MkdirAll(options.BackupsPath, 0777); err != nil && !os.IsExist(err) {
+		glog.Fatalf("Could not create default backup path at %s: %s", options.BackupsPath, err)
```
alapidas: is a ```Fatalf``` really appropriate?

smousa: yes, because we need a backup path configured in order to do backups

PR #2086, opened 2015-10-26
CC-1379: Bump isvcs to v38

smousa:


alapidas:
LGTM, merge when tests complete

PR #2092, opened 2015-10-28
Ignore ECHILD in distributed storage healthcheck.

wwwtyro:


tgjonestx:
I think this covers things functionality-wise, but it would be nice for future reference to include a comment that points to the SO post that explains the `stat` call.

tgjonestx:
:+1: just waiting for a green build

wwwtyro:
test this

PR #2098, opened 2015-10-28
Ignore ECHILD in distributed storage healthcheck.

wwwtyro:


tgjonestx:
Looks like this is targeted at the wrong branch for UCS - should be support/1.1.x

wwwtyro:
William asked for a port to 1.0.8; this is for that. The 1.1.x port is up, too.

PR #2106, opened 2015-10-30
Use lock instead of waitgroup when starting instances.

wwwtyro:


jplouis:
This will also panic

node/agent.go
```diff
@@ -383,19 +382,16 @@ func (a *HostAgent) StartService(svc *service.Service, state *servicestate.Servi
 		a.removeInstance(state.ID, ctr)
 	})
 
-	if expectDockerEvents, err := ctr.Start(); err != nil {
+	if err := ctr.Start(); err != nil {
 		glog.Errorf("Could not start service state %s (%s) for service %s (%s): %s", state.ID, ctr.ID, svc.Name, svc.ID, err)
-		if !expectDockerEvents {
```
jplouis: Why isn't the bool check needed anymore?

node/agent.go
```diff
@@ -383,19 +382,16 @@ func (a *HostAgent) StartService(svc *service.Service, state *servicestate.Servi
...
-			// The Docker event handlers we just registered won't get called, so perform those cleanup steps here
-			glog.Infof("Cleaning up after start failure for service state %s (%s) for service %s (%s)", state.ID, ctr.ID, svc.Name, svc.ID)
-			started.Done()
-			a.removeInstance(state.ID, ctr)
-		}
+		startBarrier <- struct{}{}
```
tgjonestx: Is this really necessary? 

Since we return at line 389, then we'll never read from `startBarrier` instance at line 394. Right?

PR #2116, opened 2015-10-30
Snapshot Tagging

johnlhamilton:
Added CLI for listing/adding/deleting tags on snapshots, as well as removing snapshots based on tags.
Modified TTL to not apply to snapshots that have any tags (these must now be manually deleted)
Modified script to optionally accept tag names when creating snapshots.

https://jira.zenoss.com/browse/CC-1370
https://jira.zenoss.com/browse/CC-836


cli/cmd/service.go
```diff
@@ -1309,13 +1314,36 @@ func (c *ServicedCli) cmdServiceListSnapshots(ctx *cli.Context) {
...
 	} else {
-		for _, s := range snapshots {
-			fmt.Println(s)
+		if showTags { //print a table of snapshot, description, tag list
+			t := NewTable("Snapshot,Description,Tags")
+			for _, s := range snapshots {
+				//build a comma-delimited list of the tags
+				var tags string
+
+				for _, tag := range s.Tags {
```
smousa: Use strings.Join here

cli/cmd/service_test.go
```diff
@@ -347,7 +347,15 @@ func (t ServiceAPITest) AddSnapshot(config api.SnapshotConfig) (string, error) {
 	if t.errs["AddSnapshot"] != nil {
 		return "", t.errs["AddSnapshot"]
 	}
-	return fmt.Sprintf("%s-snapshot description=%q", config.ServiceID, config.Message), nil
+
+	tagstr := ""
+
+	for _, tag := range config.Tags {
```
smousa: use strings.Join here

dao/elasticsearch/filesystem.go
```diff
@@ -335,3 +342,45 @@ func (dao *ControlPlaneDao) ReadyDFS(serviceID string, _ *int) (err error) {
 	err = dao.facade.ResetLock(ctx, serviceID)
 	return
 }
+
+// TagSnapshot tags an existing snapshot with 1 or more strings
+func (dao *ControlPlaneDao) TagSnapshot(request model.TagSnapshotRequest, newTagList *[]string) error {
+	ctx := datastore.Get()
+
+	// synchronize the dfs
```
smousa: Why do you think that we need to lock the DFS if we are adding a tag to an existing snapshot?

dfs/dfs.go
```diff
@@ -57,6 +57,12 @@ type DFS interface {
 	Restore(r io.Reader) (*BackupInfo, error)
 	// BackupInfo provides detailed info for a particular backup
 	BackupInfo(r io.Reader) (*BackupInfo, error)
+	// Tag adds one or more tags to an existing snapshot
+	Tag(snapshotID string, tagNames []string) ([]string, error)
+	// RemoveTags removes one or more tags from an existing snapshot
+	RemoveTags(snapshotID string, tagNames []string) ([]string, error)
```
smousa: I think that tags should be unique across all snapshots and not just unique to a particular snapshot

smousa: Also, I think we should have methods like Tag and Untag and they should be per tag and not really a bulk operation

dfs/dfs.go
```diff
@@ -57,6 +57,12 @@ type DFS interface {
...
+	// RemoveAllTags remvoes all tags from an existing snapshot
+	RemoveAllTags(snapshotID string) error
```
smousa: see above comment

dfs/ttl/snapshotttl.go
```diff
@@ -67,18 +67,21 @@ func (ttl *SnapshotTTL) Purge(age time.Duration) (time.Duration, error) {
...
-				if parts := strings.Split(s.SnapshotID, "_"); len(parts) == 2 {
-					if ts, err := time.Parse(timeFormat, parts[1]); err == nil {
-						if timeToLive := ts.Sub(expire); timeToLive <= 0 {
-							if err := ttl.client.DeleteSnapshot(s.SnapshotID, nil); err != nil {
-								glog.Errorf("Could not delete snapshot %s for tenant service %s (%s): %s", s.SnapshotID, svc.Name, svc.ID, err)
-								return 0, err
+				//ignore snapshots that have any tag
+				if len(s.Tags) == 0 {
+					// check the age of the snapshot
+					if parts := strings.Split(s.SnapshotID, "_"); len(parts) == 2 {
```
smousa: Need to not be parsing the snapshot tag here.  Need to call GetTenantVolume instead and SnapshotInfo

dao/elasticsearch/filesystem.go
```diff
@@ -335,3 +342,45 @@ func (dao *ControlPlaneDao) ReadyDFS(serviceID string, _ *int) (err error) {
...
+	dfslocker := dao.facade.DFSLock(ctx)
```
smousa: Don't use DFSLock here...if you need to add a Tag to a snapshot use the volume level lock instead

dao/elasticsearch/filesystem.go
```diff
@@ -335,3 +342,45 @@ func (dao *ControlPlaneDao) ReadyDFS(serviceID string, _ *int) (err error) {
...
+	*newTagList, err = dao.facade.TagSnapshot(request.SnapshotID, request.TagNames)
+	return err
+}
+
+// RemoveSnapshotTags removes specific tags from an existing snapshot
+func (dao *ControlPlaneDao) RemoveSnapshotTags(request model.TagSnapshotRequest, newTagList *[]string) error {
+	ctx := datastore.Get()
+
+	// synchronize the dfs
+	dfslocker := dao.facade.DFSLock(ctx)
```
smousa: Don't use DFSLock here, see above

dao/elasticsearch/filesystem.go
```diff
@@ -335,3 +342,45 @@ func (dao *ControlPlaneDao) ReadyDFS(serviceID string, _ *int) (err error) {
...
+	*newTagList, err = dao.facade.RemoveSnapshotTags(request.SnapshotID, request.TagNames)
+	return err
+}
+
+// RemoveAllSnapshotTags removes all tags from an existing snapshot
+func (dao *ControlPlaneDao) RemoveAllSnapshotTags(snapshotID string, _ *int) error {
+	ctx := datastore.Get()
+
+	// synchronize the dfs
+	dfslocker := dao.facade.DFSLock(ctx)
```
smousa: Don't use DFSLock here, see above

PR #2119, opened 2015-11-04
Fix acceptance tests... probably

rDr4g0n:
commit 22d0da91c81efc15c21b00292929807e6415d842
Author: John <rDr4g0n@gmail.com>
Date:   Wed Nov 4 14:42:07 2015 -0600

    yay, fixed test breakers

commit 857c8216dd525739abaff59283cb55f7593beab5
Author: John Matos <rDr4g0n@gmail.com>
Date:   Wed Nov 4 11:55:43 2015 -0600

    first pass at fix tests

### DIFF
--- a/acceptance/ui/features/login.feature
+++ b/acceptance/ui/features/login.feature
@@ -8,6 +8,7 @@ Feature: User login
       And I fill in the user id field with the default user id
       And I fill in the password field with the default password
       And I click the sign-in button
+      And I close the deploy wizard if present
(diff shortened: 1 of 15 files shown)


tgjonestx:
I had problems in both Firefox and Chrome with the Host Management 
Scenario: Add an invalid host with an invalid port

I was able to reproduce that independent of Cucumber, so to me like a separate application bug that snuck in while some of the other tests were broken.

PR #2122, opened 2015-11-06
Feature/gostacktool

kwalkerz:
New tool for analyzing Go stack traces. (Currently, only counts number of occurrences of each stack)


tools/gostacktool/gostack/tests/stackpanic.txt
```diff
@@ -0,0 +1,887 @@
+goroutine 2365 [running]:
```
smousa: Would it be possible for you to mask the file paths?

kwalkerz: You mean leave your ID (and mine in the other test file) out of it?

PR #2128, opened 2015-11-10
Feature/cc 1503.2

tgjonestx:
Another step towards reporting and validating endpoints, [CC-1503](https://jira.zenoss.com/browse/CC-1503).  This one includes a lot of refactoring to remove dao from the picture (RPC server calls facade layer directly), as well as changing the return data type to be more specific to reporting and including optionally imported endpoints in the results.

The remaining work for future PR(s) includes:
 * reporting on multiple services; e.g {{serviced service endpoints serviceID1 servicedID2}}
 * reporting on all children of a service; e.g. {[serviced service endpoints --all tenantID}}
 * returning validation messages (either info, warn and error)
 * expanded validation


cli/api/api.go
```diff
@@ -173,7 +173,7 @@ func (a *api) StartServer() error {
 }
 
 // Opens a connection to the master if not already connected
-func (a *api) connectMaster() (*master.Client, error) {
+func (a *api) connectMaster() (master.ClientInterface, error) {
```
smousa: Why do we need to pass in an interface here?  The api itself is an interface

tgjonestx: From a language perspective, `master.Client` is an implementation of an RPC client.  By returning an interface here, I can inject a mock for unit testing methods in the api package which make calls to the RPC client.    For instance, see TestGetEndpoints_fails in service_test.go

PR #2131, opened 2015-11-10
CC-1570: Deal with invalidly handled errors from pool factory functions; add tests

iancmcc:
https://jira.zenoss.com/browse/CC-1570


tgjonestx:
This is basically the same fix as #2129 :-)
But your unit-test is a bit more readable than mine and I you added a test to the client, so I suggest closing my PR but including the acceptance test fixes from it.

PR #2140, opened 2015-11-11
CC-1581: Added a defer to the progress update

smousa:
Same code is used in the backup logic (look at backup function in this file)

### DIFF
--- a/dao/elasticsearch/filesystem.go
+++ b/dao/elasticsearch/filesystem.go
@@ -125,14 +125,19 @@ func (dao *ControlPlaneDao) AsyncBackup(dirpath string, filename *string) (err e
 }
 
 // Restore restores the full application stack from a backup file.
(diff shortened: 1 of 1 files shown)


smousa:
@kwalkerz not for err because the err var is declared in the same scope, so it doesn't create a new object.

kwalkerz:
Okay, I see.

On Wed, Nov 11, 2015 at 11:12 AM, smousa <notifications@github.com> wrote:

> @kwalkerz <https://github.com/kwalkerz> not for err because the err var
> is declared in the same scope, so it doesn't create a new object.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/control-center/serviced/pull/2140#issuecomment-155849474>
> .
>



-- 
*Ken Walker* | Technical Lead
kwalker@zenoss.com | 512.687.6859

<http://www.zenoss.com/>

PR #2152, opened 2015-11-12
CC-1562: Service Definition can apply tenant images to its services

smousa:
RegistryIndex does not pass forward NoSuchEntity errors and download does an extra check to verify the service image

### DIFF
--- a/dfs/download.go
+++ b/dfs/download.go
@@ -17,9 +17,10 @@ import (
 	"errors"
 
 	"github.com/control-center/serviced/commons"
-	"github.com/control-center/serviced/datastore"
 	"github.com/control-center/serviced/dfs/docker"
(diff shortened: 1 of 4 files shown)


dfs/download.go
```diff
@@ -30,6 +31,74 @@ var (
...
+}
+
+// findImage will verify whether the image has already been deployed with the
+// application.
+func (dfs *DistributedFilesystem) findImage(image, tenantID string) (string, error) {
+	// Is the image in the index?
+	rImage, err := dfs.index.FindImage(image)
+	if err != nil {
+		if err == index.ErrImageNotFound {
+			return "", nil
```
wwwtyro: Are we using a return value set of `"", nil` to indicate that an image was not found? Any reason not to use index.ErrImageNotFound to convey that?

PR #2166, opened 2015-11-17
Move the docker registry path to v2 directory

rDr4g0n:
- Added PreStart to IServiceDefinition to run at the initialization of the IService
- Added PreStart method to the docker isvc
- If migrating from 1.1.0 to 1.1.x, moves docker registry directory from `registry` to `v2`
- otherwise, do not do anything as it will be handled elsewhere


rDr4g0n:
retest this

PR #2179, opened 2015-11-18
Merge pull request #2175 from control-center/feature/CC-1608

smousa:
CC-1608: query zookeeper after establishing a connection during healthcheck

### DIFF
--- a/isvcs/zk.go
+++ b/isvcs/zk.go
@@ -78,31 +78,46 @@ func initZK() {
 
 // a health check for zookeeper
 func zkHealthCheck(halt <-chan struct{}) error {
-	lastError := time.Now()
+	// minUptime til zookeeper is considered available
 	minUptime := time.Second * 2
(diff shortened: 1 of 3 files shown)


isvcs/zk.go
```diff
@@ -78,31 +78,46 @@ func initZK() {
...
 // a health check for zookeeper
 func zkHealthCheck(halt <-chan struct{}) error {
-	lastError := time.Now()
+	// minUptime til zookeeper is considered available
 	minUptime := time.Second * 2
-	zookeepers := []string{"127.0.0.1:2181"}
-
+	// timer to verify that the service has been available for the min duration
+	timer := time.NewTimer(minUptime)
+	timer.Stop()
```
alapidas: Why stop the timer here?

smousa: because we don't want the timer running while we are in the process of acquiring a connection

PR #2181, opened 2015-11-18
Feature/local rpc calls

jplouis:


rpc/rpcutils/local.go
```diff
@@ -0,0 +1,116 @@
...
+
+	l.RLock()
+	server, ok := l.rcvrs[name]
+	l.RUnlock()
+	if !ok {
+		return fmt.Errorf("Server Not Found for %s", serviceMethod)
+
+	}
+
+	method := reflect.ValueOf(server).MethodByName(methodName)
```
tgjonestx: What if MethodByName() fails?

cli/api/daemon.go
```diff
@@ -713,15 +713,23 @@ func (d *daemon) startAgent() error {
 func (d *daemon) registerMasterRPC() error {
 	glog.V(0).Infoln("registering Master RPC services")
 
-	if err := d.rpcServer.RegisterName("Master", master.NewServer(d.facade)); err != nil {
+	server := master.NewServer(d.facade)
+	bypass := os.Getenv("BYPASS_LOCAL_RPC")
```
tgjonestx: So this is an opt-in model?

If it works and we trust, wouldn't we want it used by default, and just leave ourselves an escape hatch if we find an unanticipated problem later?

PR #2192, opened 2015-11-19
Feature/local rpc calls

jplouis:


cli/api/daemon.go
```diff
@@ -713,15 +713,23 @@ func (d *daemon) startAgent() error {
 func (d *daemon) registerMasterRPC() error {
 	glog.V(0).Infoln("registering Master RPC services")
 
-	if err := d.rpcServer.RegisterName("Master", master.NewServer(d.facade)); err != nil {
+	server := master.NewServer(d.facade)
+	disableLocal := os.Getenv("DISABLE_RPC_BYPASS")
+	if disableLocal == "" {
+		rpcutils.RegisterLocalAddress(options.Endpoint, fmt.Sprintf("localhost:%s", options.RPCPort),
```
tgjonestx: maybe a log message here to record the fact that we're running in a non-default configuration

rpc/rpcutils/local.go
```diff
@@ -0,0 +1,162 @@
...
+		mVal := rcvrV.MethodByName(mName)
+		if !mVal.IsValid() {
+			continue
+		}
+		key := fmt.Sprintf("%s.%s", name, mName)
+		m := method{}
+		m.methodV = mVal
+		m.replyT = mVal.Type().In(1)
+		l.rcvrMs[key] = m
+	}
```
tgjonestx: Do we want to log an error if nothing is registered for `name`?

tgjonestx:
:+1:

PR #2201, opened 2015-11-23
Feature/cc 1243 2

rDr4g0n:
Preserves state of service tree during navigation, so services that are collapsed stay collapsed. The newly created CCUIState service can be used in the future to store other state stuff like graph settings.

Note that state is not persisted to disk. A page refresh will clear it, but angularjs route navigations (almost all links in the UI are route navigation) will retain state.

### DIFF
--- a/web/ui/src/Services/ServiceDetailsController.js
+++ b/web/ui/src/Services/ServiceDetailsController.js
@@ -615,6 +615,26 @@
 
                 // setup breadcrumbs
                 $scope.breadcrumbs = makeCrumbs($scope.services.current);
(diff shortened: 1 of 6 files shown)


web/ui/src/Services/ServiceDetailsController.js
```diff
@@ -644,15 +669,70 @@
             }
         };
```
rDr4g0n: Its bad practice to touch the DOM in an angular app, so getting rid of these DOM based view updates and using regular data binding is good news.

PR #2211, opened 2015-12-01
Fixes CC-1595: Leak around zk bindings

kwalkerz:


coordinator/client/connection.go
```diff
@@ -27,7 +27,7 @@ type Lock interface {
 
 // Leader is the interface that a Leaer implementation must implement
 type Leader interface {
-	TakeLead() (<-chan Event, error)
```
smousa: use <-chan struct{} instead of <-chan bool{} because it is smaller

coordinator/client/connection.go
```diff
@@ -59,10 +59,10 @@ type Connection interface {
 	EnsurePath(path string) error
 	Exists(path string) (bool, error)
 	Delete(path string) error
-	ChildrenW(path string) (children []string, event <-chan Event, err error)
```
smousa: see previous comment

coordinator/client/connection.go
```diff
@@ -59,10 +59,10 @@ type Connection interface {
...
+	ChildrenW(path string, done <-chan bool) (children []string, event <-chan Event, err error)
 	Children(path string) (children []string, err error)
 	Get(path string, node Node) error
-	GetW(path string, node Node) (<-chan Event, error)
```
smousa: see previous comment

coordinator/client/zookeeper/connection.go
```diff
@@ -219,13 +219,17 @@ func (c *Connection) Delete(path string) error {
 	return xlateError(c.conn.Delete(join(c.basePath, path), stat.Version))
 }
 
-func toClientEvent(zkEvent <-chan zklib.Event) <-chan client.Event {
```
smousa: see previous

coordinator/client/zookeeper/connection.go
```diff
@@ -233,7 +237,7 @@ func toClientEvent(zkEvent <-chan zklib.Event) <-chan client.Event {
 
 // ChildrenW returns the children of the node at the given path and a channel of
 // events that will yield the next event at that node.
-func (c *Connection) ChildrenW(path string) (children []string, event <-chan client.Event, err error) {
```
smousa: see previous

smousa: Just an aside--
I like the idea of instead of passing in a done channel, to instead to offer a StopWatch method on the the return on the watcher.  See https://golang.org/pkg/time/#Timer as an example.

PR #2217, opened 2015-12-02
UpgradeRegistry can now move images from old docker registry

smousa:
Added a registryHost field when upgrading images so that images can be migrated from old docker registry


dfs/upgraderegistry.go
```diff
@@ -49,6 +51,17 @@ func (dfs *DistributedFilesystem) UpgradeRegistry(svcs []service.Service, tenant
 			glog.Warningf("Cannot parse image name %s under service %s (%s)", image, svc.Name, svc.ID)
 			continue
 		}
+		// download image from old registry
+		if registryHost != "" {
+			glog.V(2).Info("Downloading image %s from %s registry", image, registryHost)
+			oldImage := fmt.Sprintf("%s/%s", registryHost, rImage)
+			if err := dfs.docker.PullImage(oldImage); err != nil {
```
jplouis: Are the V2 registry and old registry going to be running at the same time?  If so how will you ensure it pulls from the old registry and not the V2 registry?

dfs/upgraderegistry.go
```diff
@@ -49,6 +51,17 @@ func (dfs *DistributedFilesystem) UpgradeRegistry(svcs []service.Service, tenant
 			glog.Warningf("Cannot parse image name %s under service %s (%s)", image, svc.Name, svc.ID)
 			continue
 		}
+		// download image from old registry
+		if registryHost != "" {
+			glog.V(2).Info("Downloading image %s from %s registry", image, registryHost)
```
tgjonestx: If this is only supposed to happen on an upgrade, then I suggest removing `V(2)` and letting it log everytime because it's an unusual condition.

dfs/upgraderegistry.go
smousa: Yes, and you will set the host of the v1 registry from where the images are being pulled

dfs/upgraderegistry.go
smousa: ok

dfs/upgraderegistry.go
jplouis: Won't they both be localhost?

smousa: yea, but they will be different ports, so you would be passing in "localhost:5001" for the v1 registry

PR #2218, opened 2015-12-02
CC-1642 Use devicemapper as default storage driver

ed-:
Updates the default configs to use devicemapper
instead of rsync.


smousa:
Good work!  Waiting for tests to pass.

PR #2230, opened 2015-12-07
Feature/cc 1636 - Docker image IDs may change on push, breaking multi-host serviced

johnlhamilton:
Fixes 
CC-1636

The commit now calls a different push function, that actually does a push/remove/pull to get the correct ID.  The commit waits for the push/remove/pull to complete, then re-pushes with the correct ID.  If the push fails, commit will re-push the old image with the old ID.


dfs/commit.go
```diff
@@ -48,15 +48,34 @@ func (dfs *DistributedFilesystem) Commit(ctrID string) (string, error) {
...
 	img, err := dfs.docker.CommitContainer(ctr.ID, ctr.Config.Image)
 	if err != nil {
 		glog.Errorf("Could not commit container %s: %s", ctr.ID, err)
 		return "", err
 	}
 	// push the image into the registry
-	if err := dfs.index.PushImage(rImage.String(), img.ID); err != nil {
-		glog.Errorf("Could not push image %s (%s): %s", rImage, img.ID, err)
-		return "", err
+	if (img.ID != oldID) { //If the commit produced a new ID, call PushImageAfterCommit, which will actually push the image twice to address an issue with imageIDs changing in docker 1.9.1
```
smousa: I don't understand the purpose of this if statement.  The image id ALWAYS changes on a commit.

dfs/registry/index.go
```diff
@@ -38,6 +39,7 @@ var _ = RegistryIndex(&RegistryIndexClient{})
 type facade interface {
 	GetRegistryImage(ctx datastore.Context, image string) (*registry.Image, error)
 	SetRegistryImage(ctx datastore.Context, rImage *registry.Image) error
+	SetRegistryImageAfterCommit(ctx datastore.Context, rImage *registry.Image) error
```
smousa: Why do you need a SetAfterCommit?  SetRegistryImage indiscriminantly writes a value to the database

dfs/registry/index.go
```diff
@@ -105,6 +107,25 @@ func (client *RegistryIndexClient) PushImage(image, uuid string) error {
 	return client.facade.SetRegistryImage(client.ctx, rImage)
 }
 
+//Due to an issue in docker 1.8.3 - 1.9.1, after a commit, the push will result in the master having a different imageID than the registry
+//  This code path will have to fix the issue by getting the correct imageID after the push and updating the registry index
+func (client *RegistryIndexClient) PushImageAfterCommit(image, uuid string) error {
```
smousa: I don't think you need this function either, just call PushImage

dfs/registry/listener.go
```diff
@@ -156,11 +160,32 @@ func (l *RegistryListener) Spawn(shutdown <-chan interface{}, id string) {
...
+							glog.V(1).Infof("This is a post-commit push, so we will Push, Remove, Pull")
+							var newID	string
+							if newID, err = l.docker.PushImageAfterCommit(registrypath); err != nil {
+								glog.Errorf("Could not perform post-commit push for %s: %s", registrypath, err)
+								node.NewID = ""
+								node.AfterCommit = true
+								node.AfterCommitPushFailed = true	
+								node.PushedAt = time.Unix(0, 0)
+							} else {
+								node.NewID = newID
```
smousa: How do you update this data in the facade?

johnlhamilton: utils.go - SetRegistryImageAfterCommit watches the node until the push is complete, then returns the new ID to the facade, which re-pushes with the correct ID.

dfs/commit.go
johnlhamilton: I thought I was handling an edge case where the container is no different from the base image, but it looks like that is an impossible case.

PR #2245, opened 2015-12-10
Feature/cc 1636

johnlhamilton:
Fixes CC-1636

If the ID of the pulled image does not match what is expected, we compare a hash of the image histories, excluding IDs.


dfs/docker/docker.go
```diff
@@ -169,3 +174,25 @@ func (d *DockerClient) fetchCreds(registry string) (auth dockerclient.AuthConfig
...
+	var buffer bytes.Buffer
+	for _, history := range historyList {
+		imageDataString := fmt.Sprintf("%d-%s-%d\n", history.Created, history.CreatedBy, history.Size)
+		buffer.WriteString(imageDataString)
+	}
+
+	h := sha256.New()
+	h.Write(buffer.Bytes())
+
+	sha := base64.URLEncoding.EncodeToString(h.Sum(nil))
```
smousa: why are you passing nil?

dfs/registry/index.go
```diff
@@ -96,11 +98,19 @@ func (client *RegistryIndexClient) PushImage(image, uuid string) error {
 	if imageID.IsLatest() {
 		imageID.Tag = docker.Latest
 	}
+
+	//get the hash and store it in the image record
+	hash, err := client.dc.GetImageHash(uuid)
```
smousa: Can this be called in the dfs code instead where you add a param for hash in the PushImage function?
 so, PushImage(image, uuid, hash string) error

dfs/registry/mocks/Docker.go
```diff
@@ -0,0 +1,179 @@
...
+
+import "io"
+
+import dockerclient "github.com/fsouza/go-dockerclient"
+
+type Docker struct {
+	mock.Mock
+}
+
+func (_m *Docker) FindImage(image string) (*dockerclient.Image, error) {
```
smousa: You don't need this file, use /dfs/docker/mocks/Docker object

PR #2246, opened 2015-12-10
Feature/cc 1636

johnlhamilton:
fixes CC-1636

When pushing an image, we generate a hash of the image's complete history, omitting image IDs.  When pulling, if the pulled image does not have the expected ID, we compare hashes


smousa:
retest this

PR #2249, opened 2015-12-11
Feature/cc 1640

alapidas:
This change adds some extra info into the `servicestate` nodes in ZK, which, among other things, allows us to accurately update the `InSync` attribute on a `servicestate` when a container commit happens.

https://jira.zenoss.com/browse/CC-1640

### DIFF
--- a/dfs/registry/pull.go
+++ b/dfs/registry/pull.go
@@ -32,7 +32,7 @@ var (
 // Registry performs specific docker actions based on the registry index
 type Registry interface {
 	SetConnection(conn client.Connection)
-	PullImage(cancel <-chan time.Time, image string) (string, error)
+	PullImage(cancel <-chan time.Time, image string) error
 	ImagePath(image string) (string, error)
 }
 
@@ -54,11 +54,11 @@ func (l *RegistryListener) ImagePath(image string) (string, error) {
 }
 
 // PullImage waits for an image to be available on the docker registry so it
-// can be pulled (if it does not exist locally).  Returns the repo string, eg. "tenantID/repo:latest"
-func (l *RegistryListener) PullImage(cancel <-chan time.Time, image string) (string, error) {
+// can be pulled (if it does not exist locally).
+func (l *RegistryListener) PullImage(cancel <-chan time.Time, image string) error {
 	imageID, err := commons.ParseImageID(image)
 	if err != nil {
-		return "", err
+		return err
 	}
 	rImage := &registry.Image{
 		Library: imageID.User,
@@ -80,7 +80,7 @@ func (l *RegistryListener) PullImage(cancel <-chan time.Time, image string) (str
 			if err == client.ErrNoNode {
 				glog.Errorf("Image %s not found", regaddr)
 			}
-			return "", err
+			return err
 		}
 		// check if the image exists locally
 		glog.Infof("Looking up image %s", regaddr)
@@ -89,7 +89,7 @@ func (l *RegistryListener) PullImage(cancel <-chan time.Time, image string) (str
 			glog.Infof("Pulling image %s from the docker registry", regaddr)
 			if err := l.docker.PullImage(regaddr); err != nil && !docker.IsImageNotFound(err) {
 				glog.Errorf("Could not pull %s: %s", regaddr, err)
-				return "", err
+				return err
 			}
 			// was the pull successful?
 			if err := l.docker.TagImage(node.Image.UUID, regaddr); docker.IsImageNotFound(err) {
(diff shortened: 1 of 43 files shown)


dfs/registry/pull.go
```diff
@@ -32,7 +32,7 @@ var (
 // Registry performs specific docker actions based on the registry index
 type Registry interface {
 	SetConnection(conn client.Connection)
-	PullImage(cancel <-chan time.Time, image string) error
+	PullImage(cancel <-chan time.Time, image string) (string, error)
```
smousa: Why do we need to return the string of the image that we are pulling? Is the image field not sufficient?

alapidas: We are actually returning a normalized version of the `image` string that gets passed in.  It's possible that the supplied string has a host/port in it, and we don't want to return that.

dfs/registry/utils.go
```diff
@@ -34,38 +35,68 @@ func GetRegistryImage(conn client.Connection, id string) (*registry.Image, error
...
 		leader := conn.NewLeader(leaderpath, leadernode)
 		leaderDone := make(chan struct{})
 		defer close(leaderDone)
 		_, err := leader.TakeLead(leaderDone)
 		if err != nil {
-			return err
+			return "", err
 		}
 		defer leader.ReleaseLead()
 		if err := conn.Get(imagepath, node); err != nil {
```
smousa: You should be able to acquire the image UUID from zk on this call, there is not need to make an additional call for GetImageUUID

facade/registry.go
```diff
@@ -35,9 +37,37 @@ func (f *Facade) SetRegistryImage(ctx datastore.Context, rImage *registry.Image)
 	if err := f.registryStore.Put(ctx, rImage); err != nil {
 		return err
 	}
-	if err := f.zzk.SetRegistryImage(rImage); err != nil {
+	uuid, err := f.zzk.SetRegistryImage(rImage)
+	if err != nil {
 		return err
 	}
+	svcs, err := f.GetAllServices(ctx)
```
smousa: why not use that GetServicesByImage call?

facade/service.go
```diff
@@ -508,6 +508,37 @@ func (f *Facade) GetHealthChecksForService(ctx datastore.Context, serviceID stri
 	return svc.HealthChecks, nil
 }
 
+// GetServicesByImage fetches, from Elastic, all services using the supplied image ID.
+// Empty parts of the supplied image ID will not be considered.  For example,
+// "alskdjalskdjas/myImage:latest", "myImage:latest", "myImage"
+func (f *Facade) GetServicesByImage(ctx datastore.Context, imageID string) ([]service.Service, error) {
```
smousa: Is there a way that we can delegate this behavior on the direct elastic call?

facade/service.go
```diff
@@ -575,6 +606,15 @@ func (f *Facade) GetServices(ctx datastore.Context, request dao.EntityRequest) (
 	}
 }
 
+// GetServices will get all the services
```
smousa: We have a GetServices call that almost does the same thing.  Make sure to note that this call returns incomplete or "raw" services data if it has to be exposed publically.

dfs/registry/utils.go
alapidas: Agreed - but if the `Create` call here succeeds without error, then the `node` here would not be available to get the UUID from, and we'd end up needing a new connection to get it below

facade/registry.go
alapidas: I can switch over to that.  I originally had it in there, but switched to `GetAllServices` because I thought I had to cover the swapping repos case in here, but I ended up covering that in the service facade.

zzk/service/host.go
```diff
@@ -292,10 +292,13 @@ func (l *HostStateListener) startInstance(shutdown <-chan interface{}, locker sy
 		}
 	}()
 	// Pull the image
-	if err := l.handler.PullImage(timeoutC, svc.ImageID); err != nil {
+	repo, uuid, err := l.handler.PullImage(timeoutC, svc.ImageID)
+	if err != nil {
 		glog.Errorf("Error trying to pull image %s for service %s (%s): %s", svc.ImageID, svc.Name, svc.ID, err)
 		return nil, err
 	}
+	state.ImageRepo = repo
```
smousa: This can be svc.ImageID

dfs/registry/utils.go
smousa: The intention of SetRegistryImage implies that the image you are passing in should already have the UUID value populated.

facade/service.go
alapidas: Perhaps

PR #2256, opened 2015-12-12
Feature/ta25730

smousa:
New CLI command:
serviced docker migrate-registry [--endpoint HOST:PORT] [--override]
* endpoint (string, optional): is the endpoint to a remote docker registry
* override (bool, optional): overrides existing images that may have already been written to the current docker registry

MigrateRegistry moves images from an older or remote docker registry, specifically used in the v1=>v2 registry migration. See https://rally1.rallydev.com/#/46743532340d/detail/userstory/47069032153 for more details.

### DIFF
--- a/cli/api/docker.go
+++ b/cli/api/docker.go
@@ -13,22 +13,30 @@
 
 package api
 
-import "github.com/control-center/serviced/dao"
-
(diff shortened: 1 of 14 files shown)


iancmcc:
The branch name has two numbers transposed, for posterity. It should be TA27530.

PR #2258, opened 2015-12-14
CC-1666: Update InSync on servicestates

alapidas:
After a container commit, update InSync on the affected services.


smousa:
waiting for tests to pass

PR #2261, opened 2015-12-14
Use docker logging options for isvcs containers

tgjonestx:
Fixes CC-1665


tgjonestx:
Demo:

On Centos/RHEL machines. add the following to `/etc/default/serviced` to route docker output to journald instead of the default docker logging (`json-file`):
```
SERVICED_DOCKER_LOG_DRIVER=journald
SERVICED_DOCKER_LOG_CONFIG=,
```
Note that the single comma for SERVICED_DOCKER_LOG_CONFIG is required since serviced hard-codes default options for `json-file` logging which are invalid for `journald` logging. The single comma ultimately produces an empty map of key/value pairs.

After restarting serviced, use the following to inspect zookeeper logs:
```
journalctl CONTAINER_NAME=serviced-isvcs_zookeeper --since=today
```

PR #2266, opened 2015-12-15
CC-691: Add recursion to SVC_WAIT

alapidas:
This change adds the ability to recursively `SVC_WAIT` for services in a
`serviced service script` script.  This means we can now wait for all
services under an organizer to start/stop, for example.


smousa:
looks good, waiting for build to pass

PR #2269, opened 2015-12-16
snip out the cli interface for service migration

wwwtyro:


tgjonestx:
Can you remove the rest of the 'dead' code from cmd.cmdServiceMigrate method back to, but not including, the methods on Facade?

Also, we added an SVC_MIGRATE directive to the CLI scripting code which is no longer needed - that can be removed as well

PR #2273, opened 2015-12-16
strip out unecessary fields before marshalling into zookeeper

smousa:


zzk/service/service.go
```diff
@@ -417,36 +402,79 @@ func StopService(conn client.Connection, serviceID string) error {
...
+			glog.Errorf("Could not create node at %s: %s", svcNodePath, err)
 			return err
 		}
+	} else if err != nil {
+		glog.Errorf("Could not look up node for service %s: %s", svc.ID, err)
+		return err
+	}
+	// setLockOnUpdate sets the lock to true if enabled, otherwise it uses
+	// whatever existing value was previously set on the node.
+	if setLockOnUpdate {
```
tgjonestx: Is `setLockOnUpdate` used anywhere? I couldn't find a call to UpdateService() that had `true` for that argument.

PR #2281, opened 2015-12-17
CC-249: Validate virtual ip interface

smousa:
TODO: need to update the virtual ip listener to only bind to a hosts provided static ips

Before adding a new virtual ip, there is additional validation against whether any of the hosts in the given resource pool actually has a host that allows an ip to bind to the given interface.


smousa:
retest this

smousa:
It has been decided that limiting the interfaces based on the configured static ips would be too confusing to the user.

PR #2288, opened 2015-12-18
Feature/cc 1633

morr-zenoss:
Fix for CC-1633 - add routine to update servicevhosts in zookeeper on every sync. 
Fix for bug discovered while fixing CC-1633: return value from SyncServices was not being checked. When I checked it, it was returning an error. Properly initializing the ServiceNode on Update() fixed this problem.
Add unit test for servicevhost. This also entails adding a mock for the Zookeeper connection, to make it a true unit test.

### DIFF
--- a/zzk/service/service.go
+++ b/zzk/service/service.go
@@ -65,6 +65,29 @@ type ServiceNode struct {
 	version interface{}
 }
 
+// ID implements zzk.Node
+func (node *ServiceNode) GetID() string {
+	return node.ID
+}
+
+// Create implements zzk.Node
+func (node *ServiceNode) Create(conn client.Connection) error {
+	return UpdateService(conn, *node.Service, false)
+}
+
+// Update implements zzk.Node
+func (node *ServiceNode) Update(conn client.Connection) error {
+	// We have to get the node first, so the version is set properly.
+	var tempNode ServiceNode=ServiceNode{Service:&service.Service{}}
+	svcPath := servicepath(node.ID)
+	if err := conn.Get(svcPath, &tempNode); err != nil {
+		glog.V(2).Infof("Error getting ServiceNode for %s: %s", node.ID, err)
+		return err
+	}
+	tempNode.Service = node.Service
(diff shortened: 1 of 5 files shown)


morr-zenoss:
retest this

zzk/service/service.go
```diff
@@ -65,6 +65,19 @@ type ServiceNode struct {
 	version interface{}
 }
 
+// Update implements zzk.Node
+func (node *ServiceNode) Update(conn client.Connection) error {
```
tgjonestx: I don't think this is used anywhere - can you remove it without breaking anything?

zzk/service/servicevhost.go
```diff
@@ -119,6 +119,17 @@ func newVhostKey(serviceID string, vhostName string, enabled bool) VHostKey {
 	return VHostKey(fmt.Sprintf("%s_%s_%s", state, serviceID, vhostName))
 }
 
+
+func UpdateServicesVhosts(conn client.Connection, svcs []service.Service) error {
+	for _,svc := range svcs {
+		if err := UpdateServiceVhosts(conn, &svc) ; err != nil {
+			glog.Infof("Error Updating ServiceVhosts for Service %s: %s", svc.Name, err)
```
tgjonestx: should be `glog.Errorf`

zzk/service/servicevhost_test.go
```diff
@@ -0,0 +1,70 @@
+package service
```
tgjonestx: should have boilerplate file header

Also, please run gofmt on this file

zzk/service/servicevhost_test.go
```diff
@@ -0,0 +1,70 @@
...
+					Name: "test1",
+					Enabled: true,
+				},
+			},
+		},
+	}
+	return svc
+}
+
+func TestUpdateServiceVHosts(t *testing.T) {
```
tgjonestx: Can you add a unit-test for the new method UpdateService_s_Vhosts?

PR #2290, opened 2015-12-31
CC-1691: Ensure service path exists in zookeeper

kwalkerz:
During 5.0.x -> 5.1.1 upgrade testing, it disappeared somehow. So
make sure it's there when we need it.


kwalkerz:
Retest

PR #2305, opened 2016-01-06
Feature/cc 1683

rphillipsz:
Added a 'unique' filter and applied this to the Add Public Endpoints form


rphillipsz:
retest this

smousa:
I see that you are just deduping entries on the front-end, but why are there dupes on the front-end to begin with?

PR #2309, opened 2016-01-06
Feature/cc 1679

rphillipsz:
The problem is in the routes.  When passing "abc.xyz.com" into ":name", it gets interpreted as the value "abc" with a format parameter.  The /enable wasn't necessary because it's already part of the payload, so changing ":name" to "*name" and dropping "/enable" from the url argument results in the route getting the complete vhost name.  There was nothing wrong in cpserver.go


rphillipsz:
retest this

PR #2311, opened 2016-01-07
Small refinement for HA scenarios

tgjonestx:
Ignore file-exists errors when setting up the volume directory on an agent.  

Not sure why this hasn't been a problem in other multi-host deployments, but after an HA failover I ran into cases where the agent nodes were failing to start services because the client-side mount point for the application already existed.


tgjonestx:
Retest this.

PR #2321, opened 2016-01-07
fixes CC-1722 Two zookeeper clients are started on the master

johnlhamilton:
In startAgent, we do not start a zookeeper client if we are also a master.


johnlhamilton:
retest this

johnlhamilton:
retest this

PR #2330, opened 2016-01-08
Fixes CC-1722: 'Two zookeeper clients are started on the master'.  

johnlhamilton:
zk client is now created before startAgent and startMaster are called.  Removes redundant code from previous fix.


johnlhamilton:
retest this

PR #2338, opened 2016-01-11
Print bad lines along with warnings and errors

kwalkerz:
Every now and then I see a parsing error reported, and when the data is
being fetched from a running serviced you can't look up what the bad
line was. So print the bad line to assist with analysis and fixing.


kwalkerz:
retest

kwalkerz:
retest this

kwalkerz:
retest this

PR #2341, opened 2016-01-11
CC-1731: Tag not exists causes serviced to panic

smousa:
What is new:
serviced snapshot tag SNAPSHOTID TAG _returns empty on success_
serviced snapshot untag SERVICEID TAG _returns the affected snapshot id_

dfs.TagInfo replaces dfs.GetSnapshotByTag and returns the SnapshotInfo object
dfs.Untag replaces dfs.RemoveTag and returns the affected snapshot id
dfs.Tag only returns err or nil (on success)


smousa:
retest this

PR #2348, opened 2016-01-13
Exclude layer size from docker image hash

tgjonestx:
Fixes CC-1750

### DIFF
--- a/dfs/registry/listener.go
+++ b/dfs/registry/listener.go
@@ -120,6 +120,7 @@ func (l *RegistryListener) Spawn(shutdown <-chan interface{}, id string) {
 		glog.V(1).Infof("Spawn id=%s node: %s:%s %s", id, node.Image.Repo, node.Image.Tag, node.Image.Tag)
 		if node.PushedAt.Unix() == 0 {
 			// Do I have the image?
+			glog.Infof("Checking if push required for id=%s node: %s:%s %s (UUID=%s)", id, node.Image.Repo, node.Image.Tag, node.Image.Tag, node.Image.UUID)
 			if img, err := l.docker.FindImage(node.Image.UUID); err == nil {
 				glog.V(1).Infof("Found image %s locally, acquiring lead", node.Image)
 				func() {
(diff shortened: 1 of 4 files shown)


dfs/registry/utils.go
```diff
@@ -57,6 +57,7 @@ func SetRegistryImage(conn client.Connection, rImage registry.Image) error {
 			return err
 		}
 		node.Image = rImage
+		glog.Infof("Adding image %s (%s) to ZK", imagepath, node.Image.UUID)
```
smousa: This glog could be construed to mean that the image has been added, event though the request hasn't been made to zk yet.

dfs/registry/pull.go
```diff
@@ -103,9 +103,12 @@ func (l *RegistryListener) PullImage(cancel <-chan time.Time, image string) erro
 						//if the match, the image we have is current, just return
 						return nil
 					}
+					glog.Infof("Image %s (%s) has different hashes", regaddr, node.Image.UUID)
 				}
 
 				if node.PushedAt.Unix() > 0 {
+					glog.Infof("For image %s (idpath=%s), requesting push", regaddr, idpath)
```
smousa: I don't know what this message means

dfs/registry/pull.go
```diff
@@ -103,9 +103,12 @@ func (l *RegistryListener) PullImage(cancel <-chan time.Time, image string) erro
 						//if the match, the image we have is current, just return
 						return nil
 					}
+					glog.Infof("Image %s (%s) has different hashes", regaddr, node.Image.UUID)
```
smousa: This message seems unnecessary, why do we care if there are different hashes?

dfs/registry/listener.go
```diff
@@ -170,7 +171,7 @@ func (l *RegistryListener) Spawn(shutdown <-chan interface{}, id string) {
 				glog.Errorf("Could not find image %s: %s", node.Image.UUID, err)
 			}
 		}
-		glog.V(1).Infof("Waiting for image %s to update", node.Image)
+		glog.Infof("Waiting for image %s to update (imagepath=%s)", node.Image, imagepath)
```
smousa: This should go back to V(1) I think

dfs/registry/listener.go
```diff
@@ -120,6 +120,7 @@ func (l *RegistryListener) Spawn(shutdown <-chan interface{}, id string) {
 		glog.V(1).Infof("Spawn id=%s node: %s:%s %s", id, node.Image.Repo, node.Image.Tag, node.Image.Tag)
 		if node.PushedAt.Unix() == 0 {
 			// Do I have the image?
+			glog.Infof("Checking if push required for id=%s node: %s:%s %s (UUID=%s)", id, node.Image.Repo, node.Image.Tag, node.Image.Tag, node.Image.UUID)
```
smousa: This message should be a V(1)

PR #2350, opened 2016-01-13
Not creating reverse proxy when run as a master

achow-zenoss:


smousa:
1.  startDockerRegistry does not need all of the preceding validation.  All that it needs to do is check if it is master, do not start reverse proxy
2.  options.DockerRegistry is a user configured value and should be ignored altogether if running on the master because you will be running the docker registry locally on the container.

PR #2351, opened 2016-01-13
Allow interface specification in public ports

jhuie-zenoss:


web/ui/src/Services/ServiceDetailsController.js
```diff
@@ -180,9 +180,10 @@
                 return resourcesFactory.addVHost(serviceId, serviceEndpoint, name);
             } else if(newPublicEndpoint.type === "port"){
                 var port = newPublicEndpoint.port;
+                var ip = newPublicEndpoint.ip;
```
rDr4g0n: where does this `ip` value come from?

web/ui/src/Services/ServiceDetailsController.js
```diff
@@ -297,10 +298,13 @@
...
-                var host = $scope.defaultHostAlias;
-                // Port public endpoint port listeners are always on http
-                return "http://" + host + ":" + publicEndpoint.PortNumber;
+                if(publicEndpoint.PortNumber.startsWith(":")){
+                    var host = $scope.defaultHostAlias;
+                    // Port public endpoint port listeners are always on http
+                    return "http://" + host + publicEndpoint.PortNumber;
+                }else{
+                    return "http://" + publicEndpoint.PortNumber;
+                }
```
rDr4g0n: This code is looking for `PortNumber` property, but I think that property was renamed to `PortAddr`?

web/ui/src/Services/servicesFactory.js
```diff
@@ -519,7 +519,7 @@
                         if(endpoint.PortList){
                             endpoint.PortList.forEach(function(port){
                                 acc.push({
-                                    PortNumber: port.PortNumber,
+                                    PortNumber: port.PortAddr,
```
rDr4g0n: Rename `PortNumber` for consistency with the backend.

rDr4g0n:
Also, since the port field has changed to "Port Address", i think it might be useful to include some placeholder text in the input that clarifies that it expects ip:port type format.

container/endpoint.go
```diff
@@ -210,9 +210,9 @@ func buildExportedEndpoints(conn coordclient.Connection, tenantID string, state
 				}
 			}
 			if len(defep.PortList) > 0 {
-				exp.ports = []uint16{}
+				exp.ports = []string{}
```
smousa: ports is not a good name for this attribute if it doesn't refer to ports anymore

PR #2355, opened 2016-01-14
CC-1698 - Shutdown takes too long

morr-zenoss:
Address CC-1698 - Shutdown takes too long

This PR includes a couple of changes JP suggested to help with shutdown. 
The primary fix is non-code: make sure you're using lvm thin pool storage, and not loopback.


cli/api/daemon.go
```diff
@@ -361,7 +361,6 @@ func (d *daemon) run() (err error) {
 	}
 
 	zzk.ShutdownConnections()
-	volume.ShutdownAll()
```
smousa: Why did you remove this line?

morr-zenoss: Because JP asked me to - it was causing problems with NFS mounts.

PR #2356, opened 2016-01-14
CC-1707 (Shutdown takes too long) - port CC-1698 changes to support branch

morr-zenoss:
The primary fix is non-code: make sure you're using lvm thin pool storage (not loopback).
This PR includes a couple of changes JP suggested to help with shutdown issues.

### DIFF
--- a/cli/api/daemon.go
+++ b/cli/api/daemon.go
@@ -361,7 +361,6 @@ func (d *daemon) run() (err error) {
 	}
 
 	zzk.ShutdownConnections()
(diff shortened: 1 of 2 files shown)


cli/api/daemon.go
```diff
@@ -361,7 +361,6 @@ func (d *daemon) run() (err error) {
 	}
 
 	zzk.ShutdownConnections()
-	volume.ShutdownAll()
```
morr-zenoss: This change is to prevent NFS issues.

PR #2360, opened 2016-01-14
Removed restriction from the UI that prevented the last Public Endpoi…

rphillipsz:
…nt from being removed


rphillipsz:
retest this please

PR #2367, opened 2016-01-18
don't call testing.Fatal from secondary goroutines

wwwtyro:
`Fatal` is equivalent to a log and a `FailNow`, and `FailNow` can't be called from goroutines secondary to the one running the test: https://golang.org/pkg/testing/#B.FailNow

Fixes [CC-1784](https://jira.zenoss.com/browse/CC-1784)


alapidas:
LGTM - merge once tests pass

alapidas:
please retest this

PR #2369, opened 2016-01-18
CC-1760: Service migrations always overwrite original config data for a service

smousa:
* Fixed the boilerplates
* Service validation functions
* Service migrations overwrite existing original configs
* Service updates do not overwrite original configs
* MigrateServices do additional validation against deployed services
* moved service configuration CRUD into another file


facade/addressassignment.go
```diff
@@ -55,18 +63,6 @@ func (f *Facade) RemoveAddressAssignment(ctx datastore.Context, id string) error
 		return err
 	}
 
-	var svc *service.Service
```
jplouis: Why is this gone?

facade/service.go
```diff
@@ -66,89 +84,295 @@ func (f *Facade) AddService(ctx datastore.Context, svc service.Service) (err err
...
+func (f *Facade) validateServiceAdd(ctx datastore.Context, svc *service.Service) error {
+	store := f.serviceStore
+	// verify that the service does not already exist
+	if _, err := store.Get(ctx, svc.ID); !datastore.IsErrNoSuchEntity(err) {
+		if err != nil {
+			glog.Errorf("Could not check the existance of service %s (%s): %s", svc.Name, svc.ID, err)
+			return err
+		} else {
+			glog.Errorf("Could not add service %s (%s): %s", svc.Name, svc.ID, ErrServiceExists)
+			return ErrServiceExists
```
jplouis: When did we decide errors without context was desirable?

facade/addressassignment.go
smousa: Before, we were saving address assignment information on the service and in its own document.  So, now when we make changes to address assignments, we only save to the address assignment document.  And when we load services from the database, we load the address assignment onto the service from that document (which is something that we are already doing today)

facade/service.go
smousa: This isn't a stylistic regression, we have written validators like this for a while.  Is there a better way that you would prefer?

smousa:
retest this

smousa:
retest this

smousa:
retest this

facade/service.go
```diff
@@ -66,89 +84,295 @@ func (f *Facade) AddService(ctx datastore.Context, svc service.Service) (err err
...
-	// Strip the database version; we already know this is a create
-	svc.DatabaseVersion = 0
+func (f *Facade) validateServiceAdd(ctx datastore.Context, svc *service.Service) error {
+	store := f.serviceStore
+	// verify that the service does not already exist
+	if _, err := store.Get(ctx, svc.ID); !datastore.IsErrNoSuchEntity(err) {
+		if err != nil {
+			glog.Errorf("Could not check the existance of service %s (%s): %s", svc.Name, svc.ID, err)
+			return err
+		} else {
```
alapidas: Would we hit this on a nil error from the above `store.Get()` call?  Is that desired?

smousa: store.Get always returns ErrNoSuchEntity if not found
however, store.Find... will return nil on success

facade/serviceconfig.go
```diff
@@ -0,0 +1,123 @@
...
+			glog.Infof("Updating config file %s for service %s", configFile.Filename, serviceID)
+		} else {
+			svcConfigFile, err = serviceconfigfile.New(tenantID, servicePath, configFile)
+			if err != nil {
+				glog.Errorf("Could not create new service config file %s for service %s: %s", configFile.Filename, serviceID, err)
+				return err
+			}
+			glog.Infof("Adding config file %s for service %s", configFile.Filename, serviceID)
+		}
+		if err := configStore.Put(ctx, serviceconfigfile.Key(svcConfigFile.ID), svcConfigFile); err != nil {
```
alapidas: Could we end up with some config files being added to the DB, and some not?  Is it worth making this atomic?

PR #2379, opened 2016-01-19
Add new syntax for port numbers

jhuie-zenoss:


jplouis:
Please make ports actual number or string representation of numbers without colons.

jplouis:
there is a separate ticket to allow handling of regular port number format.

PR #2381, opened 2016-01-19
Add scrubber for almost valid port inputs

jhuie-zenoss:


domain/service/service.go
```diff
@@ -441,6 +446,20 @@ func (s *Service) EnablePort(application string, portAddr string, enable bool) e
 	return nil
 }
 
+// Make best effort to make a port address valid
+func ScrubPortString(port string) string{
+	// remove possible protocol at string beginning
+	re := regexp.MustCompile("^(.+://)")
```
iancmcc: This can be a global var.

domain/service/service.go
```diff
@@ -441,6 +446,20 @@ func (s *Service) EnablePort(application string, portAddr string, enable bool) e
...
+	scrubbed := re.ReplaceAllString(port, "")
+
+	matched, _ := regexp.MatchString("^[0-9]*$", port)
```
iancmcc: also this one. regexp.MustCompile("^[0-9]*$"). Shouldn't it be ^\d+$?

domain/service/service.go
```diff
@@ -441,6 +446,20 @@ func (s *Service) EnablePort(application string, portAddr string, enable bool) e
...
+	if  matched {
+		scrubbed = fmt.Sprintf(":%s", port)
```
iancmcc: This seems fairly dangerous. Why not, say, fmt.Sprintf(":%s", strings.TrimPrefix(port, ":"))?

domain/service/service.go
```diff
@@ -441,6 +446,20 @@ func (s *Service) EnablePort(application string, portAddr string, enable bool) e
...
+	}
+
+	return scrubbed
+}
```
iancmcc: You gots to test this yo

jplouis:
What will happen if the servicedefintion defines the port as "8080".  Seems to me that it would create a service with "8080" as the port and something down the line will fail.

Seems like you would want to convert here https://github.com/control-center/serviced/blob/develop/domain/service/service.go#L162-162

And you will want to validate that the port format is correct before saving in the DB here https://github.com/control-center/serviced/blob/develop/domain/service/validation.go#L43-43

jhuie-zenoss:
JP,  If the service definition defines the port as "8080" then it will be scrubbed as the endpoint is being created during template deployment. See the change in domain/service/service.go

domain/service/service.go
jhuie-zenoss: The regex guarantees that this code will not be reached if the string has a prefix or a suffix.

domain/service/service_test.go
```diff
@@ -148,3 +148,19 @@ func TestBuildServiceBuildsMetricConfigs(t *testing.T) {
...
+		"1234": ":1234",
+		":1234": ":1234",
+		"128.0.0.1:1234": "128.0.0.1:1234",
+		"http://128.0.0.1:1234": "128.0.0.1:1234",
+	}
+
+	for portString, expectedString := range testStrings {
+		scrubbedString := ScrubPortString(portString)
+		if scrubbedString != expectedString{
+			t.Fail("Port String scrubbing failed: %s != %s", scrubbedString, expectedString)
```
iancmcc: Failf? Does Fail accept format parameters?

jhuie-zenoss:
Retest this, please

PR #2393, opened 2016-01-20
CC-1800: pointers to arrays were affected 'copies' of services

smousa:
https://jira.zenoss.com/browse/CC-1800


smousa:
retest this

smousa:
retest this

PR #2397, opened 2016-01-21
Merge pull request #2393 from control-center/feature/CC-1800

smousa:
CC-1800: pointers to arrays were affected 'copies' of services
#2393


smousa:
ok to test

PR #2404, opened 2016-01-21
Fix serviced-develop-merge-root-integration tests

tgjonestx:
Corrects test failures in the merge build like http://jenkins.zendev.org/view/Control%20Center/job/serviced-develop-merge-root-integration/129 which were introduced with #2389

### DIFF
--- a/volume/devicemapper/devicemapper.go
+++ b/volume/devicemapper/devicemapper.go
@@ -289,21 +289,26 @@ func (d *DeviceMapperDriver) Release(volumeName string) error {
 	}
 	devices := vol.Metadata.ListDevices()
 	for _, device := range devices {
+		if len(device) == 0 {
+			// this can happen when all previously active devices have been deactivated
+			continue;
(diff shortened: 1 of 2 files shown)


volume/devicemapper/devicemapper.go
```diff
@@ -289,21 +289,26 @@ func (d *DeviceMapperDriver) Release(volumeName string) error {
 	}
 	devices := vol.Metadata.ListDevices()
 	for _, device := range devices {
+		if len(device) == 0 {
```
smousa: what is device?

tgjonestx: In a DM device name like `docker-147:1-195-2qnUYyFCcpgT4B4T3VRN8b`, device is `2qnUYyFCcpgT4B4T3VRN8b`.  It's the ID we track in the metadata we store in files under /opt/serviced/var/volumes/.devicemapper.

PR #2410, opened 2016-01-26
ZEN-21286: Do not fail, when stop stopped container

vikt0rs:
It is superficial to fail, when we try to stop docker container which
was already stopped, so let's ignore this situation.

Co-Authored-By: "Summer Mousa <smousa@zenoss.com>"

### DIFF
--- a/cli/api/shell.go
+++ b/cli/api/shell.go
@@ -238,7 +238,9 @@ func (a *api) RunShell(config ShellConfig) error {
 		// Delete the container
 
 		if err := dockercli.StopContainer(container.ID, 10); err != nil {
-			glog.Fatalf("failed to stop container: %s (%s)", container.ID, err)
+			if _, ok := err.(*dockerclient.ContainerNotRunning); !ok {
+			        glog.Fatalf("failed to stop container: %s (%s)", container.ID, err)
+			}
 		} else if err := dockercli.RemoveContainer(dockerclient.RemoveContainerOptions{ID: container.ID}); err != nil {
 			glog.Fatalf("failed to remove container: %s (%s)", container.ID, err)
 		}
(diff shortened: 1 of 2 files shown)


container/controller.go
```diff
@@ -755,10 +755,9 @@ func (c *Controller) checkPrereqs(prereqsPassed chan bool, rpcDead chan struct{}
...
 			for _, script := range c.prereqs {
 				glog.Infof("Running prereq command: %s", script.Script)
-				cmd := exec.Command("sh", "-c", script.Script)
-				err := cmd.Run()
+				out, err := exec.Command("sh", "-c", script.Script).CombinedOutput()
 				if err != nil {
-					msg := fmt.Sprintf("Not starting service yet, waiting on prereq: %s", script.Name)
+					msg := fmt.Sprintf("Service %s not starting. Output: %s; error: %s", script.Name, out, err)
 					glog.Warning(msg)
 					fmt.Fprintln(os.Stderr, msg)
```
smousa: Doesn't this print the message twice?

cli/api/shell.go
```diff
@@ -238,7 +238,9 @@ func (a *api) RunShell(config ShellConfig) error {
 		// Delete the container
 
 		if err := dockercli.StopContainer(container.ID, 10); err != nil {
-			glog.Fatalf("failed to stop container: %s (%s)", container.ID, err)
+			if _, ok := err.(*dockerclient.ContainerNotRunning); !ok {
+			        glog.Fatalf("failed to stop container: %s (%s)", container.ID, err)
+			}
 		} else if err := dockercli.RemoveContainer(dockerclient.RemoveContainerOptions{ID: container.ID}); err != nil {
```
smousa: This won't work, you need to do something like
dockercli.StopContainer
if err := dockercli.RemoveContainer(); err != nil {
   glog.Errorf
}

vikt0rs: We are removing container at the "elif.. " block below, this code was not changed. 

PS - Sorry, if I didn't get your point.

smousa:
We are not releasing anything on 1.0.x, closing pr

PR #2411, opened 2016-01-26
CC-2270: Add more information to log output

vikt0rs:
Function checkPrereqs() provides no information about the original error
in logs. This looks like a bad idea, so this commit adds error
status and output to logs to make debug easier.

For example it converts message [0] to [1]

[0] - `Not starting service yet, waiting on prereq: MariaDB connectivity`
[1] - `Service MariaDB connectivity not starting. Output: ERROR 2003
(HY000): Can't connect to MySQL server on '127.0.0.1' (111)
ERROR:zen.zendb:Error executing command: ['mysql', '--batch',
'--skip-column-names', '--user=zenoss', '--host=127.0.0.1',
'--port=3306', '--database=zodb']
; error: exit status 1
`


smousa:
LGTM, waiting for tests to pass

PR #2414, opened 2016-01-26
Fixes CC-1700 "Error executing step 3: SNAPSHOT - During RM upgrade"

johnlhamilton:
This problem is caused when an old snapshot from CC 1.0.x exists on the system after an upgrade to CC 1.1.x or higher. This PR ignores those snapshots for normal operations, and also adds an ability to query for and delete invalid snapshots through the API. "serviced snapshot list" will now include (and mark) invalid snapshots so that the user may delete them.

### DIFF
--- a/volume/nfs/nfs.go
+++ b/volume/nfs/nfs.go
@@ -222,6 +222,12 @@ func (v *NFSVolume) Snapshots() ([]string, error) {
 	return nil, ErrNotSupported
 }
 
+// InvalidSnapshots implements volume.Volume.InvalidSnapshots
+func (v *NFSVolume) InvalidSnapshots() ([]string, error) {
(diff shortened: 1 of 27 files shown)


volume/nfs/nfs.go
```diff
@@ -222,6 +222,12 @@ func (v *NFSVolume) Snapshots() ([]string, error) {
 	return nil, ErrNotSupported
 }
 
+// InvalidSnapshots implements volume.Volume.InvalidSnapshots
+func (v *NFSVolume) InvalidSnapshots() ([]string, error) {
+	//TODO:  Can we have invalid snapshots in devicemapper?
```
smousa: remove TODO

PR #2415, opened 2016-01-26
CC-1823: remove deprecated export volumes

smousa:


dfs/nfs/server.go
```diff
@@ -398,6 +398,40 @@ func (c *Server) cleanupBindMounts() {
...
 	}
+	// CC-1816: clean up remnants of old /exports/serviced_var_volumes directory
+	removeDeprecated("/exports/serviced_var_volumes")
+}
+
+// removeDeprecated will clean up any old exports path
+func removeDeprecated(dirpath string) {
+	if dirContents, err := ioutil.ReadDir(dirpath); !os.IsNotExist(err) {
+		if err != nil {
+			glog.Warningf("Could not look up deprecated exports path %s: %s", dirpath, err)
```
wwwtyro: Should this be a warning? Seems like if it's not there, we're cool?

dfs/nfs/server.go
```diff
@@ -398,6 +398,40 @@ func (c *Server) cleanupBindMounts() {
...
+			return
+		}
+		// Unmount path
+		if mounted, err := isBindMounted(dirpath); err != nil {
+			glog.Warningf("Could not determine if deprecated path %s is bind mounted: %s", dirpath, err)
```
wwwtyro: Since this is recursive, are we going to get a lot of errors for all the subdirs?

dfs/nfs/server.go
```diff
@@ -398,6 +398,40 @@ func (c *Server) cleanupBindMounts() {
...
+			}
+		}
+		// recurse through the filesystem
+		for _, file := range dirContents {
+			if file.IsDir() {
+				removeDeprecated(filepath.Join(dirpath, file.Name()))
+			}
+		}
+		// remove the path
+		if err := os.RemoveAll(dirpath); err != nil {
```
wwwtyro: I must be missing something - why is this recursive if we're just going to RemoveAll anyway? Are we checking for nested bindmounts?

dfs/nfs/server.go
smousa: I set everything to warning because the caller doesn't set a return value, I was maintaining consistency.

dfs/nfs/server.go
smousa: It will only complain about the bind mounted paths

dfs/nfs/server.go
smousa: what this is doing is 
1) unmounting the base path
2) unmount mounted children
3) delete mounted children
4) delete basepath

PR #2417, opened 2016-01-26
Fixes CC-1700 "Error executing step 3: SNAPSHOT - During RM upgrade"

johnlhamilton:
This problem is caused when an old snapshot from CC 1.0.x exists on the system after an upgrade to CC 1.1.x or higher. This PR adds better error handling, and will flag these snapshots as [INVALID] in the CLI so they can be removed.


dao/model.go
```diff
@@ -126,13 +126,19 @@ type SnapshotInfo struct {
 	Description string
 	Tags        []string
 	Created     time.Time
+	Invalid     bool
 }
 
 func (s SnapshotInfo) String() string {
+	snapshotID := s.SnapshotID
+	if s.Invalid {
+		snapshotID += " [INVALID]"
```
smousa: Change this to say DEPRECATED instead

cli/cmd/snapshot.go
```diff
@@ -187,10 +187,14 @@ func (c *ServicedCli) cmdSnapshotList(ctx *cli.Context) {
 			for _, s := range snapshots {
 				//build a comma-delimited list of the tags
 				tags := strings.Join(s.Tags, ",")
+				snapshotID := s.SnapshotID
+				if s.Invalid {
+					snapshotID += " [INVALID]"
+				}
```
smousa: Why are you setting this value again?  Wouldn't you have gotten this from the dao/model object?

PR #2430, opened 2016-01-28
CC-1832: update devinfo when devmapper volume is resized

smousa:


volume/devicemapper/devicemapper.go
```diff
@@ -380,6 +405,66 @@ func (d *DeviceMapperDriver) deactivateDevice(devname string) error {
...
+	defer fh.Close()
+	if err := json.NewDecoder(fh).Decode(devInfo); err != nil {
+		glog.Errorf("Could not decode device info for %s: %s", devname, err)
+		return err
+	}
+	return nil
+}
+
+// writeDeviceInfo performs an atomic write to the device metadata file
+func (d *DeviceMapperDriver) writeDeviceInfo(devname string, devInfo devInfo) error {
```
jplouis: There is a util for doing atomic writes https://github.com/control-center/serviced/blob/feature/CC-1435-3/commons/atomicfile/atomicfile.go#L24-24

PR #2433, opened 2016-01-29
enable TLS for rpc connections. Optional certificate validation.

jplouis:


pkg/serviced.default
```diff
@@ -37,6 +37,12 @@
 # Set the RPC port to listen on
 # SERVICED_RPC_PORT=4979
 
+# Enable server certificate verifcation for TLS RPC request
+# SERVICED_RPC_CERT_VERIFY=falsS
```
tgjonestx: typo

cli/api/daemon.go
```diff
@@ -182,7 +182,7 @@ func (d *daemon) stopISVCS() {
 	glog.Infof("isvcs shut down")
 }
 
-func (d *daemon) startRPC() {
+func (d *daemon) startRPC() error {
```
tgjonestx: I don't see a case that this method ever returns an error, and there are no checks for errors when we call `startRPC()`, so seems like this should be a void return value.

tgjonestx:
:+1:   just waiting for PR builds to finish

PR #2434, opened 2016-01-29
fixes CC-1834 "Cannot remove [DEPRECATED] snapshots"

johnlhamilton:
https://jira.zenoss.com/browse/CC-1834


dfs/info.go
```diff
@@ -37,7 +37,7 @@ func (dfs *DistributedFilesystem) getSnapshotVolumeAndInfo(snapshotID string) (v
 	info, err := vol.SnapshotInfo(snapshotID)
 	if err != nil {
 		glog.Errorf("Could not get info for snapshot %s: %s", snapshotID, err)
-		return nil, nil, err
+		return vol, nil, err
```
smousa: this is a bad practice, why not instead make the call GetTenant instead of getSnapshotVolumeAndInfo when you are calling delete?

PR #2444, opened 2016-02-02
Continuation of fix for CC-1814

wwwtyro:


shell/server.go
```diff
@@ -449,6 +450,17 @@ func StartDocker(cfg *ProcessConfig, dockerRegistry, port, controller string) (*
 	argv = append(argv, "-e", fmt.Sprintf("SERVICED_IS_SERVICE_SHELL=true"))
 	argv = append(argv, "-e", fmt.Sprintf("SERVICED_SERVICE_IMAGE=%s", image))
 
+	uiport := os.Getenv("SERVICED_UI_PORT")
```
jplouis: Shouldn't the value be passed in?  The ui port can be set at the command line to override ENV vars.

alapidas:
Plz look @ jp's comment

PR #2446, opened 2016-02-02
Fixes CC-1826 "Public Endpoint: Starting a public endpoint with a previously used and deleted port fails to start"

johnlhamilton:
Users are now forced to restart the affected service when adding or deleting a public endpoint.  New language in the dialog and on the button makes it clear that the action will restart the service.

Additionally, new public endpoints are enabled by default when they are created.

### DIFF
--- a/web/ui/src/Services/ServiceDetailsController.js
+++ b/web/ui/src/Services/ServiceDetailsController.js
@@ -118,7 +118,7 @@
                         }
                     },{
                         role: "ok",
-                        label: "add_public_endpoint",
+                        label: "add_public_endpoint_confirm",
                         action: function(){
                             var newPublicEndpoint = $scope.publicEndpoints.add;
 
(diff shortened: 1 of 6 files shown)


web/ui/src/Services/ServiceDetailsController.js
```diff
@@ -499,7 +499,8 @@
 
             $modalService.create({
                 template: $translate.instant("remove_public_endpoint") + ": <strong>"+
-                          (publicEndpoint.Name ? publicEndpoint.Name : "port " + publicEndpoint.PortAddr) + "</strong>",
+                          (publicEndpoint.Name ? publicEndpoint.Name : "port " + publicEndpoint.PortAddr) + "</strong><br><br>"+
+                          "The <strong>" + publicEndpoint.Application + "</strong> service will be restarted.",
```
rDr4g0n: The action that is going to occur should be very explicitly spelled out here. Something like "After the public endpoint is removed, the HMaster service will automatically be restarted."

johnlhamilton:
retest this

rphillipsz:
retest this please

PR #2450, opened 2016-02-03
Cleanup loopback devices created for smoke and acceptance tests

tgjonestx:
The smoke test and acceptance tests were removing $SERVICED_VARPATH/volumes, which deleted the loop back files, but because the actual devicemapper devices were not disabled, the space for those loop back files was not released. This would periodically cause some of the Jenkins build slaves to run out of disk.  Rebooting the boxes was a brute force workaround.

These changes use the `serviced-storage` tool to disable the devicemapper device so that space is really freed when the test is complete.

### DIFF
--- a/acceptance.sh
+++ b/acceptance.sh
@@ -13,7 +13,10 @@
 #######################################################
 
 DIR="$(cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd)"
-SERVICED=${DIR}/serviced
+SERVICED=$(which serviced)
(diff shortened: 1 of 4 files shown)


alapidas:
Talked to Glenn about handing cleanup failures.

tgjonestx:
FYI - In cases where some of the tests fail, we still want to do as much cleanup as we can. so I modified the cleanup so that certain actions are done unconditionally without causing a build failure, but the filesystem/storage cleanup failures will cause a build failure.

PR #2468, opened 2016-02-05
move localAddrs map and initialization to type init instead of packag…

jplouis:
…e init


kwalkerz:
We should consider retrying once or twice if we get the "unable to use InterfaceAddrs" error, as it is usually very short-lived. glog.Fatal was reasonable when we were seeing it while running short-lived commands, but if this could abort the Master starting up it might be worth double-checking that this error persists...

PR #2473, opened 2016-02-08
Fixes CC-1877 "Public Endpoints with ports appears to open only 1 port"

johnlhamilton:
We now prevent the user from creating a port if the port is already in use.  Additionally, a deadlock has been removed that occurred if a newly-created port failed to start.

### DIFF
--- a/web/portservers.go
+++ b/web/portservers.go
@@ -67,14 +67,14 @@ func (sc *ServiceConfig) ServePublicPorts(shutdown <-chan (interface{}), dao dao
 	go sc.syncAllPublicPorts(shutdown)
 }
 
-func (sc *ServiceConfig) CreatePublicPortServer(publicEndpointKey service.PublicEndpointKey, stopChan <-chan int, shutdown <-chan (interface{})) {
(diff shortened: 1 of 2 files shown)


web/portservers.go
```diff
@@ -67,14 +67,14 @@ func (sc *ServiceConfig) ServePublicPorts(shutdown <-chan (interface{}), dao dao
 	go sc.syncAllPublicPorts(shutdown)
 }
 
-func (sc *ServiceConfig) CreatePublicPortServer(publicEndpointKey service.PublicEndpointKey, stopChan <-chan int, shutdown <-chan (interface{})) {
+func (sc *ServiceConfig) CreatePublicPortServer(publicEndpointKey service.PublicEndpointKey, stopChan <-chan int, shutdown <-chan (interface{})) error {
 	port := publicEndpointKey.Name()
 	listener, err := net.Listen("tcp", port)
 	stopChans := []chan bool{}
 	if err != nil {
 		glog.Errorf("Could not setup TCP listener - %s", err)
```
kwalkerz: Can you log the port and/or public endpoint key here (since the caller doesn't log it)?

PR #2478, opened 2016-02-10
Feature/cc 1885

rDr4g0n:
Update UI to use RAMLimit instead of RAMCommitment per:
https://jira.zenoss.com/browse/CC-1829
https://jira.zenoss.com/browse/CC-1885


acceptance/ui/features/data/default/hosts.json
```diff
@@ -25,7 +25,7 @@
       "hostID": 4982,
       "memory": 10737418240,
       "memoryGB": "10.00 GB",
-      "ramGB": "5.34 GB",
+      "ramGB": "5737418240",
```
tgjonestx: Is this really the value we show to the user?

web/ui/src/common/miscUtils.js
```diff
@@ -183,6 +192,37 @@
                 }, 0);
 
                 return childCount;
+            },
+
+            validateRAMLimit(limitStr, max=Infinity){
```
tgjonestx: Can you add some unit-tests for this method?

web/ui/src/common/miscUtils.js
```diff
@@ -183,6 +192,37 @@
...
+
+                var isPercent = (limitStr.endsWith("%"));
+
+                // if this is a percent, ensure its between 1 and 100
+                if(isPercent){
+                    let val = +limitStr.slice(0, -1);
+                    if(val > 100){
+                        return "RAM Limit cannot exceed 100%";
```
tgjonestx: Assuming these are error messages to be displayed to the user, do they need to be localized?

acceptance/ui/features/data/default/hosts.json
rDr4g0n: We are making it a point to keep the unit the user enters, so if they choose to use bytes, then they will always see bytes. They can go back in and change it to megs or gigs or whatever they prefer.

web/ui/src/common/miscUtils.js
rDr4g0n: I haven't been localizing phrases like these, mostly just sticking to nouns and verbs. In the future we can make a pass at localization (and automation of localization using google translate api)

PR #2515, opened 2016-02-18
briefly delay initial isvc healthchecks

tgjonestx:
Give the ZK process in the container a couple of seconds to start so that the underlying ZK client lib in serviced doesn't spam the CC log

Fixes [CC-1701](https://jira.zenoss.com/browse/CC-1701)


smousa:
Can you check the socket before establishing a client connection?

tgjonestx:
Yes - that's a better idea.
I'll close this PR, and replace with a different implementation

PR #2531, opened 2016-02-26
CC-1974: NFSClient disabled does not break serviced on a remote agent

smousa:
TODO: rename NFSDriver to something that doesn't say nfs
* Added an option on the storage driver to disable networking
* Use the option for when the nfs client is disabled
* Make the varpath configurable for isvcs and volumes on the agent

### DIFF
--- a/volume/nfs/nfs.go
+++ b/volume/nfs/nfs.go
@@ -26,12 +26,16 @@ import (
 	"github.com/control-center/serviced/volume"
 )
 
+const NetworkDisabled = "network_disabled"
+
 var (
 	ErrNotSupported = errors.New("not supported by nfs driver")
 )
 
+// TODO: name this better, since this is more of a passthrough
 type NFSDriver struct {
-	root string
+	networkDisabled bool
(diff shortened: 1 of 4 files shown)


wwwtyro:
LGTM

volume/nfs/nfs.go
```diff
@@ -98,11 +100,12 @@ func (d *NFSDriver) Get(volumeName string) (volume.Volume, error) {
 		driver: d,
 		tenant: getTenant(volumeName),
 	}
-	//actual NFS mount
-	if err := mount(volumeName, volumePath); err != nil {
-		return nil, err
+	if !d.networkDisabled {
```
jplouis: Mounting will create the directory if it isn't there and then mount the remote FS.  What  creates the directory on the local filesystem if the network is disabled?

smousa: setupVolume in node/agent.go

jplouis: Thanks!

PR #2533, opened 2016-02-29
Port to OSX

tgjonestx:
This does NOT allow CC to run on Mac OSX, but it does all it to compile and pass unit tests.

This does not mean that we can run CC on OSX, but it does allow early development on OSX for many parts of the code base.


Godeps/Godeps.json
```diff
@@ -331,7 +331,7 @@
 		},
 		{
 			"ImportPath": "github.com/zenoss/glog",
-			"Rev": "09a83f7c5c0f425c00cb1e063e9179698579f296"
+			"Rev": "c2ed1c00faf68d35a20ac43134f202268636d7d8"
```
kwalkerz: Seems like you need the last commit (https://github.com/zenoss/glog/commit/4ac6e9330b9c1f1e168b3660b789729a4716c469), which corrects a typo (missing ") in an import statment.

dfs/nfs/mounts_test.go
```diff
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// +build unit
+// +build unit, linux
```
kwalkerz: You have a couple of test files where you've limited it to linux. Shouldn't we provide _darwin versions of these? (I'll admit it makes it easier to pass on Mac OS. ;-) ) Maybe at least open a ticket to come back to this later...

Godeps/Godeps.json
tgjonestx: Good catch - thanks

dfs/nfs/mounts_test.go
tgjonestx: I don't have any OSX support for several things like NFS and devicemapper.  So if I follow the `_platform`  naming convention then we end up with many of the files in that package name `someName_linux.go` as compared to the `utils` package where the majority of the business logic is platform neutral

kwalkerz: My general concern is the further apart the platforms vary in their testing, the less valuable the non-linux platforms become. Maybe I'm confused about the purpose of this PR and why we want to be able to (not run but) test on OSX. Certainly you can install NFS on OSX. (Not sure about devmapper.)

tgjonestx: The point is to be able to do something on OSX as opposed to today's status quo, which is nothing.  There's no plan to include OSX in the PR or nightly builds so breakages will occur.
Even though breakages will occur, and even though several subsystems will not compile or test, there are several areas where someone could make some progress on some issues without the need to VPN into a Linux machine.

For instance, one could make great head way on across-the-board replacement of glog, or replacement of codegansta, or standardizing tests on gochk.  The work wouldn't be definitive or final until it could be built/tested on Linux, but again, it is more than could be done today on OSX.

domain/host/host.go
```diff
@@ -168,11 +168,13 @@ func Build(ip string, rpcport string, poolid string, memory string, ipAddrs ...s
 	}
 	host, err := currentHost(ip, rpcPort, poolid)
 	if err != nil {
+		glog.Errorf("currentHost failed: %s", err)
```
iancmcc: Could we do this one V(2)? We use this pattern throughout, but it (sometimes) leads to situations where we swallow or switch on an error at a higher level, but can't avoid the scary log message lower down.

domain/host/host.go
```diff
@@ -168,11 +168,13 @@ func Build(ip string, rpcport string, poolid string, memory string, ipAddrs ...s
...
 		return nil, err
 	}
 	glog.Infof("Building host %s (%s) with ipsAddrs: %v [%d]", host.ID, host.IPAddr, ipAddrs, len(ipAddrs))
 	hostIPs, err := getIPResources(host.ID, host.IPAddr, ipAddrs...)
 	if err != nil {
+		glog.Errorf("getIPResources failed: %s", err)
```
iancmcc: Ditto.

domain/host/utils.go
```diff
@@ -57,6 +57,7 @@ func currentHost(ip string, rpcPort int, poolID string) (host *Host, err error)
 	host.Name = hostname
 	hostidStr, err := utils.HostID()
 	if err != nil {
+		glog.Errorf("HostID failed: %s", err)
```
iancmcc: Ditto.

domain/host/utils.go
```diff
@@ -97,6 +98,7 @@ func currentHost(ip string, rpcPort int, poolID string) (host *Host, err error)
 
 	routes, err := utils.RouteCmd()
 	if err != nil {
+		glog.Errorf("RouteCmd failed: %s", err)
```
iancmcc: Ditto.

iancmcc:
Waiting for tests to pass, but otherwise looks good. Built and ran tests successfully on OS X with Go 1.6.

PR #2540, opened 2016-03-02
CC-1896: Protect zookeeper data from multithreaded access

kwalkerz:
=== DO NOT MERGE ===

Alternative investigation is pending...


smousa:
My concern here is you are synchronizing calls to zookeeper by calling lock.Lock instead of lock.RLock

iancmcc:
I have a few issues with this.

1) Why are we using concurrency control in the client, rather than Zookeeper itself? Any number of clients could be accessing these data simultaneously; I don't understand the benefit of managing concurrency within a client. That should be provided by ZK.

2) This is a huge change for something we only see in tests. Is there any more targeted diagnosis or fix we can implement?

smousa:
I think that the problem might stem from this code: 
https://github.com/control-center/serviced/blob/develop/zzk/service/pool.go#L67

Where we are creating a new empty node and then populating it with data. We used to have to create nodes this way because of a bug in the zookeeper library that was fixed a while back, but the fix may be that we need to create nodes like we do here:
https://github.com/control-center/serviced/blob/develop/zzk/service/service.go#L440

iancmcc:
Okay, after looking at this more, I have better comments. I get the read lock implementation, but I'm concerned about the overhead of locking at this low level. Can we lock specific things at the application level instead?

kwalkerz:
Summer has suggested a more targeted fix. Let me look at her suggestion.

kwalkerz:
Which library was that bug fixed in? And do we have that fix? (Maybe...is
that the library update you are blocked on me for?)

I'd say go ahead and finish your Go 1.6 work (stop waiting on me) and I'll
merge my stuff (whatever it ends up being) with yours.

On Thu, Mar 3, 2016 at 7:59 AM, smousa <notifications@github.com> wrote:

> I think that the problem might stem from this code:
>
> https://github.com/control-center/serviced/blob/develop/zzk/service/pool.go#L67
>
> Where we are creating a new empty node and then populating it with data.
> We used to have to create nodes this way because of a bug in the zookeeper
> library that was fixed a while back, but the fix may be that we need to
> create nodes like we do here:
>
> https://github.com/control-center/serviced/blob/develop/zzk/service/service.go#L440
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/control-center/serviced/pull/2540#issuecomment-191771531>
> .
>



-- 
*Ken Walker* | Technical Lead
kwalker@zenoss.com | 512.687.6859

<http://www.zenoss.com/galaxz/>

smousa:
Ken, I removed that comment because I was mistaken, but the change for that is here:
https://github.com/control-center/serviced/commit/30de663a207a5d82e696219517919ff426589d60

kwalkerz:
Ha! GitHub doesn't bother to send emails when comments are deleted... :-)
 Thanks.

On Thu, Mar 3, 2016 at 11:52 AM, smousa <notifications@github.com> wrote:

> Ken, I removed that comment because I was mistaken, but the change for
> that is here:
> 30de663
> <https://github.com/control-center/serviced/commit/30de663a207a5d82e696219517919ff426589d60>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/control-center/serviced/pull/2540#issuecomment-191885441>
> .
>



-- 
*Ken Walker* | Technical Lead
kwalker@zenoss.com | 512.687.6859

<http://www.zenoss.com/galaxz/>

kwalkerz:
We're not going this route.

PR #2541, opened 2016-03-02
Send verbose flag on into test cases

kwalkerz:
serviced-tests.py has a verbose flag, which it keeps to itself and does
not pass on to 'go test'. This PR makes it share that flag.

### DIFF
--- a/serviced-tests.py
+++ b/serviced-tests.py
@@ -218,6 +218,9 @@ def main(options):
     if usep1:
         cmd.extend(['-p', '1'])
 
+    if options.verbose:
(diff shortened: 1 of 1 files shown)


wwwtyro:
I made a PR identical to this, but it got shot down. Can't for the life of me remember why. We can instead pass args to go test after a `--`.   So in this case, `-- -v`.

iancmcc:
That's right. The verbosity flag is different depending on the tests you're running. That's why we kept it out of the options we pulled up to the front, and left it up to the passthrough options.

PR #2547, opened 2016-03-09
New TestableMutex class

kwalkerz:
Similar to Go's sync.Mutex, but provides:
- A method for trying to acquire the lock without blocking.
- A field to indicate the current holder of the lock.

Rationale: During tasks like backing up the system, we don't want data in ES or ZK to be changed. But tasks started by the user (like deleting a pool) should not just block--would prefer to report "busy performing backup now, try again later."

This implementation does not rely on any specific technologies, so putting it in the commons folder for potential use by others.


kwalkerz:
retest this

PR #2551, opened 2016-03-09
New TimedMutex class

kwalkerz:
Similar to Go's sync.Mutex, but provides:
- A method that allows a limit on the amount of time you want to
  wait on the lock request.
- A field that indicates who is currently holding the lock.

Rationale: During long-running tasks like backing up the system, we
don't want user-initiated tasks like deleting a pool to block. We'd
rather pass back some error like "busy doing X, try again later."
(Where X is filled in from the TimedMutex's field value.)

The implementation does not rely on any specific technologies, so
putting it in the commons directory for potential use by others.


kwalkerz:
retest this\

commons/sync/timedmutex.go
```diff
@@ -0,0 +1,86 @@
...
+package sync
+
+import (
+	"sync/atomic"
+	"time"
+)
+
+// A TimedLocker represents an object that can be locked and unlocked.
+type TimedLocker interface {
+	Lock(string)
```
iancmcc: I guess there's no way to embed the Locker interface here, because of the addition of a Lock(string) method.

commons/sync/timedmutex.go
```diff
@@ -0,0 +1,86 @@
...
+	LockWithTimeout(name string, timeoutMs int) (gotLock bool, holder string)
```
johnlhamilton: It may be better to pass the timeout as a time.Duration object so callers know exactly what to send.

iancmcc: Agreed. That's also a pretty common pattern in Goville. It's Gothonic. I don't know the correct adjective.

kwalkerz: Good point--I will fix that.

commons/sync/timedmutex.go
kwalkerz: Yeah, I don't see how to do it without subverting part of the reason for using this class over sync.Mutex.

kwalkerz:
retest this

kwalkerz:
retest this

kwalkerz:
retest this

sigh.......

PR #2552, opened 2016-03-10
Revert "Revert "CC-1925: Add option to change logstash purge frequency""

vikt0rs:
Reverts control-center/serviced#2545

... and yes, I know that this looks funny.


tgjonestx:
@vikt0rs - the comments in #2545 say it was reverted to fix some failed tests, so looking at this PR I kind of expected some additional changes related to tests in some way, but I don't see.   What tests failed because of #2545 and why isn't there a testing concern with this PR?

vikt0rs:
@tgjonestx - these failures were not fixed by my revert, so I supposed that that was an another error and propose this change to CC "as is"

PR #2556, opened 2016-03-10
Fixes ZEN-19791

wwwtyro:


web/cpserver.go
```diff
@@ -144,7 +144,7 @@ func (sc *ServiceConfig) Serve(shutdown <-chan (interface{})) {
 		} else {
 			glog.V(2).Infof("httphost: calling uiHandler")
 			if r.TLS == nil {
-				http.Redirect(w, r, fmt.Sprintf("https://%s:%s", r.Host, sc.bindPort), http.StatusMovedPermanently)
+				http.Redirect(w, r, fmt.Sprintf("https://%s:%s", r.Host, strings.Split(sc.bindPort, ":")[1]), http.StatusMovedPermanently)
```
smousa: This looks risky.  What if there isn't a colon?  This would panic

web/cpserver.go
```diff
@@ -189,7 +189,7 @@ func (sc *ServiceConfig) Serve(shutdown <-chan (interface{})) {
 
 	go func() {
 		redirect := func(w http.ResponseWriter, req *http.Request) {
-			http.Redirect(w, req, fmt.Sprintf("https://%s:%s%s", req.Host, sc.bindPort, req.URL), http.StatusMovedPermanently)
+			http.Redirect(w, req, fmt.Sprintf("https://%s:%s%s", req.Host, strings.Split(sc.bindPort, ":")[1], req.URL), http.StatusMovedPermanently)
```
smousa: See previous note

web/cpserver.go
wwwtyro: We already validate it here: https://github.com/control-center/serviced/blob/develop/validation/validators.go#L92

PR #2563, opened 2016-03-15
Fixes CC-2007

wwwtyro:


tgjonestx:
I tried to find other places where we called `os.Stat()` without checking the error code, and that was the only one.

PR #2564, opened 2016-03-15
no more ram values without % or units

sockterrier:


tgjonestx:
The changes look fine, but they might break the acceptance tests.  There is a reference to the error message "Invalid RAM Limit value" in `acceptance/ui/features/hosts.feature`. I _think_ the tests should pass because it shouldn't look for complete match.  Can you please run acceptance.sh with `--tags @hosts` to confirm?

PR #2565, opened 2016-03-15
Feature/zen 21575

rphillipsz:
Commit 1) Fix acceptance tests for Ubuntu 15.10
Commit 2) Split the host:rpcport into two fields in the UI
Commit 3) Update the cucumber tests


web/ui/src/DeployWiz/DeployWizard.js
```diff
@@ -210,7 +214,9 @@
         };
 
         $scope.addHostStart = function() {
-            $scope.newHost = {};
+            $scope.newHost = {
+                port: '4979'
```
tgjonestx: Please replace 4979 with something like `$translate.instant("placeholder_port")`

web/ui/src/DeployWiz/DeployWizard.js
```diff
@@ -113,7 +115,9 @@
             // if there is not at least one host, add an
             // "add host" step to the wizard
             if(hostsFactory.hostList.length === 0){
-                $scope.newHost = {};
+                $scope.newHost = {
+                    port: '4979'
```
tgjonestx: Please replace 4979 with something like `$translate.instant("placeholder_port")`

tgjonestx:
:+1: after those 2 minor comments.

PS. I think Summer is close to fixing the bug that's preventing the current tests from passing.

web/ui/src/common/miscUtils.js
```diff
@@ -207,6 +207,14 @@
                 return childCount;
             },
 
+            validateHostName: function(hostStr){
```
tgjonestx: Actually, we need to add another validation. For Port, the only valid values are positive, non-zero, unsigned 16-bit ints

rphillipsz: did the rpc port also need to be >1024?  I know those up to 1024 are priviledged ports that can only be bound by root, but not sure if the rpcport is setup as root or not

tgjonestx: Let's allow restricted ports for now since I'm not sure if we cases that need to override them or not.

rphillipsz:
retest this, please

PR #2566, opened 2016-03-16
CC-2020: Ensure dir nodes are empty

smousa:
https://jira.zenoss.com/browse/CC-2020

### DIFF
--- a/coordinator/client/zookeeper/connection.go
+++ b/coordinator/client/zookeeper/connection.go
@@ -22,13 +22,6 @@ import (
 	"github.com/control-center/serviced/coordinator/client"
 )
 
-type dir struct {
-	version interface{}
-}
-
-func (n *dir) SetVersion(version interface{}) { n.version = version }
-func (n *dir) Version() interface{}           { return n.version }
-
 // Connection is a Zookeeper based implementation of client.Connection.
 type Connection struct {
 	sync.RWMutex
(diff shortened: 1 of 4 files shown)


coordinator/client/zookeeper/connection.go
```diff
@@ -159,43 +152,23 @@ func (c *Connection) CreateDir(path string) error {
...
-	return c.create(p, &dir{})
+	pth := path.Join(c.basePath, p)
+	_, err := c.conn.Create(pth, []byte{}, 0, zklib.WorldACL(zklib.PermAll))
+	return xlateError(err)
 }
 
 func (c *Connection) ensurePath(p string) error {
 	dp := path.Dir(p)
-	if ok, err := c.exists(dp); err != nil {
+	if p == "" || p == "/" {
```
kwalkerz: Shouldn't you be checking dp here instead of p?

coordinator/client/zookeeper/connection_test.go
```diff
@@ -23,7 +23,7 @@ import (
 	"time"
```
kwalkerz: We should probably add a test like Create("/a/b/c/d/e") to validate ensurePath().

coordinator/client/zookeeper/connection.go
smousa: path.Dir("/') and path.Dir("") are reflective.

To your second comment, I agree, and will fix.  Thanks!

coordinator/client/zookeeper/connection.go
```diff
@@ -159,11 +159,13 @@ func (c *Connection) createDir(p string) error {
...
-	// path.Dir returns "." if the argument is ""
-	if dp == "/" || dp == "." {
+	// check p instead of dp because if the path is /a/b, we still need to make
+	// sure /a is created and if we return nil because the dirpath is root and
+	// not the node itself, then the node will not get created below.
+	if p == "/" || p == "" {
 		return nil
 	}
-	if exists, err := c.exists(dp); err != nil && err != client.ErrNoNode {
+	if exists, err := c.exists(dp); err != nil {
```
smousa: FYI: exists function swallows ErrNoNode error (see below)

PR #2567, opened 2016-03-16
CC-1896: Protect long-running ops from changing data

kwalkerz:
The original problem seen: When the scheduler was syncing Zookeeper wtih Elasticsearch, and you deleted a pool during the sync, the sync half-recreated the pool. This caused problems like a listener trying to spawn would loop on trying to use the zombie pool.

The solution implemented here: Use the DFS's lock to prevent significant data from changing during Zookeeper syncs, backup/restore, and snapshot/rollback. I changed the DFS lock type here to our new TimedMutex, which allows operations initiated by the user to return quickly if they cannot grab the lock and print a "busy, try again later" message (rather than seeming to hang).

I did not change some existing places where the DFS lock is obtained (for example, ListSnapshots) to use LockWithTimeout instead because they are called from some of the long-running operations which probably expect to block and I did not want to change their behavior.


dao/elasticsearch/service.go
```diff
@@ -52,11 +79,25 @@ func (this *ControlPlaneDao) CloneService(request dao.ServiceCloneRequest, clone
...
+	}
+	defer this.facade.DFSLock(datastore.Get()).Unlock()
+
 	return this.facade.UpdateService(datastore.Get(), svc)
 }
 
 //
 func (this *ControlPlaneDao) MigrateServices(request dao.ServiceMigrationRequest, unused *int) error {
+	if gotLock, blocker := this.facade.DFSLock(datastore.Get()).LockWithTimeout("migrate service", userLockTimeout); !gotLock {
+		err := errors.New(fmt.Sprintf(userLockTimeoutMessage, blocker))
```
smousa: instead of errors.New, can you instead create an error type?

kwalkerz: My problem is the %s in the message I want to fill in with the operation that is blocking you from obtaining the lock. Is there a way to fill that in later when the error has been predefined?

smousa: To create an error type in go, the struct you create has to implement Error() string

smousa: https://golang.org/pkg/errors/

dao/elasticsearch/service.go
```diff
@@ -25,14 +24,21 @@ import (
...
+	blocker string
+}
+
+func (e ErrSystemBusy) Error() string {
+	return fmt.Sprintf("Cannot interrupt system operation (%s). Try again later.", e.blocker)
+}
+
 // AddService adds a new service. Returns an error if service already exists.
 func (this *ControlPlaneDao) AddService(svc service.Service, serviceId *string) error {
 	if gotLock, blocker := this.facade.DFSLock(datastore.Get()).LockWithTimeout("add service", userLockTimeout); !gotLock {
```
smousa: I feel like LockWithTimeout should return error instead of blocker

facade/common.go
```diff
@@ -43,6 +43,14 @@ func (f *Facade) afterEvent(event afterEvent, eventCtx eventContext, entity inte
 type beforeEvent string
 type afterEvent string
 
+type ErrSystemBusy struct {
```
smousa: Shouldn't have two declarations of the same error

dao/elasticsearch/service.go
kwalkerz: I wanted TimedMutex to be more generic and usable in multiple situations. The 'holder' field does not have to be a human-readable string for use in error messages--I'm just using it that way here. (And that's why the errors are defined at the application level, not within the mutex.)

PR #2568, opened 2016-03-16
blank is 100%

sockterrier:


web/ui/test/common/miscUtilsSpec.js
```diff
@@ -179,11 +179,13 @@ describe('miscUtils', function() {
...
 
     describe("validateRAMLimit", function(){
+        it("Validates an empty string", function(){
+            expect(miscUtils.validateRAMLimit("")).toBe(null);
+        });
         it("Validates percentage values", function(){
             expect(miscUtils.validateRAMLimit("50%")).toBe(null);
         });
         it("Validates various byte values", function(){
-            expect(miscUtils.validateRAMLimit("972")).toBe(null);
```
tgjonestx: Why isn't "972" a valid byte value?

PR #2569, opened 2016-03-16
Add ability to override images in docker registry

johnlhamilton:
Fixes CC-1997

### DIFF
--- a/cli/cmd/docker.go
+++ b/cli/cmd/docker.go
@@ -50,7 +50,7 @@ func (c *ServicedCli) initDocker() {
 			}, {
 				Name:        "override",
 				Usage:       "Replace an image in the registry with a new image",
-				Description: "serviced docker override NEWIMAGE OLDIMAGE",
+				Description: "serviced docker override OLDIMAGE NEWIMAGE",
 				Action:      c.cmdDockerOverride,
 			},
 		},
@@ -91,7 +91,10 @@ func (c *ServicedCli) cmdDockerOverride(ctx *cli.Context) {
 		return
 	}
 
-	if err := c.driver.DockerOverride(args[0], args[1]); err != nil {
+	oldImage := args[0]
+	newImage := args[1]
+
+	if err := c.driver.DockerOverride(newImage, oldImage); err != nil {
 		fmt.Fprintln(os.Stderr, err)
 	}
 }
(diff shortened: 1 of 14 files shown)


cli/cmd/docker.go
```diff
@@ -44,6 +47,11 @@ func (c *ServicedCli) initDocker() {
 					cli.StringFlag{"registry", "", "host:port where the registry is running"},
 					cli.BoolFlag{"override, f", "overrides all existing image records"},
 				},
+			}, {
+				Name:        "override",
+				Usage:       "Replace an image in the registry with a new image",
+				Description: "serviced docker override NEWIMAGE OLDIMAGE",
```
kwalkerz: To match typical Unix command style, should we switch this to "OLDIMAGE NEWIMAGE". cp, mv, ln, even diff kind of, all use this pattern.

johnlhamilton: Yeah, that would be a little more intuitive.  I'll make that change

dfs/override.go
```diff
@@ -0,0 +1,50 @@
...
+	newImage, err := dfs.docker.FindImage(newimg)
+	if err != nil {
+		glog.Errorf("Could not find replacement image %s: %s", newimg, err)
+		return err
+	}
+
+	// push the image into the registry
+	hash, err := dfs.docker.GetImageHash(newImage.ID)
+	if err != nil {
+		glog.Errorf("Could not get has for image %s: %s", newimg, err)
```
kwalkerz: Missing h in "hash"

dfs/override.go
```diff
@@ -0,0 +1,50 @@
...
+		return err
+	}
+
+	if err := dfs.index.PushImage(oldImage.String(), newImage.ID, hash); err != nil {
+		glog.Errorf("Could not push image %s (%s): %s", oldImage, newImage.ID, err)
```
kwalkerz: Can we make this error message a little clearer? More along the lines "Could not replace image A with B"...

facade/dfs.go
```diff
@@ -637,3 +637,9 @@ func (info *registryVersionInfo) start(isvcsRoot string, hostPort string) (*dock
 
 	return container, nil
 }
+
+// DockerOverride will replace a docker image in the registry with a new image for all services that use
+//  it under the given tenant
```
kwalkerz: Is this comment accurate? There's nothing in the API that limits the image to a specific tenant...

johnlhamilton: Nope, leftover from an earlier approach.  I'll remove the part about a given tenant.

PR #2572, opened 2016-03-16
valid value

sockterrier:


tgjonestx:
:+1: Should be good to go once acceptance tests pass

PR #2575, opened 2016-03-17
ZEN-22546 Add uintToInt cast function

carlcamera:
add cast function for use in GO templates

### DIFF
--- a/domain/service/evaluate.go
+++ b/domain/service/evaluate.go
@@ -196,6 +196,7 @@ func (service *Service) evaluateTemplate(gs GetService, fc FindChildService, ins
 		"percentScale":  percentScale,
 		"bytesToMB":     bytesToMB,
 		"plus":          plus,
+		"uintToInt":     uintToInt,
 		"each":          each,
 	}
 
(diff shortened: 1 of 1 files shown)


jpeacock-zenoss:
I suppose the user of `unitToInt` just needs to be careful about using `uint` values bigger than `int`.

iancmcc:
This isn't the only place those functions are defined for templates. There are a couple others you should look for.

carlcamera:
So @iancmcc are you saying
1. This change is unacceptable until those other locations are found and also updated or
2. Know Thee Well that other template function files exist and Carry On.

I need this function for the MariaDB-model template change and it works, so not sure how to proceed in light of your comment above.

iancmcc:
The former. Unless it's unsuited to the other template evaluators.

iancmcc:
The other template evaluators got consolidated, so no other changes are necessary.

carlcamera:
Update: Ian checked and someone has consolidated the GO functions into just the one evaluate.go file so only one file needs updating after all.

PR #2578, opened 2016-03-17
CC-2024: empty pool node can still be modified to store data

smousa:


acceptance/ui/features/steps/pools_steps.rb
```diff
@@ -211,28 +220,23 @@ def addPoolJson(pool)
...
 def removeAllPoolsExceptDefault()
     visitApplicationsPage()
     removeAllEntries("service")
     removeAllHostsCLI()
     removeAllPoolsCLI()
-    addDefaultPool()
     refreshPage()
 end
 
 def removeAllPoolsCLI()
```
kwalkerz: Can we rename this to 'removeAllPoolsExceptDefaultCLI' now? Because now it does not remove all the pools...

facade/pool.go
```diff
@@ -323,6 +324,12 @@ func (f *Facade) RemoveResourcePool(ctx datastore.Context, id string) error {
 	}
 	defer f.DFSLock(ctx).Unlock()
 
+	// CC-2024: do not delete the default resource pool
+	if id == "default" {
```
kwalkerz: Why would we not let the user delete the default pool if they want to?

acceptance/ui/features/steps/pools_steps.rb
smousa: yes

facade/pool.go
smousa: Because it causes bad things to happen to the system

smousa: And the system will try to add it back on restart anyway.

smousa:
retest this

kwalkerz:
Looks good, once tests pass...

PR #2584, opened 2016-03-21
fix for go 1.6: metrics caching to use pointers to locks

johnlhamilton:
Port of #2529


smousa:
You need a support ticket for this before this can go into the support branch

PR #2586, opened 2016-03-21
CC-1991: Unmount NFS on agent shutdown

kwalkerz:
To avoid stale NFS mounts during upgrades. May also help with normal
shutdown and startup of agents?


volume/nfs/nfs.go
```diff
@@ -157,7 +157,14 @@ func (d *NFSDriver) Status() (*volume.Status, error) {
 
 // Release implements volume.Driver.Release
 func (d *NFSDriver) Release(volumeName string) error {
-	return ErrNotSupported
+	volumePath := filepath.Join(d.root, volumeName)
```
tgjonestx: This is a nit - but there's no need to create `volumePath` if the network is disabled. Can you move inside the `if` block please?

dfs/nfs/common.go
```diff
@@ -205,3 +205,22 @@ func Mount(driver Driver, remotePath, localPath string) error {
 
 	return nil
 }
+
+// Unmount attempts to unmount the localPath
+func Unmount(driver Driver, localPath string) error {
```
tgjonestx: Can you add some unit-tests for this method?

PR #2587, opened 2016-03-21
Allowing convenient cron job to cleanup unused blocks

achow-zenoss:


pkg/serviced-fstrim
```diff
@@ -0,0 +1,26 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.        
+# See the License for the specific language governing permissions and          
+# limitations under the License. 
+
+# this convenient script is intended to be run as SU by a cron job.
+# It discards the blockes unused by the mounted dfs volumes of serviced
+# It should not run more often than necessary.
+# Please refer to the description of fstrim man page.
+
+which fstrim || { echo "Error:fstrim not found! Please install." >&2; exit 1; }
```
iancmcc: `which fstrim &> /dev/null`

pkg/serviced-fstrim
```diff
@@ -0,0 +1,26 @@
...
+
+for i in $(mount | grep '/opt/serviced/var/volumes' | cut -d ' ' -f 1)
```
iancmcc: Instead of running `mount`, use `grep /etc/mtab`. Cheaper.

pkg/serviced-fstrim
```diff
@@ -0,0 +1,36 @@
...
+# limitations under the License. 
+
+# this convenient script is intended to be run as SU by a cron job.
+# It discards the blockes unused by the mounted dfs volumes of serviced
+# It should not run more often than necessary.
+# Please refer to the description of fstrim man page.
+which fstrim>/dev/null 2>&1 || { echo "Error:fstrim not found! Please install." >&2; exit 1; }
+
+# capture block devices of the related serviced volumes
+blk_devs=$(lsblk -o MOUNTPOINT,TYPE | grep '/opt/serviced/' | grep ' dm' | sort | uniq)
```
iancmcc: This is not specific enough. There will frequently be dm mounts under /opt/serviced to which this script isn't applicable.

iancmcc: I suggest you use `lsblk --output` to specify the columns you want to parse, then get them directly. Merely searching for the string " dm" is too easy to fail.

iancmcc: I suggest using a more targeted regex, such as `/opt/serviced/var/volumes/[^/]+/?\s+dm`, and also passing `--noheadings` or whatever the options is to lsblk.

pkg/serviced-fstrim
```diff
@@ -0,0 +1,36 @@
...
+
+# no need for fstrim if serviced volumes are not on device-mapper
+(( $? == 0 )) ||    { echo "Serviced volumes not on device-mapper" >&2; exit 0;  }
+
+# handle only the mounted serviced volumes
+for i in $(mount | grep '/opt/serviced/var/volumes' | cut -d ' ' -f 1)
```
iancmcc: Why re-grep? just use the value of `blk_devs`.

pkg/serviced-fstrim
achow-zenoss: Yes. Got it.

pkg/serviced-fstrim
achow-zenoss: Yup. Thx

PR #2591, opened 2016-03-22
CC-2039 - configurable UI polling interval

rDr4g0n:
https://jira.zenoss.com/browse/CC-2039


web/resources.go
```diff
@@ -924,6 +924,11 @@ func restGetStorage(w *rest.ResponseWriter, r *rest.Request, client *node.Contro
 	w.WriteJson(storageInfo)
 }
 
+func restGetUIConfig(w *rest.ResponseWriter, r *rest.Request, client *node.ControlClient) {
+	fmt.Printf("writing uiConfig %v", uiConfig)
```
rDr4g0n: Whoops, need to remove this debug log message

smousa: good work john matos... please remove printf

cli/api/options.go
```diff
@@ -246,6 +247,7 @@ func GetDefaultOptions(config utils.ConfigReader) Options {
 		DockerLogDriver:      config.StringVal("DOCKER_LOG_DRIVER", "json-file"),
 		DockerLogConfigList:  config.StringSlice("DOCKER_LOG_CONFIG", []string{"max-file=5", "max-size=10m"}),
 		AllowLoopBack:        strconv.FormatBool(config.BoolVal("ALLOW_LOOP_BACK", false)),
+		UIPollFrequency:      config.IntVal("UI_POLL_FREQUENCY", 3000),
```
tgjonestx: Are the units really in milliseconds?  Shouldn't they be in something more user-friendly like seconds?

tgjonestx: Also, please add SERVICED_UI_POLL_FREQUENCY to pkg/serviced.default.

tgjonestx:
What about the back-off strategy on the UI side when a request takes longer than the polling frequency?

rDr4g0n:
@tgjonestx the backoff change is a separate thing. This is a smaller change that will be ported to 1.1.3, where the backoff strategy change will go to 1.2.

PR #2594, opened 2016-03-23
Make sure we have images locally before push and snapshot

johnlhamilton:
Fixes CC-2029
On push, if we can't find the image locally by ID, we look for it with the following methods:
1.) Look for it locally by registry address, and compare hashes
2.) Attempt to pull it from the registry, then compare IDs or hashes
3.) Search all local images for one with a matching hash

Additionally, we now search for each image using the above method before performing a snapshot, and return an error if any of them can't be found.  This prevents an error later when a snapshot that was taken with missing images is restored.

Both of these changes address issues that occur when the master does not have all of the images locally, even if they exist in the registry.  This state could occur after an HA fail-over, or if the user switched their docker storage.

### DIFF
--- a/dfs/docker/docker.go
+++ b/dfs/docker/docker.go
@@ -67,7 +67,7 @@ type Docker interface {
 	CommitContainer(ctr, image string) (*dockerclient.Image, error)
 	GetImageHash(image string) (string, error)
 	GetContainerStats(containerID string, timeout time.Duration) (*dockerclient.Stats, error)
-	FindImageByHash(imageHash string) (*dockerclient.Image, error)
+	FindImageByHash(imageHash string, checkAllLayers bool) (*dockerclient.Image, error)
 }
 
 type DockerClient struct {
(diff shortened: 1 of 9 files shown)


dfs/docker/docker.go
```diff
@@ -232,3 +233,28 @@ func (d *DockerClient) GetContainerStats(containerID string, timeout time.Durati
 
 	return result, nil
 }
+
+// FindImageByHash searches all local images for an image with the given hash
+func (d *DockerClient) FindImageByHash(imageHash string) (*dockerclient.Image, error) {
+	opts := dockerclient.ListImagesOptions{
+		All:     false,
```
smousa: why is All set to false?

johnlhamilton: I was afraid that setting this to true would be too expensive, since this would compute hashes for all layers.  I will change this to be a passed-in parameter, and we will call it twice from FindImages:  Once set to false, then if we don't find it we will call it again with it set to true, saving the most expensive call for the last-ditch effort.

johnlhamilton:
retest

johnlhamilton:
retest this

PR #2601, opened 2016-03-25
Fixes CC-1906

wwwtyro:
Demo:

In an isvcs container, `stress --cpu 3 --timeout 120` was run, resulting in this graph:

screenshot from 2016-03-25 08 51 58

In a service container (redis), `stress --cpu 3 --timeout 120` was run, resulting in this graph:

screenshot from 2016-03-25 09 03 51

### DIFF
--- a/stats/stats.go
+++ b/stats/stats.go
@@ -300,6 +300,7 @@ func (sr *StatsReporter) updateStats() {
 			}
 
 			// CPU Stats
+			// TODO: Consolidate this into a single object that both ISVCS and non-ISVCS can use
 			var (
 				kernelCPUPercent float64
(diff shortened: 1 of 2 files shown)


stats/stats.go
```diff
@@ -318,6 +318,11 @@ func (sr *StatsReporter) updateStats() {
 			}
 			sr.previousStats[key]["totalCPU"] = totalCPU
 
+			if totalCPUChange == 0 {
```
johnlhamilton: Why not <= like in isvcs/container.go?

tgjonestx:
Is it practical to write some kind of integration test for this?

PR #2604, opened 2016-03-25
CC-2033: Add a cache layer to reduce load on the server per browser polling the UI

iancmcc:
This led to some refactoring -- namely, cleaning up the health check registry, and moving some stuff from the dao layer to the facade layer.

Control Center ZenPack will need to be updated, because we broke the health check REST API. A ticket has been opened to track that: https://jira.zenoss.com/browse/ZEN-22714


web/status.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+	return
+}
+
+func restGetConciseServiceStatus(w *rest.ResponseWriter, r *rest.Request, client *node.ControlClient) {
+	f := func() []byte {
+		statuses, err := getAllServiceStatuses(client)
+		if err != nil {
+			glog.Errorf("Error retrieving service statuses: (%s)", err)
+			restServerError(w, err)
```
tgjonestx: I think we're going to hit a runtime error in this case - at a minimum the outcome is somewhat ambiguous. 

Consider the scenario where the cached value is empty, and the system calls `f` which then calls `restServerError()`. `restServerError()` is going to write an error payload to `w`, then `f` is going to continue, and marshal a nil value for `statuses` (which might panic?) and then return a nil value which then is passed to `w.Write()` which will do what? maybe overwrite the error payload that  `restServerError()` wrote? IDK.

Seems like it would be cleaner to have `f()` return an `error` as well as `[]byte` just to eliminate the ambiguity.

web/status_test.go
```diff
@@ -0,0 +1 @@
+package web
```
tgjonestx: Missing tests?

web/ui/src/Health/healthIconDirective.js
```diff
@@ -6,8 +6,10 @@
     'use strict';
 
     angular.module('healthIcon', [])
-    .directive("healthIcon", [
-    function(){
+    .directive("healthIcon", ["hcStatus",
+    function(hcStatus){
+        let {OK, FAILED, NOT_RUNNING, UNKNOWN} = hcStatus;
```
smousa: Where is TIMED_OUT?

web/ui/src/Health/serviceHealth.js
```diff
@@ -2,103 +2,121 @@
...
-    factory("$serviceHealth", ["$rootScope", "$q", "resourcesFactory", "$translate",
-    function($rootScope, $q, resourcesFactory, $translate){
+	// OK means health check is passing
+	const OK = "passed";
+	// Failed means health check is responsive, but failing
+	const FAILED = "failed";
+	// Timeout means health check is non-responsive in the given time
+	const TIMEOUT = "timeout";
+	// NotRunning means the instance is not running
+	const NOT_RUNNING = "not_running";
```
smousa: not_running should be "not running".  Need to change one or the other

web/ui/src/Health/healthIconDirective.js
rDr4g0n: I did not implement it this time around because it was not present before, and I wanted to minimize the changes. it should definitely be implemented in the future.

iancmcc:
retest this, please

PR #2607, opened 2016-03-28
Add fstrim path for centos

achow-zenoss:


hyangzenoss:
The change looks good and works on Centos

PR #2610, opened 2016-03-29
CC-2059 Deprecated /exports/ paths deleted

ed-:
Only remove /exports/serviced_var_volumes if it is not empty, even after stopping NFS.

### DIFF
--- a/dfs/nfs/server.go
+++ b/dfs/nfs/server.go
@@ -384,32 +384,35 @@ func (c *Server) cleanupBindMounts() {
 		}
 	}
 	// CC-1816: clean up remnants of old /exports/serviced_var_volumes directory
-	removeDeprecated("/exports/serviced_var_volumes")
+	c.removeDeprecated("/exports/serviced_var_volumes")
 }
 
 // removeDeprecated will clean up any old exports path
-func removeDeprecated(dirpath string) {
+func (c *Server) removeDeprecated(dirpath string) {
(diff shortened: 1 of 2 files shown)


dfs/nfs/server_test.go
```diff
@@ -232,3 +246,57 @@ func TestWriteExports(t *testing.T) {
...
+	}()
+
+	s1 := Server{
+		network:      "1.2.3.4/8",
+		basePath:     path.Join(emptyTempDir, "baseDir"),
+		exportedName: "foobar",
+	}
+
+	// deprecated dir is empty, deleted with no problem
+	assertPathExists(t, emptyTempDir)
```
tgjonestx: This doesn't hurt anything, but doesn't seem necessary as it basically verifies that `ioutil.TempDir()` works as advertised.

dfs/nfs/server.go
```diff
@@ -384,32 +384,35 @@ func (c *Server) cleanupBindMounts() {
...
 // removeDeprecated will clean up any old exports path
-func removeDeprecated(dirpath string) {
+func (c *Server) removeDeprecated(dirpath string) {
 	if dirContents, err := ioutil.ReadDir(dirpath); !os.IsNotExist(err) {
 		if err != nil {
 			glog.Warningf("Could not look up deprecated exports path %s: %s", dirpath, err)
 			return
 		}
+		// Stop NFS
+		stop()
```
jplouis: Where is NFS restarted?

PR #2613, opened 2016-03-29
Feature/zen 21574 support

sockterrier:


jplouis:
This is all front end?  What about adding a host via the CLI?  Does this change the value that gets saved in the DB?

sockterrier:
All front end. Adding a host via the CLI is unchanged. This only prevents raw bytes being used in the GUI, and stores 100% when no value was entered.

PR #2623, opened 2016-03-31
Remove service template methods from dao

tgjonestx:
This is another step in the ongoing quest to eliminate dao.   This PR removes methods and tests related to service templates from dao. The RPC mechanisms formerly in dao and node are now in rpc/master. The tests are now in facade.   The REST methods for service templates call the facade directly.


smousa:
Looks fine, I just had a question about how we present the RPC client methods:

Is it better to pass in request objects as opposed to passing in individual parameters?

tgjonestx:
For the new methods in rpc/master/interface.go, I tried to follow the pattern established by the Hosts and Pools methods where the method signature in the interface hides the fact that the underlying RPC has to use request and response objects.   The Request/Response params are basically a constraint imposed by GO - no need to impose the same restraint on our application-level code in places like cli/api.

tgjonestx:
Not to pile on, but here's another thought. The fact that the `dao.ControlPlane` interface follows the GO RPC paradigm makes it a leaky abstraction.

PR #2631, opened 2016-04-06
CC-2049, CC-2060: Rewrite backups to operate on a single tar stream.

iancmcc:
Backwards compatible with the previous backups.

This fixes tickets:
https://jira.zenoss.com/browse/CC-2049
https://jira.zenoss.com/browse/CC-2060


dfs/backup.go
```diff
@@ -34,30 +37,18 @@ const (
...
-	if _, err := buffer.WriteTo(tarfile); err != nil {
-		glog.Errorf("Could not write backup metadata: %s", err)
+
+	// A set of pipes that output tar streams
+	var tarpipes []pipe.Pipe
+
+	// write the backup metadata. This is too tiny to parallelize
+	metadataBuffer := &bytes.Buffer{}
+	if err := dfs.writeBackupMetadata(data, metadataBuffer); err != nil {
+		glog.V(2).Infof("Unable to write backup metadata: %s", err)
```
smousa: should this be an errorf?

dfs/backup.go
```diff
@@ -82,7 +74,7 @@ func (dfs *DistributedFilesystem) Backup(data BackupInfo, w io.Writer) error {
 			return err
 		}
 		// load the images from this snapshot
-		glog.Infof("Preparing images for tenant %s", info.TenantID)
+		glog.V(2).Infof("Preparing images for tenant %s", info.TenantID)
```
smousa: I don't know if V(2) is the way to go for this.

dfs/backup.go
iancmcc: That's a good question, and I don't know the right answer. The reason I made it not an errorf is because it ends up being double-logged then, once here, and once when the error gets returned up the stack. I think the right approach is debug log it at this low level, and error log it at the high level.

dfs/dfs.go
```diff
@@ -54,7 +54,7 @@ type DFS interface {
 	// Backup saves and exports the current state of the system
 	Backup(info BackupInfo, w io.Writer) error
 	// Restore restores the system to the state of the backup
-	Restore(r io.Reader) (*BackupInfo, error)
+	Restore(r io.Reader, backupInfo *BackupInfo) error
```
smousa: Why did you change the signature here?

dfs/restore.go
```diff
@@ -18,89 +18,82 @@ import (
...
-var ErrRestoreNoInfo = errors.New("backup is missing metadata")
+var (
+	ErrRestoreNoInfo        = errors.New("backup is missing metadata")
+	ErrInvalidBackupVersion = errors.New("backup has an invalid version")
+)
+
+// ExtractBackupInfo extracts the backup metadata from a tarball on disk in as
+// cheaply a manner as possible. The serialized BackupInfo is stored at the
+// front of the tarball to facilitate this.
+func ExtractBackupInfo(filename string) (*BackupInfo, error) {
```
smousa: I don't think this makes sense to be in this file.  I think this should probably go into backupinfo.go

dfs/dfs.go
iancmcc: Because we can extract the BackupInfo most cheaply from the file, rather than from the stream. Otherwise we have to recreate the stream after getting out the info. It seemed silly to pull it out to check the version and then pull it out again inside the restore.

dfs/restore.go
iancmcc: Done.

dfs/backup.go
iancmcc: The backup logging was too chatty, so I took stuff down. I can bring stuff back up, but this (along with others) seemed like a message you would only ever care about if something went wrong, hence debug.

dfs/backup.go
```diff
@@ -105,41 +97,68 @@ func (dfs *DistributedFilesystem) Backup(data BackupInfo, w io.Writer) error {
...
+}
+
+// dockerSavePipe streams the tar archive output by Docker to pipe's stdout
+func (dfs *DistributedFilesystem) dockerSavePipe(images ...string) pipe.Pipe {
+	return pipe.TaskFunc(func(s *pipe.State) error {
+		return dfs.docker.SaveImages(images, s.Stdout)
+	})
+}
+
+func (dfs *DistributedFilesystem) snapshotSavePipe(vol volume.Volume, label string) pipe.Pipe {
```
smousa: missing comment

dfs/restore.go
```diff
@@ -109,6 +85,188 @@ func (dfs *DistributedFilesystem) Restore(r io.Reader) (*BackupInfo, error) {
...
+	wg.Wait()
+	// Return any errors, if produced
+	if len(errs) > 0 {
+		return errs[0]
+	}
+	// Everything's ok, so finalize by pushing images to the registry if necessary
+	return dfs.loadRegistry(registryImages)
+}
+
+func (dfs *DistributedFilesystem) restoreVolume(tenant, label string, tarfile io.Reader) ([]string, error) {
```
smousa: missing comment

PR #2637, opened 2016-04-07
maintain list of pending GETs against api, and reject requests if ano…

tgjonestx:
…ther is still pending

### DIFF
--- a/web/ui/src/common/resourcesFactory.js
+++ b/web/ui/src/common/resourcesFactory.js
@@ -2,6 +2,10 @@
 (function(){
 
     const REQUEST_TIMEOUT = 30000;
+    const GET = "get";
+    const PUT = "put";
+    const DELETE = "delete";
+    const POST = "post";
 
     controlplane.factory("resourcesFactory", ["$http", "$location", "$notification", "DSCacheFactory", "$q", "$interval", "miscUtils",
     function($http, $location, $notification, DSCacheFactory, $q, $interval, utils) {
@@ -34,19 +38,19 @@
          */
         var methodConfigs = {
             assignIP: {
-                method: "PUT",
+                method: PUT,
                 url: (id, ip) => `/services/${id}/ip/${ip}`,
             },
             getPools: {
-                method: "GET",
+                method: GET,
                 url: "/pools"
             },
             getPoolIPs: {
-                method: "GET",
+                method: GET,
                 url: id => `/pools/${id}/ips`
             },
             addVHost: {
-                method: "PUT",
+                method: PUT,
                 url: (serviceID, endpointName, vhostName) => {
                     return `/services/${serviceID}/endpoint/${endpointName}/vhosts/${vhostName}`;
                 },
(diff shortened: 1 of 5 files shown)


web/ui/src/common/resourcesFactory.js
```diff
@@ -269,16 +287,56 @@
...
+
+                // NOTE: all of our code expects angular's wrapped
+                // promises which provide a success and error method
+                // TODO - remove the need for this
+                httpify(deferred);
+
+                if(pendingGETRequests[resourceName]){
+                    // theres already a pending request to
+                    // this endpoint, so fail!
+                    deferred.reject(`a request to ${resourceName} is pending`);
```
tgjonestx: Maybe I’m missing something, but it looks like it doesn’t exactly backoff (e.g. poll slower). Instead it just doesn’t poll at all, correct?   I guess that's OK - just wanted to confirm the strategy.

tgjonestx: Also, does the request-pending message from this reject bubble back up to the user?  Cuz the URL is not likely to be anything that makes sense to the user.  However, at a minimum we should log to the console I think

web/ui/src/common/resourcesFactory.js
```diff
@@ -269,16 +287,56 @@
...
+                if(resourceName.indexOf("?")){
+                    resourceName = resourceName.split("?")[0];
+                }
+
+                // NOTE: all of our code expects angular's wrapped
+                // promises which provide a success and error method
+                // TODO - remove the need for this
+                httpify(deferred);
+
+                if(pendingGETRequests[resourceName]){
```
tgjonestx: Seems like this "if" statement should include a `&& config.method === "get"` so that if this callback is invoked for other types of methods, we don't reject them just because a GET is a little slow.

web/ui/src/common/resourcesFactory.js
```diff
@@ -269,16 +287,56 @@
...
+                .success(function(data, status){
+                    deferred.resolve(data);
+                })
+                .error(function(data, status) {
+                    // TODO - include data as well?
+                    deferred.reject(status);
+                    redirectIfUnauthorized(status);
+                })
+                .finally(function() {
+                    pendingGETRequests[resourceName] = null;
```
tgjonestx: Ditto the previous comment about checking the method. We don't want a POST or DELETE to remove a cached promise for a GET that's still pending

tgjonestx:
I need to think about the cache of pendingGet promises vs other kinds of requests against the same URL.  I get that if a GET is taking a long time, maybe shouldn't try other types of requests on the same resource, but what if the inverse is true.  If a POST or a PUT is taking a really long time, should we continue polling the same endpoint with GETs? 

Does it make more sense to throttle all kinds of request to a given URL?

web/ui/src/common/resourcesFactory.js
rDr4g0n: Fixed!

web/ui/src/common/resourcesFactory.js
rDr4g0n: Fixed!

rDr4g0n:
future improvements can address other concerns:

1) backoff in the case of failures
2) notify the user when failures are stacking up
3) manage more than just GETs

PR #2646, opened 2016-04-12
CC-2125: Better pipe handling in the case where a snapshot volume being restored already exists

iancmcc:
When we encountered a snapshot to be restored, we would create a tar substream and start restoring from it. If that restore process detected that the snapshot already existed, it would return, saying it was complete, but the incoming files would still be written to the tar, which would block forever because nothing was reading from it anymore. 

This checks first to avoid restoring snapshots that already exist at a higher level.


smousa:
need unit tests to verify the thing you fixed

PR #2648, opened 2016-04-12
CC-2127: Better pipe handling in the case where a snapshot volume being restored already exists

iancmcc:
Backport of the fix for CC-2125 in #2646


smousa:
need unit tests to verify the thing you tested

PR #2661, opened 2016-04-19
Cleanup error messages for RPC registration failures

tgjonestx:
Fixes [CC-1057](https://jira.zenoss.com/browse/CC-1057?filter=-1)


cli/api/daemon.go
```diff
@@ -709,7 +709,7 @@ func (d *daemon) startAgent() error {
 		// register the API
 		glog.V(0).Infoln("registering ControlPlaneAgent service")
 		if err = d.rpcServer.RegisterName("ControlPlaneAgent", hostAgent); err != nil {
-			glog.Fatalf("could not register ControlPlaneAgent RPC server: %v", err)
+			glog.Fatalf("could not register RPC server named ControlPlaneAgent: %v", err)
```
smousa: We shouldn't be using the name "ControlPlane" if we can avoid it.  Please change to control center

PR #2662, opened 2016-04-20
Adds more verbose help function to service run

ed-:
"serviced service run <svc> help" displays commands available to that service.

Additionally, if a Command has a Description defined, the help message will
display that in the "Available commands" listing.

Closes CC-1102


ed-:
Still lacking new tests, don't merge yet.

PR #2666, opened 2016-04-21
Refactor tests in cli/api and commons/... to use gocheck

tgjonestx:
Partial implementation of Tech Debt Story [US14588](https://rally1.rallydev.com/#/46700477353d/detail/userstory/52148559712/tasks?igRequest=false&tpsPS=30&tpsSI=0)

Update all unit-tests in cli/api and commons to use gochk per the Zenoss GO testing [standard](https://docs.google.com/document/d/1mxkUeqabuz3LRuwwkR41DEM2acCa37bRgmWo2k4cV90/edit#).


commons/imageid_test.go
```diff
@@ -331,53 +326,37 @@ func TestValidateInvalid(t *testing.T) {
...
+		c.Assert(err, IsNil)
 
 		iid2, err := ParseImageID(tt.id2)
-		if err != nil {
-			t.Fatalf("error parsing %s on %d: %s", tt.id2, i, err)
-		}
-
-		expected := "match"
-		if !tt.expected {
-			expected = "mismatch"
```
wwwtyro: Do we not need to worry about this mismatch case anymore?

tgjonestx: I think the asserts I added at lines 354-355 cover the mismatch cases.  In the previous impl, the stack var `expected` was simply used in the test-failure messages.

PR #2674, opened 2016-04-27
CC-2175: update hosts if they are in a different pool during restore from backup

smousa:


smousa:
Does it make sense to leave the host alone if it exists in another pool or to restore it from backup?  Restoring the host from backup requires the host to be restarted.

PR #2690, opened 2016-05-03
Feature/cc 1908

smousa:
* Added the go version to serviced 
* Removed giturl because it was redundant and was never set anyway in any of our releases


tgjonestx:
Please update acceptance/mockAgent/main.go as well.

PR #2691, opened 2016-05-03
updated license file for CC-1.2

smousa:


smousa:
Per William, closing pr

PR #2698, opened 2016-05-05
Feature/us 14893 porthttp

rphillipsz:


web/ui/src/Services/AppsController.js
```diff
@@ -95,32 +95,26 @@
...
-                    endpoint.PortList.forEach(port => endPoints.push(port));
+                    endpoint.PortList.forEach(port => {
+                        port.URL = utils.publicEndpointURL($scope, port);
+                        endPoints.push(port);
+                    });
                 }
             });
 
-            endPoints.sort();
+            endPoints.sort(function(pep1, pep2) { return pep1.URL.localeCompare(pep2.URL); });
```
rDr4g0n: not strictly necessary, but es6 arrow functions are more succint for this exactly usage:
    endPoints.sort((pep1, pep2) =? pep1.URL.localeCompare(pep2.URL));

rphillipsz: I'll get that changed after lunch, no problem.

web/ui/src/Services/AppsController.js
```diff
@@ -95,32 +95,26 @@
 
             service.model.Endpoints.forEach(endpoint => {
                 if(endpoint.VHostList){
-                    endpoint.VHostList.forEach(vHost => endPoints.push(vHost));
+                    endpoint.VHostList.forEach(vHost => {
+                        vHost.URL = utils.publicEndpointURL($scope, vHost);
```
rDr4g0n: anything past the `model` property of a service (or instance, host, etc) should be immutable. The view is responsible for translating this data into something it can display to users, not the controller.

web/ui/src/common/miscUtils.js
```diff
@@ -264,6 +264,35 @@
 
                 }
                 return null;
+            },
+            
+            // Forms the name displayed for a public endpoint.
+            publicEndpointURL: function($scope, publicEndpoint){
```
rDr4g0n: passing `$scope` into a util is leaky. In this case, all that is needed is `$scope.defaultHostAlias`, so pass that in instead.

web/ui/src/Services/view-apps.html
```diff
@@ -38,10 +38,10 @@ <h3 class="pull-left" translate>applications</h3>
 
           <td data-title="'label_public_endpoints'|translate">
               <span ng-if="!service.deploying && !service.isIsvc()">
-                  <span ng-repeat="vhost in aggregateVHosts(service) track by $index">
-                      <health-status ng-if="service.desiredState !== 1 || !vhost.Enabled">{{publicEndpointURL(vhost)}}</health-status>
-                      <a ng-if="service.desiredState === 1 && vhost.Enabled" href="{{publicEndpointURL(vhost)}}" target="_blank" class="link">{{publicEndpointURL(vhost)}}</a>
-                  </span>
```
rDr4g0n: keep this pattern of letting the view do the work of translating the model into something that is optimal for the user.

PR #2711, opened 2016-05-10
Fixes CC-2186 "zk: api error, Could not update service state, services are not starting"

johnlhamilton:
CC-2186
This fixes a problem that was introduced when we updated go-dockerclient, which returns different error messages for "image not found" errors.


johnlhamilton:
retest this

PR #2713, opened 2016-05-16
CC-2158: make lvm monitoring configurable during serviced startup

smousa:


volume/devicemapper/devicemapper.go
```diff
@@ -494,7 +497,25 @@ func (d *DeviceMapperDriver) ensureInitialized() error {
 		return err
 	}
 	if d.DeviceSet == nil {
-		deviceSet, err := devmapper.NewDeviceSet(poolPath, true, d.options, nil, nil)
+		var (
+			thinPoolDev         string
+			enableLVMMonitoring string
```
tgjonestx: Shouldn't the default for this string be `"n"`?  We don't want customers to make a configuration change to opt-in to the no-monitoring feature.

volume/devicemapper/devicemapper.go
```diff
@@ -494,7 +497,25 @@ func (d *DeviceMapperDriver) ensureInitialized() error {
...
+			if strings.HasPrefix(option, "dm.") {
+				dmoptions = append(dmoptions, option)
+				if strings.HasPrefix(option, "dm.thinpooldev=") {
+					thinPoolDev = strings.TrimPrefix(option, "dm.thinpooldev=")
+				}
+			} else if strings.HasPrefix(option, "enablelvmmonitoring=") {
+				enableLVMMonitoring = strings.TrimPrefix(option, "enablelvmmonitoring=")
+			} else {
+				glog.Errorf("Unable to parse option %s", option)
+				return ErrInvalidOption
```
iancmcc: Maybe don't choke, just log it and move on.

volume/devicemapper/devicemapper.go
```diff
@@ -504,6 +525,36 @@ func (d *DeviceMapperDriver) ensureInitialized() error {
...
+				cmd = exec.Command("lvchange", "--monitor", "y", thinPoolDev)
+				output, err := cmd.CombinedOutput()
+				if err != nil {
+					glog.Errorf("Could not run command %v: %s (%s)", cmd, string(output), err)
+					return err
+				}
+			} else {
+				glog.Warningf("Ignoring option 'enablelvmmonitoring'; no thin pool device specified")
+			}
+		case "n", "":
```
iancmcc: Can we just use `default:` here?

volume/devicemapper/devicemapper.go
tgjonestx: NVM - I missed the extra `""` in the case statement below

volume/devicemapper/devicemapper.go
smousa: @jplouis prefers that it chokes in case someone fat-fingers something

PR #2714, opened 2016-05-17
fix baseFactory not passing errors along promise chain, make hosts pa…

rDr4g0n:
…ge wait on success


web/ui/src/common/baseFactory.js
```diff
@@ -107,6 +107,8 @@
                 })
                 .error((data, status) => {
                     log.warn("Unable to update factory", data);
+                    // TODO - include status?
+                    deferred.reject(data);
```
rDr4g0n: in case of failure, the promise wasnt being rejected

PR #2724, opened 2016-05-17
Add NODEJS path for the build container

rphillipsz:


tgjonestx:
Please add the same kind of change in 3 other places:
* in this same makefile on/around line 231 for the `docker_build` target
* in the 2 invocations of `zenoss/serviced-build` in `web/ui/makefile`

The later will be invoked by anyone who just runs `make` on their local without having node installed

PR #2729, opened 2016-05-18
CC-2129 Upgrade to Angular 1.5.5

carlcamera:
Changes to $cookie require changes to html templates and JS files.


carlcamera:
retest this.

PR #2733, opened 2016-05-19
CC-2219: resize volume if snapshot import is bigger

smousa:
* If the tenant is created during a restore and the snapshot import for the restore fails, we will remove the tenant volume that is created.
* Save the device size in the devicemapper snapshot export.
* If the device size is saved in the exported snapshot and it is bigger than its parent volume, resize the parent volume.
* Added tests for new test cases


volume/devicemapper/devicemapper_test.go
```diff
@@ -120,6 +124,83 @@ func (s *DeviceMapperSuite) TestDeviceMapperExportImport(c *C) {
...
+	c.Assert(err, IsNil)
+	defer drv1.Remove("basetest")
+	v1, ok := vol1.(*DeviceMapperVolume)
+	c.Assert(ok, Equals, true)
+	size, err := v1.SizeOf()
+	c.Assert(err, IsNil)
+	c.Assert(size, Equals, uint64(basesize1))
+	f1, err := ioutil.TempFile(vol1.Path(), "dump-")
+	c.Assert(err, IsNil)
+	err = f1.Close()
```
iancmcc: Right here you need to write 11G of actual data into that volume.

volume/devicemapper/devicemapper.go
```diff
@@ -1160,6 +1197,22 @@ func (v *DeviceMapperVolume) Import(label string, reader io.Reader) error {
 	if err := v.driver.DeviceSet.MountDevice(device, filepath.Join(mountpoint, label), fmt.Sprintf("%s_import", label)); err != nil {
 		return err
 	}
+	// The size of the staging volume is determined by the size of the
+	// tenant volume; we need it to be based off of the dm.basesize value if it
+	// is smaller.
+	curSize, err := v.SizeOf()
```
iancmcc: this is TECHNICALLY not valid. The size of the staging volume is determined by the size of the BASE volume; the tenant volume could easily have been resized (and often will be).

smousa:
devicemapper code changed in docker-1.11 code.
Will need a different solution to commit to 1.1.x from 1.2

volume/devicemapper/devicemapper.go
```diff
@@ -104,44 +108,123 @@ func (d *DeviceMapperDriver) ListTenants() (result []string) {
...
+}
+
+// deviceSize returns the size of a device.
+func (d *DeviceMapperDriver) deviceSize(deviceID string) (uint64, error) {
+	return getDeviceSize(d.deviceRoot(deviceID))
+}
+
+// deviceRoot returns the /dev/mapper path of a device
+func (d *DeviceMapperDriver) deviceRoot(deviceID string) string {
+	return fmt.Sprintf("/dev/mapper/%s", d.deviceName(deviceID))
```
iancmcc: I believe this needs to be fmt.Sprintf("/dev/mapper/%s", strings.TrimPrefix(d.deviceName(deviceID), "/dev/mapper/"))

smousa: why?

iancmcc: So one does not have to keep track of the type of device id being passed
through, as there are three. This hit us elsewhere, remember. It's a
basically free way to prevent future mistakes.

smousa: do you mean on raw device ID? not d.deviceName(deviceID)

smousa: Also, would I also need to trim the device prefix, right.  I guess I don't understand all the expected values of "device ID".

iancmcc: I mean that where the code is blindly prepending the string /dev/mapper/ to whatever is passed in, it should guarantee that it isn't doubling up, that's all. Idiomatic pattern, like for absolute paths you always do path.rstrip('/') + '/my/other/path' to prevent double slashes.

iancmcc: All my example did was take the code that was already there and wrap the format argument with strings.TrimPrefix(___, "/dev/mapper/").

smousa: are any of the places where I make these calls have the potential of  having /dev/mapper prefixed?

volume/devicemapper/devicemapper.go
```diff
@@ -1135,86 +1236,144 @@ func (d *DeviceMapperDriver) GetTenantStorageStats() ([]volume.TenantStorageStat
...
+
+	return nil
+}
+
+func (v *DeviceMapperVolume) SizeOf() (uint64, error) {
+	return v.driver.deviceSize(v.deviceID())
 }
 
-func (v *DeviceMapperVolume) volumeDevice() string {
+func (v *DeviceMapperVolume) deviceID() string {
```
iancmcc: Why'd you rename this? deviceID is somewhat more ambiguous, since there are many devices associated with a volume (the volume device itself, and the snapshot devices). Maybe tenantDeviceID? or currentDeviceID? or whatever

smousa: Consistency...I guess it is better if I renamed to deviceHash (we are talking about this offline)

smousa: Also I don't know if the volumeDevice refers to the hash, the name, or the path.

volume/devicemapper/devicemapper.go
```diff
@@ -935,34 +1031,39 @@ func (v *DeviceMapperVolume) Export(label, parent string, writer io.Writer) erro
...
-		v.driver.DeviceSet.Lock()
-		if err := v.driver.deactivateDevice(device); err != nil {
-			glog.V(2).Infof("Error deactivating device (%s): %s", device, err)
-		}
-		v.driver.DeviceSet.Unlock()
-		// Be a good citizen and clean up the mountpoint
-		os.RemoveAll(mountpoint)
-	}()
-
+	defer v.driver.unmountDevice(device, mountpoint)
```
iancmcc: The deactivation and stuff is not part of unmounting and probably shouldn't be contained therein. "Unmount" does not always include "deregister the device from the kernel table" and "delete the mount point," it just does in these two cases. Maybe a separate function?

smousa: Are you asking me to rename this function?

iancmcc: Also, I think we should keep the comment and UnmountDevice here, because it is key that it comes immediately after the MountDevice (and will in fact break things if it doesn't). The rest could certainly be in a function called, say, cleanup(device, mountpoint)

volume/devicemapper/devicemapper.go
```diff
@@ -1135,86 +1236,144 @@ func (d *DeviceMapperDriver) GetTenantStorageStats() ([]volume.TenantStorageStat
...
+		glog.Errorf("Could not stage the device %s at mountpoint %s for snapshot %s: %s", deviceID, mountpoint, label, err)
 		return err
 	}
-	glog.V(2).Infof("Mounting imported snapshot device %s", device)
-	if err := v.driver.DeviceSet.MountDevice(device, filepath.Join(mountpoint, label), fmt.Sprintf("%s_import", label)); err != nil {
+	defer v.driver.unmountDevice(deviceID, mountpoint)
+	glog.V(2).Infof("Mounted staging device %s at mountpoint %s for snapshot %s", deviceID, mountpoint, label)
+
+	// Set up the metadata directory
+	// TODO: clean up the metadata on failure
```
iancmcc: Do you want to do this?

volume/devicemapper/devicemapper.go
```diff
@@ -1135,86 +1236,144 @@ func (d *DeviceMapperDriver) GetTenantStorageStats() ([]volume.TenantStorageStat
...
+		metaDir    = label + "-metadata" // Metadata
+	)
+	driverType := ""
 	tarfile := tar.NewReader(reader)
 	for {
 		header, err := tarfile.Next()
 		if err == io.EOF {
-			break
+			if driverType == "" {
+				return errors.New("incompatible snapshot")
```
iancmcc: Bring up to a global error. SORRY

volume/devicemapper/devicemapper.go
```diff
@@ -1135,86 +1236,144 @@ func (d *DeviceMapperDriver) GetTenantStorageStats() ([]volume.TenantStorageStat
...
+			} else {
```
iancmcc: No need for else here

PR #2734, opened 2016-05-19
Feature/snapshot storage

bwf93:


dfs/snapshot.go
```diff
@@ -95,3 +108,25 @@ func (dfs *DistributedFilesystem) Snapshot(data SnapshotInfo) (string, error) {
 func generateSnapshotLabel() string {
 	return time.Now().UTC().Format("20060102-150405.000")
 }
+
+// checks to see if there is enough free space on volume to perform a snapshot
+func ensureFreeSpace(vol volume.Volume, dfs *DistributedFilesystem, snapshotSpacePercent int) (bool, error) {
+	status := volume.GetStatus()
+	statusMap := status.DeviceMapperStatusMap[dfs.disk.Root()]
```
iancmcc: You should just return true here if it isn't devicemapper.

dfs/snapshot.go
```diff
@@ -95,3 +108,25 @@ func (dfs *DistributedFilesystem) Snapshot(data SnapshotInfo) (string, error) {
...
+	var amountNeeded float64
+	foundTenant := false
+	for i := 0; i < len(statusMap.Tenants); i++ {
```
iancmcc: technically you can do 
```
for _, tenant := range statusMap.Tenants {
}
```

PR #2739, opened 2016-05-20
Feature/port acceptance

achow-zenoss:


acceptance/ui/features/data/default/services.json
```diff
@@ -12,9 +12,31 @@
...
     "childServiceWithVHost": {
       "name": "s2",
-      "vhostName": "websvc",
+      "vhostName" : "websvc",
       "publicPortNumber": 1234,
+      "host"      : "%{local_ip}",
+
+      "endpoint_https"  : "websvc",
+      "protocol_https"  : "HTTPS",
+      "protocol_https_o"  : "HTTPS",
```
tgjonestx: What does the `_o` suffix mean?

acceptance/ui/features/services.feature
```diff
@@ -29,4 +29,183 @@ Feature: Service Management
...
+      And I click "Cancel"
+
+  @services_endpoint
+  Scenario: Public Endpoint dialog - VHost mode visible fields
+    When I view the details of "table://services/topService/name" in the "Applications" table
+      And I view the details of "table://services/childServiceWithVHost/name" in the "Services" table
+      And I click "Add Public Endpoint"
+      And I should see the Add Public Endpoint dialog
+      And I choose VHost type
+    Then I should see the Add Public Endpoint dialog
```
tgjonestx: Is this step really necessary?

acceptance/ui/features/services.feature
```diff
@@ -29,4 +29,183 @@ Feature: Service Management
...
+      And I click "Add Public Endpoint"
+    Then I should see the Add Public Endpoint dialog
+      And I should see "Type:"
+      And I should see "Port"
+      And I should see "VHost"
+      And I should see "Service - Endpoint:"
+      And I should see "Host:"
+      And I should see "Port:"
+      And I should see "Protocol:"
+      And I click "Cancel"
```
tgjonestx: Do you have to end all of these will Cancel?

The concern is keeping the tests as fast as possible. I thought each scenario starts over.

acceptance/ui/features/services.feature
```diff
@@ -29,4 +29,183 @@ Feature: Service Management
       And I should see "Public Endpoints"
       And VHost "table://services/childServiceWithVHost/vhostName" should exist
       And Public Port "table://services/childServiceWithVHost/publicPortNumber" should exist
-      And I should see "table://services/childServiceWithVHost/configFile"
\ No newline at end of file
+      And I should see "table://services/childServiceWithVHost/configFile"
+
+  @services_endpoint
+  Scenario: Public Endpoint dialog - check all labels
```
tgjonestx: Please add a test case that verifies that dialog is populated with all of the correct values for each field

acceptance/ui/features/steps/services_steps.rb
```diff
@@ -1,11 +1,97 @@
...
     expect(checkPublicPortExists(port))
 end
 
+Then (/^I should see the Add Public Endpoint dialog$/) do
+    initServicePageObj()
+    expect(@service_page.addPublicEndpoint_dialog.visible?).to be true
+end
+
+Then (/^I choose Port type$/) do
+    @service_page.find("div.btn-group", "//input[@value='port']" ).click
```
tgjonestx: Can this be redefined so that we can do something like `@service_page.portTypeButton.click`?

As much as possible, I'm trying to keep the CSS/xpath expressions in the page object model definitions.

acceptance/ui/features/steps/services_steps.rb
```diff
@@ -1,11 +1,97 @@
...
+end
+
+Then (/^I choose VHost type$/) do
+    @service_page.find("div.btn-group", "//input[@value='vhost']" ).click
```
tgjonestx: ditto last comment

tgjonestx:
Please rebase and retest locally because Carl recently updated all of our 3rd-party JS libs

acceptance/ui/features/data/default/services.json
achow-zenoss: The input and output are of different strings.. 
Probably should be renamed to _https_in and _https_out..

acceptance/ui/features/services.feature
achow-zenoss: Not really. Not sure if we need to test visible labels. We can remove this step.

acceptance/ui/features/services.feature
achow-zenoss: You are right. I'll remove all cancel steps.

acceptance/ui/features/steps/services_steps.rb
achow-zenoss: OK

acceptance/ui/features/steps/services_steps.rb
achow-zenoss: OK

acceptance/ui/features/services.feature
achow-zenoss: OK.. I'll change the previous check field visibility to also check the populated values

acceptance/ui/features/data/default/services.json
achow-zenoss: To make it more readable, I've changed _*_o to _*_display..

acceptance/ui/features/steps/services_steps.rb
achow-zenoss: This change (modeling the input button on the dialog) does not seem to work straight off the page model. We'll leave this change later.

acceptance/ui/features/services.feature
tgjonestx: I didn't see these changes to address this comment in 7c4fe6d528ba5f58b793b5a9a11a62130b94c2a4.  Did I miss something?

acceptance/ui/features/services.feature
```diff
@@ -29,4 +29,181 @@ Feature: Service Management
...
+      And I click "Add Public Endpoint"
+      And I should see the Add Public Endpoint dialog
+      And I select Service "table://services/childServiceWithVHost/name" - "table://services/childServiceWithVHost/endpoint_https"
+      And I fill in Host "table://services/childServiceWithVHost/host"
+      And I fill in Port "table://services/childServiceWithVHost/port_https"
+      And I select Protocol "table://services/childServiceWithVHost/protocol_https"
+      And I click "Add and Restart Service"
+    Then I should see "table://services/childServiceWithVHost/name"
+      And I should see "Public Endpoints"
+      And Endpoint details should be service:"table://services/childServiceWithVHost/name" endpoint:"table://services/childServiceWithVHost/endpoint_https" type:"port" protocol:"table://services/childServiceWithVHost/protocol_https_display" host:"table://services/childServiceWithVHost/host" port:"table://services/childServiceWithVHost/port_https"
```
tgjonestx: I thought I made this comment in the first round, but I guess I didn't click Save or something. Anyway, please split this step into separate steps so that only one value value is verified in each step. For instance something like:
```
And Endpoint details should have service: "table://services/childServiceWithVHost/name"
And Endpoint details should have endpoint: "table://services/childServiceWithVHost/endpoint_https"
...
```
This makes it easier to combine steps in future tests, and it also makes it easier to see what is wrong when a single value is incorrecct.

tgjonestx:
@achow-zenoss  - also, can you split the new test cases into an entirely new `services_endpoints.feature` file?  That way you only have to put the `@services_endpoint` on the feature, instead of on every Scenario.

acceptance/ui/features/services.feature
achow-zenoss: OK

tgjonestx:
Regarding the 16m test time, I looked at this some more and I think we can live with the time increase for now.  But I think we do need to look at the random failures with the dialog stuff because that will create random failures in the build pipeline.  Let's discuss that one

PR #2744, opened 2016-05-25
Merge pull request #2562 from control-center/feature/CC-1179.4

smousa:
CC-1962: Update the client library and set up RWLock on client connection


tgjonestx:
Doing a little code archaeology on develop to see if there were any subsequent changes that might have been made after #2562 which were a result of something that might have been broken/changed because of #2562. I found at least one:
#2566 was created for a ticket created the day after #2562 was merged. Do we also need to backport #2566?

smousa:
Yes, thanks for checking.  Will update PR

PR #2745, opened 2016-05-25
CC-2113 CC Master timeout if ZK quorum can't be established

ed-:
Changed the zookeeper healthcheck to use 'ruok' and 'stat' commands
instead of go-zookeeper connection. This way we can check separately
that the zookeeper agent is up and that quorum is established. Should
zookeeper fail to establish quorum, we can log accordingly.

Iservices that fail their initial healthcheck should now dump the last
10000 lines of their container before shutdown.

### DIFF
--- a/isvcs/container.go
+++ b/isvcs/container.go
@@ -432,7 +432,21 @@ func (svc *IService) start() (<-chan int, error) {
 	select {
 	case err := <-svc.startupHealthcheck():
 		if err != nil {
-			glog.Errorf("Healthcheck for %s failed: %s", svc.Name, err)
+			err = fmt.Errorf("Startup healthcheck for %s failed after %3.0f seconds: %s", svc.Name, svc.StartupTimeout.Seconds(), err)
+			glog.Error(err)
+			// Dump last 10000 lines of container if possible.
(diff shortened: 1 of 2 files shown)


isvcs/container.go
```diff
@@ -432,7 +432,21 @@ func (svc *IService) start() (<-chan int, error) {
...
 	case err := <-svc.startupHealthcheck():
 		if err != nil {
-			glog.Errorf("Healthcheck for %s failed: %s", svc.Name, err)
+			err = fmt.Errorf("Startup healthcheck for %s failed after %3.0f seconds: %s", svc.Name, svc.StartupTimeout.Seconds(), err)
+			glog.Error(err)
+			// Dump last 10000 lines of container if possible.
+			if output, err := exec.Command("docker", "logs", "--tail", "10000", ctr.ID).CombinedOutput(); err != nil {
+				glog.Errorf("Could not get logs for container %s", ctr.ID)
+			} else {
+				prefix := fmt.Sprintf("ctr-%s: ", ctr.ID[0:5])
```
tgjonestx: I realize this was probably copied from existing code, but for isvcs, unlike regular services, we don't have a record of the container ID for each service, and we're guaranteed to have unique names for each service. So I think it should be better to replace `ctr-<nnnnn>` prefix with something like `ctr-<svc.Name>`

PR #2751, opened 2016-05-26
Add Host info to log export index.txt file

tgjonestx:
Fixes https://jira.zenoss.com/browse/CC-2234.

Has lots of refactoring to make some of the internals more testable. The commits that don't say 'refactor' are the ones without actual functional changes.  

Below is a sample of what the new index.txt file looks like.  Note that the columns do not line up vertically because the titles and the column values are tab-separated.
```
INDEX OF LOG FILES
File	Line Count	Host Name	Host ID	Container ID	Original Filename
000.log	64894	"gjones-dev"	"570a4c80"	"43c3716a041f"	"/opt/zenoss/log/Z2.log"
001.log	64898	"gjones-dev"	"570a4c80"	"1c12ace20e0e"	"/opt/zenoss/log/Z2.log"
002.log	120	"gjones-dev"	"570a4c80"	"cf060935eae1"	"/opt/zenoss/log/zenjobs.log"
```


smousa:
The main beef I have here is the code you have where you are using zenoss/elastigo is going to have to be refactored yet again as soon as we get the latest elastigo

cli/api/logs.go
```diff
@@ -49,10 +83,103 @@ type ExportLogsConfig struct {
 //
 // TODO: This code is racy - creating then erasing the output file does not
```
smousa: Should we still concern ourselves with this race condition?

cli/api/logs.go
```diff
@@ -49,10 +83,103 @@ type ExportLogsConfig struct {
...
+		return fmt.Errorf("no logstash indexes exist for the given date range %s - %s", exporter.FromDate, exporter.ToDate)
+	}
+
+	glog.Infof("Starting part 2 of 3: sort %d output files", len(exporter.outputFiles))
+
+	indexData := []string{}
+	for i, outputFile := range exporter.outputFiles {
+		filename := filepath.Join(exporter.tempdir, fmt.Sprintf("%03d.log", i))
+		tmpfilename := filepath.Join(exporter.tempdir, fmt.Sprintf("%03d.log.tmp", i))
+		cmd := exec.Command("sort", filename, "-uo", tmpfilename)
```
smousa: there are Go utilities for most, if not all, of these calls.  Why are you shelling?

cli/api/logs.go
tgjonestx: Since the output file name is based on a timestamp with resolution of down to the second (RFC3339), the race condition is one where two or more CLI commands that are initiated within the same second on the same host from the same directory.  

Possible solutions:

1. Add some unique salt to the end of the file name (e.g. a random number) Cons - the file names are already >40 chars long.
2. Lower the probability of collision even further by using a timestamp based on milliseconds. Technically, still racy
3. Narrow the time window for potential race further by keeping tgzfile open for the duration of the operation. Move the Close/Remove calls from here https://github.com/control-center/serviced/blob/8d97521b04dbc3e122bf9b993538f3326a044d10/cli/api/logs.go#L204 down to here https://github.com/control-center/serviced/blob/8d97521b04dbc3e122bf9b993538f3326a044d10/cli/api/logs.go#L160. That 
4. Do #2 and #3

cli/api/logs.go
tgjonestx: I'm just reusing the cards I was dealt.  
What are the GO utilities for sort file and copy file?
I did find `os.Rename` to replace `mv`, but I didn't find one-line equivalents for `cp` and `sort`

tgjonestx:
re: elastigo - I looked at other uses of elastigo in CC in attempt to define a universal interface to elastigo for all of the ways we call it, but there wasn't much overlap between the typical calls to ES-serviced vs the calls in cli/cmd/logs.go to ES-logstash, and the calls were throughout much of the domain and datastore packages. So replacing all references to elastigo with one interface seemed like much more work than called for this by ticket.  Instead, I opted for a thin shim to at least make cli/cmd/logs.go unit-testable. Since most of the work in creating the shim was just copy/paste, I don't see how having the shim or not changes the overall level of effort to upgrade elastigo (or replace it).

PR #2755, opened 2016-06-01
CC-2265: fixed race condition for error caused while exporting a snapshot during backup

smousa:


tgjonestx:
Can you change the V2 Info message at line 42 of backup.go to an Errorf?
https://github.com/control-center/serviced/blob/fab290a11fdb1207e7c3408d2842f5a4065ee66d/dfs/backup.go#L42

tgjonestx:
Besides line 42, there are only a couple of other `V(2).Info` messages - why not remove the 'V(2)' from those? I don't think it hurts anything and might give us a little more clarity on future cases.

tgjonestx:
Is there any way to unit-test some of this? Maybe at least `savePipe()` and/or `rewriteTar()`?

tgjonestx:
The tests look good

PR #2756, opened 2016-06-01
ZEN-22907 Ability to alert when Control Center service is overcommitted

ssOlexBaiko:


smousa:
This is too application specific to be in the serviced repo

jpeacock-zenoss:
I thought the intent for ZEN-22907 was for RM to perform the monitoring and alerting.  How does this change leverage RM's monitoring and alerting functionality?

jpeacock-zenoss:
The ControlCenter zenpack is probably the place where the fix for ZEN-22907 should be made.

PR #2769, opened 2016-06-07
expose PullImage as an rpc call on the worker

smousa:
Check if a pull is needed before trying to start a shell for the service.

### DIFF
--- a/cli/api/daemon.go
+++ b/cli/api/daemon.go
@@ -671,7 +671,8 @@ func (d *daemon) startAgent() error {
 	// TODO: Integrate this server into the rpc server, or something.
 	// Currently its only use is for command execution.
 	go func() {
-		sio := shell.NewProcessExecutorServer(options.Endpoint, options.DockerRegistry, options.ControllerBinary, options.UIPort)
+		agentEndpoint := "localhost:" + options.RPCPort
+		sio := shell.NewProcessExecutorServer(options.Endpoint, agentEndpoint, options.DockerRegistry, options.ControllerBinary)
 		http.ListenAndServe(":50000", sio)
 	}()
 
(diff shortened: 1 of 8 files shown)


cli/api/daemon.go
```diff
@@ -671,7 +671,8 @@ func (d *daemon) startAgent() error {
 	// TODO: Integrate this server into the rpc server, or something.
 	// Currently its only use is for command execution.
 	go func() {
-		sio := shell.NewProcessExecutorServer(options.Endpoint, options.DockerRegistry, options.ControllerBinary, options.UIPort)
+		agentEndpoint := "localhost:" + options.RPCPort
```
tgjonestx: I'm confused about when we are supposed to use localhost vs 127.0.0.1, but looks like localhost is used more than 127.0.0.1 in our code.

shell/server.go
```diff
@@ -438,14 +430,9 @@ func StartDocker(cfg *ProcessConfig, dockerRegistry, port, controller string) (*
...
-	systemUser := user.User{}
-	err = cp.GetSystemUser(unused, &systemUser)
-	if err != nil {
-		glog.Errorf("Unable to get system user account for client %s", err)
-	}
-	argv = append(argv, "-e", fmt.Sprintf("CONTROLPLANE_SYSTEM_USER=%s ", systemUser.Name))
-	argv = append(argv, "-e", fmt.Sprintf("CONTROLPLANE_SYSTEM_PASSWORD=%s ", systemUser.Password))
+	// FIXME: wtf?!?!?!
+	argv = append(argv, "-e", fmt.Sprintf("CONTROLPLANE_SYSTEM_USER=system_user "))
+	argv = append(argv, "-e", fmt.Sprintf("CONTROLPLANE_SYSTEM_PASSWORD= "))
```
tgjonestx: I think this may break the CC integration in RM. 
See https://github.com/zenoss/zenoss-prodbin/blob/develop/Products/ZenUtils/controlplane/__init__.py

PR #2770, opened 2016-06-08
CC-2175: do not try to restore hosts from backup

smousa:


tgjonestx:
Is there a relatively easy way to test that if the BackupInfo in a restored tar contains Host info that it gets ignored?

PR #2771, opened 2016-06-08
Feature/us 14900

rphillipsz:


cli/api/interfaces.go
```diff
@@ -119,4 +120,7 @@ type API interface {
 
 	// Volumes
 	GetVolumeStatus() (*volume.Statuses, error)
+
+    // Public endpoints
+    AddPublicEndpointPort(serviceid, endpointName, portAddr string, usetls bool, protocol string, isEnabled bool) (*servicedefinition.Port, error)
```
tgjonestx: Looks like a tabs/spaces indenting irregularity here.

rphillipsz: Yeah. this is my editor.  I was noticing it also in other parts of the code.  I'll work on sorting that out.

cli/cmd/publicendpoint.go
```diff
@@ -0,0 +1,379 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"encoding/json"
+    "fmt"
+	"os"
+    "strconv"
```
tgjonestx: Again - more tab/space inconsistencies. The convention here for GO code is to indent with tab characters.

Also, you can try using 'gofmt' to format new files.

cli/cmd/publicendpoint.go
```diff
@@ -0,0 +1,379 @@
...
+        Application : endpoint,
+        EpType      : epType,
+        Protocol    : protocol,
+        Name        : name,
+        Enabled     : enabled,
+	}
+}
+
+// serviced service public-endpoints
+func (c *ServicedCli) cmdPublicEndpointsList(ctx *cli.Context) {
```
tgjonestx: Naming - the _ prefix is not a convention I've seen us use for the GO code. The fact that the method name is lower case already makes it package private.  I suggest changing `cmdPublicEndpointsList` to `cmdPublicEndpointsListAll` and then change `_cmdPublicEndpointsList` to just `cmdPublicEndpointsList`

facade/interface.go
```diff
@@ -48,6 +49,14 @@ type FacadeInterface interface {
 
 	ScheduleService(ctx datastore.Context, serviceID string, autoLaunch bool, desiredState service.DesiredState) (int, error)
 
+	StartService(ctx datastore.Context, request dao.ScheduleServiceRequest) (int, error)
```
tgjonestx: Rather than expand the surface area of Facade with functions that aren't used outside of Facade, I would move more of the code in the publicendpointresource into the facade - see note below

web/publicendpointresource.go
```diff
@@ -291,118 +291,35 @@ func restAddPort(w *rest.ResponseWriter, r *rest.Request, client *node.ControlCl
...
-		glog.Errorf("Unexpected error adding port to service (%s): %v", service.Name, err)
-		restServerError(w, err)
-		return
-	}
-
-	glog.V(2).Infof("Service (%s) updated", service.Name)
+		port.PortAddr, service.Name, port.UseTLS, port.Protocol)
 
 	// Restart the service if it is running
 	if service.DesiredState == int(svc.SVCRun) || service.DesiredState == int(svc.SVCRestart) {
```
tgjonestx: I would move this restart logic inside Facade.AddPort() so that the CLI add-port and the UI add-port have the same results.  If a use-case emerges for adding the port without restarting the service, we can add extra parameters to the AddPort() method to control that.

rphillipsz: I considered this, but decided against it.  The "serviced service edit"
doesn't restart the service if running, but the UI prompts the user that
this will happen when adding a new endpoint.  If you want both to do this,
I can certainly move it.

web/publicendpointresource.go
```diff
@@ -291,118 +291,35 @@ func restAddPort(w *rest.ResponseWriter, r *rest.Request, client *node.ControlCl
...
-	// Validate the port number
-	scrubbedPort := svc.ScrubPortString(request.PortName)
-	portParts := strings.Split(scrubbedPort, ":")
-	if len(portParts) <= 1 {
-		err := fmt.Errorf("Invalid port address. Port address be \":[PORT NUMBER]\" or \"[IP ADDRESS]:[PORT NUMBER]\"")
-		glog.Error(err)
-		restServerError(w, err)
-		return
-	}
-	port, err := strconv.Atoi(portParts[1])
```
tgjonestx: Lines 305-329 need to be moved into the new facade method to insure that the `portParts[1]` string is validated correctly.

web/publicendpointresource.go
```diff
@@ -291,118 +291,35 @@ func restAddPort(w *rest.ResponseWriter, r *rest.Request, client *node.ControlCl
...
-					return
-				}
-			}
-		}
-	}
-
-	// Check to make sure the port is available.  Don't allow adding a port if it's already being used.
-	err = checkPort("tcp", fmt.Sprintf("%s", scrubbedPort))
+	// Get the service for this service id.
+	service, err := facade.GetService(dataCtx, request.ServiceID)
```
tgjonestx: facade.AddPublicEndpointPort does this, so I think you can remove it from this web method.

web/publicendpointresource.go
rphillipsz: This isn't necessary, because the facade ensures the port can be bound to before adding it.  An invalid port address is rejected with appropriate messages from the facade's call to checkPort()

web/publicendpointresource.go
rphillipsz: The service was needed for restarting it further below.  See the commend on moving (or not moving) that block to the facade.

web/publicendpointresource.go
tgjonestx: Clarification - I didn't mean move the restart inside of facade.UpdateService. I meant have facade.AddPort() do the work of adding the port, and then call UpdateService() and then call this restart logic.

I'm a little confused about your comment re: UI prompts because the existing code in restAddPort() restarts the service without regard to any input from the user.

web/publicendpointresource.go
tgjonestx: What about the checks for reserved ports 5000 and 8080? I'm not sure why we added those, but this change looks like it will allows users to add an endpoint-by-port for those values.

rphillipsz: These were left over from an original mock up that Matos did.  He would probably know better why 5000 and 8080 were restricted.  I *think* this was an attempt to prevent them from adding an endpoint to a port that CC already uses for other things.  Since the code checks if the port is open already, that should take care of these (and others).. but more importantly a lot of those ports are only opened on 127.0.0.1, not all interfaces.  It should be valid to open port 5000 on a different interface.

web/publicendpointresource.go
rphillipsz: This is in errors like "glog.Errorf("Error adding port to service (%s): %v", service.Name, err)", which is displayed in a flare in the dialog.

tgjonestx:
See 26a2781 for a potential fix to one of the unit-tests

tgjonestx:
We should probably comment out support for `public-endpoints vhost *` entirely and create separate stories for those because they will entail their own refactoring to implement correctly.  Specifically, trying to avoid "bugs" like the situation below where the CLI suggests there's a feature, but it hasn't been implemented yet.

Alternatively, you could go ahead and implement "vhost list" in this PR, and we'll create separate stories to cover vhost-add, vhost-remove, etc.
```
$ serviced service public-endpoints list
Service            ServiceID                      Endpoint                Type       Protocol      Name          Enabled
├─Zenoss.core      e2e1ukqvxv4tde7qxlmvvnfo0      zproxy                  vhost      https         zenoss5       true
├─HMaster          9acvpv0qzbjlvuel07syoi8yl      hbase-masterinfo-1      vhost      https         hbase         false
├─RabbitMQ         8zh08eys603erv0pgeh566boz      rabbitmq_admin          vhost      https         rabbitmq      true
└─opentsdb         d1p00okas1799x6kz818hulha      opentsdb-reader         vhost      https         opentsdb      false

$ serviced service public-endpoints vhost list
No public endpoints found
```

rphillipsz:
Odd, l'll have to look at that when I get back.  It is actually implemented
and should have worked.

PR #2772, opened 2016-06-08
Merge pull request #2699 from control-center/feature/CC-2191

smousa:
CC-2191: update backup/restore state if running async


smousa:
The effort here may not be worth implementing in a 1.1.x release

PR #2774, opened 2016-06-09
CC-2284: Update version of go-zookeeper/zk

glujan:


smousa:
Please talk to me before merging

glujan:
We've changed approach to this change

PR #2785, opened 2016-06-15
wait for the command to exit when running dumpe2fs

smousa:


volume/devicemapper/stats.go
```diff
@@ -219,6 +219,8 @@ func getFilesystemStats(dev string) (uint64, uint64, error) {
 	if err = cmd.Start(); err != nil {
 		return 0, 0, err
 	}
+	defer cmd.Wait()
```
jplouis: Why not use cmd.Run instead of cmd.Start?

PR #2788, opened 2016-06-15
Keep serviced from messing with the base device

johnlhamilton:
This fixes CC-2294, which occurs when the thin pool is deleted out from under an existing application, causing serviced to use the devmapper base device for that application, ultimately leading to the deletion of the base device and failure of serviced to start.

We address this issue on 3 fronts:
1. Prevent serviced from deleting the base device
2. Prevent serviced from loading a volume that is using the base device
3. If somehow the base device is missing, create a new one on startup.

Item 3 involved copying the docker devmapper library into serviced and modifying it.

### DIFF
--- a/dfs/destroy.go
+++ b/dfs/destroy.go
@@ -22,31 +22,34 @@ import (
 func (dfs *DistributedFilesystem) Destroy(tenantID string) error {
 	vol, err := dfs.disk.Get(tenantID)
 	if err != nil {
-		glog.Errorf("Could not get volume for tenant %s: %s", tenantID, err)
-		return err
-	}
-	snapshots, err := vol.Snapshots()
-	if err != nil {
-		glog.Errorf("Could not get snapshots for tenant %s: %s", tenantID, err)
-		return err
-	}
-	for _, snapshot := range snapshots {
-		if err := dfs.Delete(snapshot); err != nil {
-			glog.Errorf("Could not remove snapshot %s for tenant %s: %s", snapshot, tenantID, err)
-			return err
+		glog.Errorf("Error destroying DFS:  Could not get volume for tenant %s: %s", tenantID, err)
+	} else {
+		// Remove all the stuff we need a volume object for
+		snapshots, err := vol.Snapshots()
+		if err != nil {
+			glog.Errorf("Could not get snapshots for tenant %s: %s", tenantID, err)
+		} else {
+			for _, snapshot := range snapshots {
+				if err := dfs.Delete(snapshot); err != nil {
+					glog.Errorf("Could not remove snapshot %s for tenant %s: %s", snapshot, tenantID, err)
+				}
+			}
+		}
+
+		if err := dfs.unexport(vol.Path()); err != nil {
+			glog.Errorf("Could not unexport path %s: %s", vol.Path(), err)
 		}
 	}
+
 	if err := dfs.deleteImages(tenantID, docker.Latest); err != nil {
-		return err
+		glog.Errorf("Could not delete images for tenant %s: %s", tenantID, err)
 	}
-	if err := dfs.unexport(vol.Path()); err != nil {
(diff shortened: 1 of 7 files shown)


dfs/destroy.go
```diff
@@ -22,31 +22,34 @@ import (
...
-	snapshots, err := vol.Snapshots()
-	if err != nil {
-		glog.Errorf("Could not get snapshots for tenant %s: %s", tenantID, err)
-		return err
-	}
-	for _, snapshot := range snapshots {
-		if err := dfs.Delete(snapshot); err != nil {
-			glog.Errorf("Could not remove snapshot %s for tenant %s: %s", snapshot, tenantID, err)
-			return err
+		glog.Errorf("Error destroying DFS:  Could not get volume for tenant %s: %s", tenantID, err)
```
smousa: do you need to return error here?

volume/devicemapper/devicemapper.go
```diff
@@ -455,29 +473,37 @@ func (d *DeviceMapperDriver) Remove(volumeName string) error {
...
-	snapshots, err := v.Snapshots()
+	v, err := d.loadVolume(volumeName)
 	if err != nil {
-		return err
-	}
-	for _, snapshot := range snapshots {
-		if err := v.RemoveSnapshot(snapshot); err != nil {
-			return err
+		//log the error, but continue trying to remove things
+		glog.Errorf("Error loading volume %s: %s", volumeName, err)
```
smousa: do you need to return error here?

volume/devicemapper/devicemapper.go
```diff
@@ -677,6 +716,14 @@ func (d *DeviceMapperDriver) ensureInitialized() error {
 	return nil
 }
 
+func (d *DeviceMapperDriver) deleteDevice(deviceName string, syncDelete bool) error {
+	if deviceName == "base" || deviceName == "" {
```
smousa: If we fail to delete the device if the hash is empty, should we clean up the device data anyway? Or somehow label it as a defunct volume?

volume/devicemapper/devicemapper.go
```diff
@@ -455,29 +473,37 @@ func (d *DeviceMapperDriver) Remove(volumeName string) error {
...
+
+		// Release the device (requires another call to loadVolume)
+		if err := d.Release(volumeName); err != nil {
+			glog.V(1).Infof("Error releasing device: %s", err)
+		}
+
+		// Delete the device
+		glog.V(1).Infof("Removing volume %s", volumeName)
+		if err := d.deleteDevice(v.deviceHash(), false); err != nil {
+			glog.Errorf("Could not delete device %s: %s", volumeName, err)
```
smousa: missing return

smousa:
Are you just pulling in docker's devicemapper code for these new files?

dfs/destroy.go
johnlhamilton: If we return here, then the dfs.disk.Remove call below will never happen, which we need in order to remove the volume itself (or just the metadata if the volume is gone).

volume/devicemapper/devicemapper.go
johnlhamilton: We could, but I was trying to be consistent with my change, which is to keep trying to remove whatever we can rather than failing at the first error.

volume/devicemapper/devicemapper.go
johnlhamilton: Same thing, if we fail here we won't delete the metadata directory below, which is important.

volume/devicemapper/devicemapper.go
johnlhamilton: How we handle that depends on if this was a snapshot device, a temporary device, or a volume's main device.  In Right now that is handled by the methods that call this one.

johnlhamilton:
Yes, we pulled in docker's code and changed one method from being private to public so that we could create a new base device on startup if something has deleted it.

PR #2792, opened 2016-06-19
Feature/us 14900 acceptance

rphillipsz:


acceptance/ui/features/support/selenium.rb
```diff
@@ -0,0 +1,22 @@
+
+#
+# Modifications to the selenium behavior for our tests.
+#
+
+class Capybara::Selenium::Driver
```
tgjonestx: By default, we use the `selenium_chrome` driver instead of the `selenium` driver  (i.e test against Chrome instead of Firefox). Does this change work for both browsers, or just one?

acceptance/ui/features/api/cli.rb
```diff
@@ -0,0 +1,327 @@
...
+        end
+    end
+
+    # Ensures that only the host provided is there.  If not, remove them and
+    # add the default host.
+    def ensure_only_default_host_exists()
+        host = getTableValue("table://hosts/defaultHost/name")
+        json = get_json("%{serviced} host list -v")
+        if (json.length != 1 || findArrayMatch(json, "Name", host) == nil)
+        #if (page.has_no_content?("Showing 1 Result") || isNotInRows("table://hosts/defaultHost/name"))
```
tgjonestx: Please remove this comment to avoid 'clutter' building up in the source

acceptance/ui/features/steps/services_steps.rb
```diff
@@ -1,164 +1,89 @@
...
-end
-
-def initServicePageObj()
-    if @service_page.nil?
-        @service_page = Service.new
-    end
+    CC.UI.remove_publicendpoint(entry)
 end
+
+When (/^I click the add Add Public Endpoint button$/) do
```
tgjonestx: Please change to "I click the Add Public Endpoint button"

acceptance/ui/features/support/application.rb
```diff
@@ -1,4 +1,5 @@
 require 'uri'
+require 'json'
```
tgjonestx: Is this require used in this file?

acceptance/ui/features/support/selenium.rb
rphillipsz: This should work in both.

acceptance/ui/features/support/application.rb
rphillipsz: No, it was part of some refactoring that got left behind after a few code moves.  I'll remove it.

acceptance/ui/features/api/cli.rb
```diff
@@ -0,0 +1,327 @@
+require "json"
+require "capybara/rspec"
+
+#
+# Encapsulates the CLI functionality to be exposed to the
+# acceptance tests: CC.CLI.<function>
+#
+class CLI
```
tgjonestx: Having all of the generic CLI processing consolidated in one class is a good thing, but I worry about this class becoming unmanageably large as we add more and more methods to interface with more and more commands. Would it make more sense to have a different subclasses for different commands. Something like HostCLI for the host commands, PoolCLI for the pool commands etc?

acceptance/ui/features/api/ui.rb
```diff
@@ -0,0 +1,231 @@
...
+    end
+
+    def UserPage
+        suppress_deploy_wizard()
+        return @pages[:user]
+    end
+
+    # Look up the table data for the given port and remove it using
+    # the UI.
+    def remove_publicendpoint_port_json(name)
```
tgjonestx: These methods related to publicendpoints seem out of place in this class since it is mostly a 'container' class for the different pages, as well has "global" methods like the login stuff. 

How about creating a Page object for the Add-public-endpoints dialog and adding these methods there? Alternatively,  you could move these endpoint-related methods to the Service page object.

PR #2796, opened 2016-06-21
CC-2339: closing dangling pipes

smousa:


smousa:
would like some feedback from @tgjonestx

PR #2797, opened 2016-06-21
Adding serviced config zookeeper quorum generator

ed-:


smousa:
would like some feedback from @tgjonestx

tgjonestx:
I'm not convinced that I want to add it to the product this way - in other words, not ready to distribute this to customers just yet.  But capturing it in source is fine for now.  

This could be something that we could reuse in the future Configuration Tool that's been discussed, though for that tool, I'm leaning towards you run it once, get N output files for N quorum members, and then distribute those files to the respective machines

PR #2807, opened 2016-06-27
Add CLI for 'vhost add' and tests

rphillipsz:
Rally US-15116

### DIFF
--- a/acceptance/ui/features/cli.feature
+++ b/acceptance/ui/features/cli.feature
@@ -10,27 +10,49 @@ Feature: CLI Validation
      And only the default host is added
      And that the "table://applications/defaultApp/template" application with the "table://applications/defaultApp/id" Deployment ID is added
 
+  @port
   Scenario: List the port public endpoints
     Given that the "port0" port is added
-    Then I should see the port public endpoint "port0" in the service
+    Then I should see the port public endpoint "port0" in the list output
 
+  @port
   Scenario: Add a new port public endpoint
     Given that the "port1" port does not exist
       And that the "port1" port is added
     Then I should see the port public endpoint "port1" in the service
 
+  @port
   Scenario: Delete the port public endpoint
     Given that the "port0" port is added
       And the port public endpoint "port0" is removed
     Then I should not see the port public endpoint "port0" in the service
 
+  @port
   Scenario: Disable a port public endpoint
     Given that the "port0" port is added
       And that the port public endpoint "port0" is disabled
     Then the port public endpoint "port0" should be "disabled"
 
+  @port
   Scenario: Disable and enable a port public endpoint
     Given that the "port0" port is added
       And that the port public endpoint "port0" is disabled
       And that the port public endpoint "port0" is enabled
     Then the port public endpoint "port0" should be "enabled"
+
+  @vhost
+  Scenario: List the vhost public endpoints
+    Given that the "vhost0" vhost is added
+    Then I should see the vhost public endpoint "vhost0" in the list output
+
+  @vhost
+  Scenario: Add a new vhost public endpoint
+    Given that the "vhost1" vhost does not exist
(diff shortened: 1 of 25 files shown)


facade/publicendpoint.go
```diff
@@ -232,3 +232,73 @@ func (f *Facade) EnablePublicEndpointPort(ctx datastore.Context, serviceid, endp
...
+		return nil, err
+	}
+
+	// check other virtual hosts for redundancy
+	vhostLowerName := strings.ToLower(vhostName)
+	if services, err := f.GetAllServices(ctx); err != nil {
+		err = fmt.Errorf("Could not get the list of services: %s", err)
+		glog.Error(err)
+		return nil, err
+	} else {
```
tgjonestx: The `else` here isn't really necessary - all it does is add a level of indentation.

facade/publicendpoint.go
```diff
@@ -232,3 +232,73 @@ func (f *Facade) EnablePublicEndpointPort(ctx datastore.Context, serviceid, endp
 	glog.V(2).Infof("Service (%s) updated", svc.Name)
 	return nil
 }
+
+// Adds a vhost public endpoint to a service
+func (f *Facade) AddPublicEndpointVHost(ctx datastore.Context, serviceid, endpointName, vhostName string, isEnabled, restart bool) (*servicedefinition.VHost, error) {
```
tgjonestx: The `restart` argument here is not used in this method AFAIK. 

Looks like the same argument in the `AddPublicEndpointPort()` method is not used.

facade/publicendpoint.go
rphillipsz: Without the else, services has to be declared outside of the if() statement.  I can do that.

facade/publicendpoint_test.go
```diff
@@ -289,3 +289,91 @@ func (ft *FacadeIntegrationTest) Test_PublicEndpoint_PortEnable(c *C) {
...
+	c.Assert(ft.Facade.AddService(ft.CTX, svcA), IsNil)
+	c.Assert(ft.Facade.AddService(ft.CTX, svcB), IsNil)
+
+	// Mock call expectations:
+	ft.zzk.On("CheckRunningPublicEndpoint", registry.PublicEndpointKey(":22222-1"), svcA.ID).Return(nil)
+	ft.zzk.On("CheckRunningPublicEndpoint", registry.PublicEndpointKey("zproxy-0"), svcA.ID).Return(nil)
+	ft.zzk.On("CheckRunningPublicEndpoint", registry.PublicEndpointKey("zproxy2-0"), svcA.ID).Return(nil)
+
+	// Add a vhost to an invalid service.
+	_, err := ft.Facade.AddPublicEndpointVHost(ft.CTX, "invalid", "zproxy", "zproxy", true, true)
```
tgjonestx: You don't have to change this, but for my $0.02, I prefer to see each of these calls to `AddPublicEndpointVHost()` as separate test cases where the name of the test method clarifies what is being tested.

facade/publicendpoint.go
rphillipsz: Good catch.  I missed this in the port method, and used that as a guide.

acceptance/ui/features/cli.feature
```diff
@@ -10,27 +10,49 @@ Feature: CLI Validation
...
 
+  @port
   Scenario: Disable and enable a port public endpoint
     Given that the "port0" port is added
       And that the port public endpoint "port0" is disabled
       And that the port public endpoint "port0" is enabled
     Then the port public endpoint "port0" should be "enabled"
+
+  @vhost
+  Scenario: List the vhost public endpoints
```
tgjonestx: No need to do this now, but at some point we may want separate CLI feature files for different CLI features rather than all CLI calls in one file.

PR #2810, opened 2016-06-28
Feature/us 15118

rphillipsz:


acceptance/ui/features/api/cli_service.rb
```diff
@@ -148,6 +161,27 @@ def check_publicendpoint_port_exists_in_service?(service, endpoint, portAddr, pr
             return false
         end
 
+        # Looks up the given vhost from the vhosts table, and returns the enabled
+        # state based on the service definition.
+        def check_publicendpoint_vhost_enabled_in_service_json?(name)
```
achow-zenoss: Thie code looks OK as is but it looks regular enough. Is it possible to have a common implementation somthing like check_publicpoint_enabled_in_service_json? with "port" and "vhost" as the parameter? It may simplify future maintenance. It applies to some other tests as well.

cli/api/interfaces.go
```diff
@@ -126,4 +126,6 @@ type API interface {
 	RemovePublicEndpointPort(serviceid, endpointName, portAddr string) error
 	EnablePublicEndpointPort(serviceid, endpointName, portAddr string, isEnabled bool) error
 	AddPublicEndpointVHost(serviceid, endpointName, vhost string, isEnabled, restart bool) (*servicedefinition.VHost, error)
+
+	EnablePublicEndpointVHost(serviceid, endpointName, vhost string, isEnabled bool) error
```
achow-zenoss: There are places using VHost or Vhost. We might want to make it consistent for future maintenance.
I saw Vhost used in more places.

facade/publicendpoint.go
```diff
@@ -304,3 +314,47 @@ func (f *Facade) AddPublicEndpointVHost(ctx datastore.Context, serviceid, endpoi
 	glog.V(2).Infof("Service %s updated after adding vhost public endpoint (%s)", svc.Name, vhost.Name)
```
achow-zenoss: Should this be "after adding port" ?

acceptance/ui/features/api/cli_service.rb
rphillipsz: I started to do this, but the two have slightly different data members.  The service check is checking for the existence of different pieces of data between the two of them and the json data has different keys/number of values.

PR #2817, opened 2016-07-06
Feature/wan support

smousa:
This code is used to ensure service connectivity during a WAN network outage.  Connection timeouts are set by pool and are configurable by CLI only. UI will be available in a separate PR.

Pool connection timeouts should be set based on the max difference in latency between the hosts and the master.

The minimum observed timeout is 2 x the configured connection timeout.  This is to allow a quiesce time when the network is perceived down, since not all hosts may necessarily show that they are down at the exact same time and the time to wait after connectivity has been restored to the network.

I am happy to explain the architecture to anyone who is interested in learning how the scheduling system works and how this change impacts scheduling as a whole.


smousa:
retest this

zzk/service/hostregistry.go
```diff
@@ -14,273 +14,449 @@
...
-
-var (
-	ErrHostInvalid = errors.New("invalid host")
-	ErrShutdown    = errors.New("listener shut down")
-)
-
-func hostregpath(nodes ...string) string {
-	p := append([]string{zkRegistry, zkHost}, nodes...)
-	return path.Clean(path.Join(p...))
+// HostRegistryListener is a monitors the availability of hosts within a pool
```
ed-: s/is a //

cli/cmd/pool.go
```diff
@@ -319,3 +326,37 @@ func (c *ServicedCli) cmdRemoveVirtualIP(ctx *cli.Context) {
...
+// serviced pool set-conn-timeout POOLID TIMEOUT
+func (c *ServicedCli) cmdSetConnTimeout(ctx *cli.Context) {
+	args := ctx.Args()
+	if len(args) != 2 {
+		fmt.Printf("Incorrect Usage.\n\n")
+		cli.ShowCommandHelp(ctx, "set-conn-timeout")
+		return
+	}
+
+	connTimeout, err := time.ParseDuration(args[1])
```
rphillipsz: We should have some cli tests to verify that negative numbers are rejected, and that 0 or positive numbers are accepted as valid.

zzk/service/hoststate.go
```diff
@@ -0,0 +1,307 @@
...
+func NewHostState(state *servicestate.ServiceState) *HostState {
+	return &HostState{
+		HostID:         state.HostID,
+		ServiceID:      state.ServiceID,
+		ServiceStateID: state.ID,
+		DesiredState:   int(service.SVCRun),
+	}
+}
+
+// Version inplements client.Node
```
ed-: s/inp/imp/

cli/cmd/pool.go
smousa: added that to the domain level

PR #2819, opened 2016-07-07
Wait before retrying to connect to zookeeper on startup

glujan:
For CC-2254


smousa:
I don't think you should ever get a nil connection: https://github.com/control-center/serviced/blob/develop/zzk/connection.go#L163

glujan:
Thanks, Summer. I will create separate PR with different approach

PR #2820, opened 2016-07-11
Change retry timeout to get zzk connection

glujan:


zzk/connection.go
```diff
@@ -181,6 +207,9 @@ func (zconn *zconn) monitor(path string) {
 			return
 		}
 
+		delayC = getExponentialDelay()
+		defer close(delayC)
```
glujan: Will this channel be properly closed in each loop run?

glujan:
@smousa Please have a look at this

smousa:
have you looked at this library: https://github.com/jpillora/backoff

5 minutes may be too long for a max backoff.  10-15 seconds seems more reasonable.

glujan:
Previous solution was over-engineered so here is simplified one

zzk/connection.go
```diff
@@ -38,6 +40,34 @@ var (
 // GetConnection describes a generic function for acquiring a connection object
 type GetConnection func(string) (client.Connection, error)
 
+type delay struct {
+	backoff time.Duration
+}
+
+func (d *delay) GetDelay() time.Duration {
+	defer func() {
```
smousa: no need to put this in a defer

zzk/connection.go
```diff
@@ -25,6 +26,7 @@ import (
 
 const (
 	DefaultConnectionTimeout = time.Minute
```
smousa: Is this still used anywhere?

zzk/connection.go
```diff
@@ -38,6 +40,34 @@ var (
...
+func (d *delay) GetDelay() time.Duration {
+	defer func() {
+		factor := 1.25
+		jitter := 0.1
+
+		d.backoff = time.Duration(float64(d.backoff) * factor)
+		if d.backoff > MaxDelay {
+			d.backoff = MaxDelay
+		}
+		d.backoff += time.Duration(rand.NormFloat64() * jitter * float64(time.Second))
```
smousa: This value will stop when it is greater than MaxBackoff
We want it to always be less than or equal to MaxDelay

zzk/connection.go
glujan: While creating _delay_ object user needs to give _baseDelay_ as argument. I want to return the same value for the first time when _GetDelay_ is being called, thus this `defer` is here

zzk/connection.go
glujan: Right, I will fix this

zzk/connection.go
smousa: Please add a comment explaining that.

glujan: Sure

glujan:
Retest this please

PR #2830, opened 2016-07-15
pool.ConnectionTimeout field to milliseconds

smousa:


domain/pool/pool.go
```diff
@@ -32,22 +32,26 @@ type VirtualIP struct {
...
+	MemoryCommitment  uint64      // Amount (bytes) of RAM committed to services
+	ConnectionTimeout int         // Wait delay on service rescheduling when an outage is reported (milliseconds)
 	CreatedAt         time.Time
 	UpdatedAt         time.Time
 	MonitoringProfile domain.MonitorProfile
 	datastore.VersionedEntity
 }
 
+func (p ResourcePool) GetConnectionTimeout() time.Duration {
+	return time.Duration(p.ConnectionTimeout) * time.Millisecond
```
jplouis: should the multiplication be in the parens?

smousa: No, because time.Millisecond is of type time.Duration--I have to cast ConnectionTimeout to be of the same type.

jplouis: My bad.  LGTM.

PR #2831, opened 2016-07-18
Feature/cc 2427

carlcamera:
1 Create new "Edit Resource Pool" Link on Resource Pool Details page
2 Create new popup to edit resource pool timeout value
3 Allow human-readable time duration values such as 30s 1m45s 3d
4 Update the resource pool definition with the correct timeout value via REST API

Note: Text value entered may not match value after save. If the duration entered can be expressed with a larger duration unit, that duration unit will be used. For example, 36h will be converted to 3d, 90s will be converted to 1m30s.


web/ui/src/Pools/edit-pool.html
```diff
@@ -0,0 +1,16 @@
+<div class="form-group">
+  <label for="pool_id">
+      <span translate>label_pool_name</span>:
+  </label>
+  <input readonly class="form-control" id="pool_id" type="text" ng-model="editablePool.ID">
+</div>
+
+<div class="form-group">
+  <label for="edit_pool_timeout">
+      <span translate>label_connection_timeout</span> (eg 2s or 1m):
```
rDr4g0n: This can demonstrate the varying units as well as the fact it can take a bunch of durations at once with something like "1h 3m 30s"

web/ui/src/common/miscUtils.js
```diff
@@ -208,6 +208,67 @@
                 return childCount;
             },
 
+            humanizeDuration: function(msecs) {
+                // converts millisecons to a time duration such as 4h45m20s
+
+                if (msecs === 0) return "0";
+
+                var humanized = "";
+                var ttoken = ["w","d","h","m","s","ms"];
```
rDr4g0n: try moving `tmultiplier` up a line, and changing `ttoken` to: `var ttoken = Object.keys(tmultiplier);`

rDr4g0n: In fact, since `tmultiplier` is used in other functions here, move these guys up to the top of this module so it can be shared.

web/ui/src/common/miscUtils.js
```diff
@@ -208,6 +208,67 @@
...
+                    throw new Error(`Numeric value ${nounit[0]} lacks time unit`);
+                }
+                
+                var humanTokens = human.match(/\d+[a-z]+/g);
+                var timeMultiplier = {w: 6048e5, d: 864e5, h: 36e5, m: 6e4, s: 1e3,  ms: 1};
+                var msecs = humanTokens
+                    .reduce(function(prev, tok){
+                        var tokPart = tok.match(/(\d+)([a-z]+)/);
+                        if (! (tokPart[2] in timeMultiplier)) {
+                            // throw new Error("Unable to convert input " + humanTime + ": invalid time unit " + tokPart[0]);
```
rDr4g0n: remove old code

web/ui/src/common/miscUtils.js
```diff
@@ -208,6 +208,67 @@
...
+                if (badchars) {
+                    throw new Error(`Found ${badchars.length} unallowed characters in time entry: "${badchars.toString()}"`);
+                }
+                var nounit = human.match(/\d+$/g);
+                if (nounit) {
+                    throw new Error(`Numeric value ${nounit[0]} lacks time unit`);
+                }
+                
+                var humanTokens = human.match(/\d+[a-z]+/g);
+                var timeMultiplier = {w: 6048e5, d: 864e5, h: 36e5, m: 6e4, s: 1e3,  ms: 1};
```
rDr4g0n: move `timeMultipler` to shared variable per comment above

web/ui/src/common/miscUtils.js
```diff
@@ -208,6 +208,67 @@
                 return childCount;
             },
 
+            humanizeDuration: function(msecs) {
```
rDr4g0n: util functions need unit tests

web/ui/src/common/miscUtils.js
```diff
@@ -208,6 +208,67 @@
...
+                        humanized += d.toString() + ttoken[i];
+                        msecs -= (d * tmultiplier[ttoken[i]]);
+                    }
+
+                }
+                // unused or duplicate tokens means bad input
+                return humanized;
+            },
+
+            parseDuration: function(humanTime) {
```
rDr4g0n: util functions need unit tests

PR #2836, opened 2016-07-19
Add the ability to exclude volumes from backups.

iancmcc:
https://jira.zenoss.com/browse/CC-2442

Trigger this by setting `ExcludeFromBackups: true` for a particular volume using `serviced service edit`, then run a backup, then restore the backup.


dfs/dfs.go
```diff
@@ -76,7 +76,7 @@ type BackupInfo struct {
 	Templates     []servicetemplate.ServiceTemplate
 	BaseImages    []string
 	Pools         []pool.ResourcePool
-	Snapshots     []string
+	Snapshots     map[string][]string
```
smousa: ??

smousa: Not backwards compatible

volume/rsync/rsync.go
```diff
@@ -649,7 +649,10 @@ func (v *RsyncVolume) Rollback(label string) (err error) {
 }
 
 // Export implements volume.Volume.Export
-func (v *RsyncVolume) Export(label, parent string, writer io.Writer) error {
+func (v *RsyncVolume) Export(label, parent string, writer io.Writer, excludes []string) error {
+	if len(excludes) > 0 {
+		glog.Warning("rsync backups do not support excluding directories")
```
smousa: This could technically support this, because it exports the same way devicemapper does

iancmcc: Technically true, but I don't want to dump effort into the rsync driver since nothing uses it anymore.

PR #2843, opened 2016-07-25
use new isvcs image with java 8 support

morr-zenoss:
For ZEN-23915: use zenoss/serviced-isvcs:v44 to pick up java 8 changes.


morr-zenoss:
Retest this

PR #2848, opened 2016-07-29
feature/CC-2491

smousa:
This is for querying service instance data in the new format.  The new format is still not integrated into the code base.

* Added business logic to parse the zk state information and transform it into a service domain instance object
* Added rest endpoint at /services/:serviceId/instances to return instance info from a service
* Added rest endpoint at /hosts/:hostId/instances to return instance info from a host
* Regenerated facade mocks
* Logging on the facade instances

### DIFF
--- a/facade/zkapi.go
+++ b/facade/zkapi.go
@@ -301,3 +301,26 @@ func (zk *zkf) GetServiceEndpoints(tenantID, serviceID string, result *[]applica
 	}
 	return nil
 }
+
+// GetServiceStates2 returns all running instances for a service
+// FIXME: update name when integration is complete
+func (zk *zkf) GetServiceStates2(poolID, serviceID string) ([]zkservice.State, error) {
+	conn, err := zzk.GetLocalConnection("/")
+	if err != nil {
+		glog.Errorf("Could not get connection to zookeeper: %s", err)
+		return nil, err
+	}
+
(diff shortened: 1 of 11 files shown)


facade/zkapi.go
```diff
@@ -301,3 +301,26 @@ func (zk *zkf) GetServiceEndpoints(tenantID, serviceID string, result *[]applica
 	}
 	return nil
 }
+
+// GetServiceStates2 returns all running instances for a service
+// FIXME: update name when integration is complete
```
morr-zenoss: Does this need to happen now, or are there more commits coming?

facade/zzk.go
```diff
@@ -46,6 +47,8 @@ type ZZK interface {
 	LockServices(svcs []service.Service) error
 	UnlockServices(svcs []service.Service) error
 	GetServiceEndpoints(tenantID, serviceID string, endpoints *[]applicationendpoint.ApplicationEndpoint) error
+	GetServiceStates2(poolID, serviceID string) ([]zkservice.State, error) // FIXME: update when integration is complete
```
morr-zenoss: Same as previous comment - do we fix name now, or later?

facade/zkapi.go
smousa: More commits are coming

facade/zzk.go
smousa: More commits are coming

PR #2850, opened 2016-08-03
Added methods to clean bad host/service states

smousa:
More PRs to come, just getting the easy stuff out of the way

### DIFF
--- a/zzk/service/state.go
+++ b/zzk/service/state.go
@@ -657,6 +657,8 @@ func IsValidState(conn client.Connection, req StateRequest) (bool, error) {
 			Message:   "could not look up host state",
 		}
 	} else if !ok {
+
+		logger.Debug("Host state not found")
 		return false, nil
 	}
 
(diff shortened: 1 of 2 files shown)


zzk/service/state.go
```diff
@@ -629,3 +629,208 @@ func DeleteHostStates(conn client.Connection, poolID, hostID string) (count int)
...
+			"Error": err,
+		}).Debug("Could not look up host state")
+
+		return false, &StateError{
+			Request:   req,
+			Operation: "exists",
+			Message:   "could not look up host state",
+		}
+	} else if !ok {
+		return false, nil
```
johnlhamilton: Should we log that the host state wasn't found here?

zzk/service/state.go
```diff
@@ -629,3 +629,208 @@ func DeleteHostStates(conn client.Connection, poolID, hostID string) (count int)
...
+			"Error": err,
+		}).Debug("Could not look up service state")
+
+		return false, &StateError{
+			Request:   req,
+			Operation: "exists",
+			Message:   "could not look up service state",
+		}
+	} else if !ok {
+		return false, nil
```
johnlhamilton: same as above about logging

PR #2856, opened 2016-08-10
Create a better mux protocol wherein the header has a non-variable size

iancmcc:
This prevents weirdness in scenarios where the message being sent was being
read into the buffer with the mux header, and then somehow failing to be
written down the pipe to the receiving service. This only occurred when TLS was disabled.

This is the first PR to allow CC-2496 to be fixed.

### DIFF
--- a/proxy/mux.go
+++ b/proxy/mux.go
@@ -14,9 +14,9 @@
 package proxy
 
 import (
+	"github.com/control-center/serviced/utils"
 	"github.com/zenoss/glog"
 
-	"bufio"
 	"fmt"
 	"io"
 	"net"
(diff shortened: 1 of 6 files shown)


proxy/mux.go
```diff
@@ -211,7 +177,7 @@ func ProxyLoop(client net.Conn, backend net.Conn,  quit chan bool) {
 			return
 		}
 	}
-//	glog.Infof("transferred %v bytes between %v", transferred, backendAddr)
+	//	glog.Infof("transferred %v bytes between %v", transferred, backendAddr)
```
smousa: delete?

PR #2862, opened 2016-08-12
Pools Pagination (CC-2473 and CC-2475)

cchristiezen:
Adding paging to pools page in Control Center.  Changes on both the front end and backend.


cchristiezen:
retest this

cchristiezen:
retest this

domain/read/pools.go
```diff
@@ -0,0 +1,29 @@
+// Copyright 2014 The Serviced Authors.
```
smousa: Why not make this part of the pools package under domain?

web/api_pools.go
```diff
@@ -0,0 +1,69 @@
+// Copyright 2016 The Serviced Authors.
```
smousa: Can this conversion be on the domain side?

web/ui/src/Pools/PoolsController.js
```diff
@@ -94,20 +122,44 @@
...
-        };
-
         $scope.isDefaultPool = function(poolID) {
           return poolID === "default";
         };
 
+        function updatePools(){
+            resourcesFactory.getV2Pools()
+                .success(data => {
+                    $scope.pools = data.results.map(result => { return new Pool(result); });
```
rDr4g0n: super itty bitty thing, arrow functions dont need brackets and have an implicit return, so you can more succinctly write

`data.results.map(result => new Pool(result));`

web/ui/src/Pools/PoolsController.js
```diff
@@ -6,15 +6,38 @@
...
+    controlplane.controller("PoolsController", ["$scope", "$routeParams",
+        "resourcesFactory", "authService", "$modalService", "$translate",
+        "$notification", "areUIReady", "$interval", "servicedConfig", "log",
+    function($scope, $routeParams, resourcesFactory, authService, $modalService,
+             $translate, $notification, areUIReady, $interval, servicedConfig, log){
+
         // Ensure logged in
         authService.checkLogin($scope);
 
+        var lastUpdate;
```
rDr4g0n: Can you group all of the update related stuff together, and put a comment indicating that its the remnants of poolsfactory? This way itll be easier to identify and remove or factor our in the future.

web/ui/src/Pools/PoolsController.js
```diff
@@ -117,32 +169,27 @@
...
-                    $scope.pools = poolsFactory.poolMap;
-                });
+            updatePools();
 
             $scope.poolsTable = {
                 sorting: {
                     id: "asc"
                 },
                 watchExpression: function(){
-                    // if poolsFactory updates, update view
```
rDr4g0n: maybe keep this comment in?

web/ui/src/Pools/poolsFactory.js
```diff
@@ -7,18 +7,13 @@
...
 
         // share resourcesFactory throughout
         resourcesFactory = _resourcesFactory;
         $q = q;
 
         var newFactory = new BaseFactory(Pool, resourcesFactory.getPools);
-
-        // alias some stuff for ease of use
-        newFactory.poolList = newFactory.objArr;
-        newFactory.poolMap = newFactory.objMap;
```
rDr4g0n: `poolList` is used in a number of other places (deploy wiz, service details, hosts map, hosts), so this needs to stay.

web/ui/src/Pools/poolsFactory.js
```diff
@@ -27,6 +22,7 @@
     // and wraps it with extra functionality and info
     function Pool(pool){
         this.update(pool);
+        this.updateFn = resourcesFactory.getPools;
```
rDr4g0n: this doesnt belong here

web/ui/src/Pools/view-pools.html
```diff
@@ -3,24 +3,24 @@
...
-      <td data-title="'pools_tbl_created_at'|translate" sortable="'model.CreatedAt'">{{pool.model.CreatedAt | date : 'medium'}}</td>
-      <td data-title="'updated_at'|translate" sortable="'model.UpdatedAt'">{{pool.model.UpdatedAt | date : 'medium'}}</td>
-      <td data-title="'pools_tbl_actions'|translate"><button ng-click="clickRemovePool(pool.id)" ng-class="{disabled: isDefaultPool(pool.id)}" class="btn btn-link action" name="{{pool.id}}">
+  <tr ng-repeat="pool in $data">
+    <td data-title="'pools_tbl_id'|translate" sortable="'id'" ng-click="click_pool(pool.id)" class="link">{{pool.id | cut:true:50}}</td>
+    <td data-title="'core_capacity'|translate" sortable="'model.CoreCapacity'">{{pool.model.CoreCapacity}}</td>
+    <td data-title="'memory_usage'|translate" sortable="'model.MemoryCapacity'"><span ng-class="{error: pool.model.MemoryCommitment>pool.model.MemoryCapacity}">{{pool.model.MemoryCommitment | toGB}}</span> / {{pool.model.MemoryCapacity | toGB}}</td>
```
rDr4g0n: any particular reason to change sorting to capacity instead of commitment? removing the `toGB` filter is a good idea though.

web/ui/static/css/main.css
```diff
@@ -1790,7 +1787,7 @@ input[type=file] {
 .jelly-table .loader{
     opacity: 1;
     width: 100%;
-    height: calc(100% - 59px);
+    height: 80%;
```
rDr4g0n: height of 80% is not enough in the case of taller tables like services. The calculated height is much closer to correct, though the hardcoded offset is not completely accurate since the table headers can sometimes be taller than 59px.

web/ui/src/common/jellyTable.js
```diff
@@ -123,65 +125,89 @@
...
-                        if(angular.isUndefined(orderedData)){
-                            $scope[tableID].loading = true;
+                            // single-entry tables that are not arrays pass through once (eg config files)
+                            tableEntries = utils.mapToArr(allItems);
+                            $scope[tableID].loading = false;
                             toggleNoData(false);
-                            orderedData = [];
+                            return;
+
+                        } else if (allItems === null) {
```
rDr4g0n: this seems to be the same as `allItems === undefined`, maybe they should be comined. does `null` ever happen? in js you have to actively set a value to null, so unless its explicitly occurring its not really worth writing a case for. also, `if(!allItems)` would catch both of those cases while ignoring empty arrays or objects

web/ui/src/common/jellyTable.js
```diff
@@ -123,65 +125,89 @@
...
-                            unorderedData = utils.mapToArr(unorderedData);
+                        if (allItems === undefined) {
 
-                        // if it's null, create empty array
-                        } else if(unorderedData === null){
-                            unorderedData = [];
-                        }
+                            // if just intitalized, show loading and default to empty array
+                            allItems = [];
+                            toggleNoData(true);
```
rDr4g0n: `toggleNoData(false)` should be used here. If we have *no* result (aka, undefined), then that doesnt mean there is no data, that means that the data hasnt loaded yet. no data specifically means, we got a result, but the result is an empty list.

web/ui/src/common/jellyTable.js
```diff
@@ -123,65 +125,89 @@
...
-                                orderBy(unorderedData, params.orderBy()) :
-                                unorderedData;
-                        }
-
-                        // if no data, show loading and default
-                        // to empty array
-                        if(angular.isUndefined(orderedData)){
-                            $scope[tableID].loading = true;
+                            // single-entry tables that are not arrays pass through once (eg config files)
+                            tableEntries = utils.mapToArr(allItems);
```
rDr4g0n: this bypasses sorting. is this intentional?

web/ui/src/Pools/PoolsController.js
cchristiezen: Made the change.

web/ui/src/Pools/PoolsController.js
cchristiezen: Everything relating to updating and polling has been grouped together.

web/ui/src/Pools/poolsFactory.js
cchristiezen: Added back in.

web/ui/src/Pools/view-pools.html
cchristiezen: Changed it back to commitment but with toGB removed.

web/ui/src/common/jellyTable.js
cchristiezen: Changed it to false.

smousa:
LGTM--waiting for @rDr4g0n

web/ui/src/common/jellyTable.js
carlcamera: Sorting is maintained and moved below to lines 163-165 click "View full changes" to expand this item.
```
 +                            // call overriden getData if available (eg services)
 +                            if(config().getData){
 +                                sortedItems = config().getData(allItems, params);
 +
                              } else {
 +                                // use default getData (eg pools hosts)
 +                                sortedItems = params.sorting() ? 
 +                                    orderBy(allItems, params.orderBy()) 
 +                                    : allItems;
```

carlcamera:
retest this.

web/ui/src/common/jellyTable.js
rDr4g0n: Chatted with carl via slack about some changes in `getData` that are still being looked at.

web/ui/src/common/jellyTable.js
```diff
@@ -131,68 +131,59 @@
...
-                            // if just intitalized, show loading and default to empty array
-                            allItems = [];
-                            toggleNoData(false);
-                            return;
-
-                        } else if (angular.isObject(allItems) && !angular.isArray(allItems)) {
-
-                            // single-entry tables that are not arrays pass through once (eg config files)
+                        if (angular.isObject(allItems) && !angular.isArray(allItems)) {
+                            // single-entry tables that are not arrays and not null
```
rDr4g0n: this comment is not correct. This block is hit if the data is an object, and is not necessarily single-entry.

web/ui/src/common/jellyTable.js
```diff
@@ -131,68 +131,59 @@
...
                             tableEntries = utils.mapToArr(allItems);
                             $scope[tableID].loading = false;
                             toggleNoData(false);
-                            return;
-
                         } else if (allItems === null) {
-
-                            // if no data, remove loading and show no data message
```
rDr4g0n: leave this comment, or put a similar comment. this code is hit when the backend explicitly sets a value to `null`, which indicates that the request completed successfully but there are no results (as opposed to data being `undefined`, meaning the request has not yet completed).

web/ui/src/common/jellyTable.js
```diff
@@ -131,68 +131,59 @@
...
-                            allItems = [];
-                            toggleNoData(false);
-                            return;
-
-                        } else if (angular.isObject(allItems) && !angular.isArray(allItems)) {
-
-                            // single-entry tables that are not arrays pass through once (eg config files)
+                        if (angular.isObject(allItems) && !angular.isArray(allItems)) {
+                            // single-entry tables that are not arrays and not null
                             tableEntries = utils.mapToArr(allItems);
```
rDr4g0n: `tableEntries` is not sorted

rDr4g0n:
Cool, Im happy with the code! waitin on build to pass.

carlcamera:
retest this.

rDr4g0n:
I just rekicked PR integration and unit test jobs.

PR #2874, opened 2016-08-22
update cc ui shrinkwrap not to point at nexus.zendev

rDr4g0n:
the last shrinkwrap update pointed many packages to our internal npm cache. This points them back to the public npm registry.


tgjonestx:
We need to discuss. The point of using nexus.zendev was to improve build reliability by removing our dependency on the global NPM registry from running and/or packages going away.

PR #2879, opened 2016-08-24
feature/CC-2474: Adding endpoint for hosts with reduced payload.

cchristiezen:
Adding a new v2 endpoint for hosts.  The endpoint returns hosts with only the properties necessary for the hosts page.  Reducing the payload (for example, monitoring profile is not included) to only those properties should help make the hosts page faster for sites with large amounts of hosts.

### DIFF
--- a/domain/host/host.go
+++ b/domain/host/host.go
@@ -82,12 +82,12 @@ type ReadHost struct {
 	RAMLimit      string
 	KernelVersion string
 	KernelRelease string
-	ServiceD      ReadServiceD
+	ServiceD      ReadServiced
 	CreatedAt     time.Time
 	UpdatedAt     time.Time
 }
 
-type ReadServiceD struct {
+type ReadServiced struct {
 	Version string
 	Date    string
 	Release string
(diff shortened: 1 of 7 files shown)


domain/host/host.go
```diff
@@ -72,6 +72,27 @@ type Host struct {
...
+	Memory        uint64
+	RAMLimit      string
+	KernelVersion string
+	KernelRelease string
+	ServiceD      ReadServiceD
+	CreatedAt     time.Time
+	UpdatedAt     time.Time
+}
+
+type ReadServiceD struct {
```
smousa: don't capitalize the d

web/api_hosts.go
```diff
@@ -0,0 +1,36 @@
+package web
```
smousa: boilerplate

web/api_hosts_test.go
```diff
@@ -0,0 +1,157 @@
+// Copyright 2014 The Serviced Authors.
```
smousa: Copywrite year

cchristiezen:
retest this

PR #2882, opened 2016-08-24
service state listener

smousa:
Updates the service state listener to use the new state id format (see https://github.com/control-center/serviced/blob/develop/zzk/service/service.go) if you need a point of reference.


smousa:
Pending host state changes

PR #2886, opened 2016-08-24
Feature/cc 2548

rDr4g0n:
* Create LogSearch angular service which makes it easy to generate a url to a specific kibana search
* Import rison encoder to convert js objects to the format kibana expects
* Make main nav "Log" link open Kibana in a new window with some sensible defaults
* Add "Service Logs" link to each service that opens Kibana in a new window, filtered down to that service
* Add "Instance Logs" link to each instance that opens Kibana in a new window, filtered down to that instance


rDr4g0n:
retest this

web/ui/src/Services/ServiceDetailsController.js
```diff
@@ -684,6 +684,40 @@
             }
         };
 
+        $scope.getServiceLogURL = function(service){
+            if(!service){
+                return "";
+            }
+            let a = LogSearch.getDefaultAppConfig(),
```
smousa: Can you make these variables more descriptive?
Also, add comments for what this is doing

web/ui/src/Services/ServiceDetailsController.js
```diff
@@ -684,6 +684,40 @@
...
+                    query: `fields.service:${service.id} AND fields.instance:* AND message:*`
+                }
+            };
+            a.columns = ["fields.instance","message"];
+
+            return LogSearch.getURL(a, g);
+        };
+
+        $scope.getInstanceLogURL = function(instance){
+            let a = LogSearch.getDefaultAppConfig(),
```
smousa: See above

web/ui/src/common/LogSearch.js
```diff
@@ -0,0 +1,78 @@
...
+            return `
+${baseURL}/app/kibana?#/discover?
+_g=${rison.encode(globalConfig)}&
+_a=${rison.encode(appConfig)}&
+indexPattern=${index}&type=histogram`;
+    }
+
+    class LogSearch {
+        constructor(){
+            // TODO - use $location
```
smousa: TODO? Is this still applicable?

PR #2887, opened 2016-08-24
Make sure we have stats for the dfs root for snapshot

rphillipsz:
CC-2542.  When we take a snapshot and check to make sure we have enough disk space, if we can't get the stats for the DFS root we need to bail rather than crash. CC-2541 displays the serviced log when stats can't be collected.

Note: Hong wasn't able to reproduce the error on the ticket, but if another system exhibits the same problem this will prevent a panic and a Control Center crash.  Root cause is unknown.


rphillipsz:
retest this

rphillipsz:
retest this

rphillipsz:
retest this

PR #2890, opened 2016-08-25
Keep trying to get a ZK connection, rather than bailing out.

rphillipsz:
Don't give up if we can't get a connection.  Loop with Connect until the connection isn't nil.


smousa:
I am currently working on this.  I don't know if it makes sense to make this change.

rphillipsz:
retest this

web/cpserver.go
```diff
@@ -462,19 +458,32 @@ func (sc *ServiceConfig) syncAllVHosts(shutdown <-chan interface{}) error {
...
-		zkServiceVHost := service.ZKServicePublicEndpoints
+		// Keep trying to get a connection to zk unless we shut down.
+		var rootConn client.Connection
+
+		select {
+		case rootConn = <-zzk.Connect("/", zzk.GetLocalConnection):
+		case <-shutdown:
+			return nil
+		}
+		if rootConn == nil {
```
smousa: zzk.Connect does sometimes return nil
https://github.com/control-center/serviced/blob/develop/zzk/connection.go#L145

web/portservers.go
```diff
@@ -394,18 +390,28 @@ func (sc *ServiceConfig) syncAllPublicPorts(shutdown <-chan interface{}) error {
...
-			glog.V(1).Infof("Will retry in 10 seconds to WatchChildren(%s) due to error: %v", zkServicePEPService, err)
-			<-time.After(time.Second * 10)
+		// Keep trying to get a connection to zk unless we shut down.
+		var rootConn client.Connection
+		select {
+		case rootConn = <-zzk.Connect("/", zzk.GetLocalConnection):
+		case <-shutdown:
+			return nil
+		}
+		if rootConn == nil {
```
smousa: See above

web/publicendpoint.go
```diff
@@ -231,29 +219,62 @@ func (sc *ServiceConfig) syncPublicEndpoints(shutdown <-chan interface{}) error
...
+		// public endpoints are at the root level (not pool aware)
+		// Keep trying to get a connection to zk unless we shut down.
+		var poolBasedConn client.Connection
+
+		select {
+		case poolBasedConn = <-zzk.Connect("/", zzk.GetLocalConnection):
+		case <-shutdown:
+			return nil
+		}
+		if poolBasedConn == nil {
```
smousa: see above

rphillipsz:
retest this

PR #2893, opened 2016-08-25
CC-2479: Adding endpoint for minimal hosts for pool.

cchristiezen:
Adding new endpoint to return hosts for a pool with minimal data.

### DIFF
--- a/facade/facade_unit_test.go
+++ b/facade/facade_unit_test.go
@@ -16,6 +16,8 @@
 package facade_test
 
 import (
+	"time"
+
 	datastoremocks "github.com/control-center/serviced/datastore/mocks"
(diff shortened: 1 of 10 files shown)


facade/facade_unit_test.go
```diff
@@ -78,3 +80,24 @@ func (ft *FacadeUnitTest) setupMockDFSLocking() {
...
+	var ok bool
+	var first, second time.Time
+
+	first, ok = params[0].(time.Time)
+	if !ok {
+		return false, "First parameter is not a Time"
+	}
+	second, ok = params[1].(time.Time)
+	if !ok {
+		return false, "Second parameter is not an Time"
```
smousa: you should use errors.New

PR #2906, opened 2016-08-30
ability to skip acceptance test

smousa:
This test is really inconsistent, so I am going to skip it for now.

### DIFF
--- a/acceptance/ui/features/applications.feature
+++ b/acceptance/ui/features/applications.feature
@@ -81,7 +81,7 @@ Feature: Application Management
 
   @clean_hosts @clean_services
   Scenario: Deploy an instance of the default template and add a host
-    Given that the "table://applications/defaultApp/template" application is not added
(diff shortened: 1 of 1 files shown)


acceptance.sh
```diff
@@ -155,6 +155,6 @@ cd ${DIR}/acceptance
 sudo GOPATH=${GOPATH} PATH=${PATH} ./startMockAgents.sh --no-wait
 
 # launch cucumber/capybara with colorized output disabled for better readability in Jenkins
-CUCUMBER_OPTS=--no-color ./runUIAcceptance.sh -a https://${HOSTNAME} $*
+CUCUMBER_OPTS=--no-color ./runUIAcceptance.sh -a https://${HOSTNAME} --tags ~@skip $*
```
tgjonestx: If you add PENDING to the Given statement, then you don't need to specify any extra arguments here

PR #2907, opened 2016-08-30
Revert "CC-2530: Exclude specific directories from application backups"

jpeacock-zenoss:
Reverts control-center/serviced#2894


jpeacock-zenoss:
retest this

PR #2918, opened 2016-09-01
CC-2567 Make the config variable TMPDIR, since that's what we use fro…

spopezen:
…m the environment

We are using TMPDIR from the environment, not TMP, so that's what we should set in config


tgjonestx:
Retest this

PR #2924, opened 2016-09-06
host registry does not delete states from zookeeper if the state has …

smousa:
…an assigned ip

### DIFF
--- a/zzk/service/hostregistry.go
+++ b/zzk/service/hostregistry.go
@@ -254,7 +254,7 @@ func (h *HostRegistryListener) Spawn(cancel <-chan interface{}, hostid string) {
 							// Only reschedule services without address
 							// assignments.
 							count := DeleteHostStatesWhen(h.conn, h.poolid, hostid, func(s *State) bool {
-								return s.DesiredState == service.SVCStop || s.AssignedIP == ""
+								return s.DesiredState == service.SVCStop || !s.Static
 							})
 							logger.WithField("unscheduled", count).Warn("Host is experiencing an outage.  Cleaned up orphaned nodes")
 
--- a/zzk/service/hostregistry.go
+++ b/zzk/service/hostregistry.go
@@ -20,6 +20,7 @@ import (
 	log "github.com/Sirupsen/logrus"
 	"github.com/control-center/serviced/coordinator/client"
 	"github.com/control-center/serviced/domain/host"
+	"github.com/control-center/serviced/domain/service"
 )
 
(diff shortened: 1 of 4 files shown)


zzk/service/hostregistry.go
```diff
@@ -250,8 +251,21 @@ func (h *HostRegistryListener) Spawn(cancel <-chan interface{}, hostid string) {
 						// soon as possible.
 						select {
 						case <-h.isOnline:
-							count := DeleteHostStates(h.conn, h.poolid, hostid)
+							// Only reschedule services without address
+							// assignments.
+							count := DeleteHostStatesWhen(h.conn, h.poolid, hostid, func(s *State) bool {
+								return s.DesiredState == service.SVCStop || s.AssignedIP == ""
```
jplouis: Will this prevent scheduling services that have a VIP assignment as well as a static?  Is that something we watn?

smousa: That is a good point.  Let me see if I can figure out a way to handle virtual ips.

zzk/service/hostregistry.go
```diff
@@ -250,8 +251,21 @@ func (h *HostRegistryListener) Spawn(cancel <-chan interface{}, hostid string) {
...
+							// Only reschedule services without address
+							// assignments.
+							count := DeleteHostStatesWhen(h.conn, h.poolid, hostid, func(s *State) bool {
+								return s.DesiredState == service.SVCStop || !s.Static
+							})
 							logger.WithField("unscheduled", count).Warn("Host is experiencing an outage.  Cleaned up orphaned nodes")
+
+							// To prevent a tight loop, wait for something to
+							// happen.
+							select {
```
jplouis: Is there no timeout here needed?  Why not go back to the top of this inner select?

smousa: Why do you think we need a timeout?
If it goes back to the inner select, it will fall through and try to delete the states all over again

jplouis: Was more a question. I'm not familiar with how/when it will actually reschedule

zzk/service/hostregistry.go
```diff
@@ -250,8 +251,21 @@ func (h *HostRegistryListener) Spawn(cancel <-chan interface{}, hostid string) {
...
+							case <-availEv:
+							case <-onlineEv:
+							case <-cancel:
```
jplouis: Don't understand when it falls out of all these nested select blocks.

smousa: I forgot it.  Will add

PR #2928, opened 2016-09-07
CC-2598: New Endpoint for Service Children.

cchristiezen:
Adding a new endpoint which returns the children for a service.  Only the minimal data is sent back.

### DIFF
--- a/domain/service/api.go
+++ b/domain/service/api.go
@@ -1,0 +1,42 @@
+// Copyright 2016 The Serviced Authors.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package service
+
+import (
+	"github.com/control-center/serviced/datastore"
+	"github.com/control-center/serviced/validation"
+)
+
+// Details is the minimal data necessary to show for a service
+type Details struct {
+	ID              string
+	Name            string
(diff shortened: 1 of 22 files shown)


domain/service/api.go
```diff
@@ -0,0 +1,42 @@
...
+
+package service
+
+import (
+	"github.com/control-center/serviced/datastore"
+	"github.com/control-center/serviced/validation"
+)
+
+// Details is the minimal data necessary to show for a service
+type Details struct {
```
smousa: Call this ServiceDetails

smousa: I think this file should be called servicedetails.go and the _store.go file should be named servicedetails_store.go

domain/service/api_store.go
```diff
@@ -0,0 +1,62 @@
...
+
+	"github.com/control-center/serviced/datastore"
+	"github.com/control-center/serviced/datastore/elastic"
+)
+
+// GetChildServiceDetails returns service details given parent service id
+func (s *storeImpl) GetChildServiceDetails(ctx datastore.Context, parentID string) ([]Details, error) {
+	id := strings.TrimSpace(parentID)
+	if id == "" {
+		return nil, errors.New("empty parent service id not allowed")
```
smousa: empty parent id could still be valid, if we want to pull the tenant services

web/routes.go
```diff
@@ -123,6 +123,7 @@ func (sc *ServiceConfig) getRoutes() []rest.Route {
 		rest.Route{"GET", "/api/v2/pools", gz(sc.checkAuth(getPools))},
 		rest.Route{"GET", "/api/v2/pools/:poolId/hosts", gz(sc.checkAuth(getHostsForPool))},
 		rest.Route{"GET", "/api/v2/hosts", gz(sc.checkAuth(getHosts))},
+		rest.Route{"GET", "/api/v2/services/:serviceId/services", gz(sc.checkAuth(getChildServiceDetails))},
```
smousa: could this route be confusing, should we call it /api/v2/services/:serviceId/children ?

smousa:
We are not doing any kind of paging for this, correct?

PR #2941, opened 2016-09-08
Added methods to encode/decode mux header

fraibacas:
Added method to build/extract the mux auth header. This code is not yet wired to the mux bc we still need to have the JWT part implemented.

To wire it we will need to:
- Build header: https://github.com/control-center/serviced/blob/develop/container/proxy.go#L261
- Extract header: https://github.com/control-center/serviced/blob/develop/proxy/mux.go#L145
    once we get the identity, drop the connection if Valid() return an error

### DIFF
--- a/auth/mux.go
+++ b/auth/mux.go
@@ -1,0 +1,119 @@
+package auth
+
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+)
+
+/*
+   When establishing a connection to the mux, in addition to the address of the receiver,
+   the sender sends an authentication token and signs the whole message. The token determines
+   if the sender is authorized to send data to the receiver or not
+
+   ----------------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


auth/mux.go
```diff
@@ -0,0 +1,119 @@
...
+	"errors"
+)
+
+/*
+   When establishing a connection to the mux, in addition to the address of the receiver,
+   the sender sends an authentication token and signs the whole message. The token determines
+   if the sender is authorized to send data to the receiver or not
+
+   ----------------------------------------------------------------------------------------
+   | address (6 bytes) |  token length (4 bytes)  |   Auth Token (N bytes)  |  Signature  |
```
iancmcc: Maybe totally academic, but I'd put the address after the token, so you don't read those extra 6 bytes if the token is invalid.

PR #2945, opened 2016-09-08
Added methods to encode/decode mux header

fraibacas:
Added method to build/extract the mux auth header. This code is not yet wired to the mux bc we still need to have the JWT part implemented.

To wire it we will need to:

Build header: https://github.com/control-center/serviced/blob/develop/container/proxy.go#L261
Extract header: https://github.com/control-center/serviced/blob/develop/proxy/mux.go#L145 once we get the identity, drop the connection if Valid() return an error

### DIFF
--- a/auth/mux.go
+++ b/auth/mux.go
@@ -15,6 +15,7 @@ package auth
 
 import (
 	"bytes"
+	"crypto"
 	"encoding/binary"
 	"errors"
 )
@@ -33,6 +34,7 @@ var (
 	endian           = binary.BigEndian
 	ErrBadMuxAddress = errors.New("Bad mux address")
 	ErrBadMuxHeader  = errors.New("Bad mux header")
+	ErrBadToken      = errors.New("Could not extract token")
 )
 
 const (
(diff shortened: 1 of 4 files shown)


auth/mux.go
```diff
@@ -0,0 +1,132 @@
...
+	signed_message := rawHeader[:offset]
+
+	// Whatever is left is the signature
+	signature := rawHeader[offset:]
+
+	// Verify the identity of the signed message
+	senderVerifier, err := senderIdentity.Verifier()
+	if err != nil {
+		return errorExtractingHeader(err)
+	} else {
```
iancmcc: No need for an else block here.

auth/mux.go
```diff
@@ -0,0 +1,132 @@
...
+	headerBuf.Write(signature)
+
+	return headerBuf.Bytes(), nil
+}
+
+func errorExtractingHeader(err error) (string, Identity, error) {
+	return "", nil, err
+}
+
+func ExtractMuxHeader(rawHeader []byte) (string, Identity, error) {
```
iancmcc: The first return type should be `[]byte`. The address isn't a string. It's packed. Four bytes for the IP and two for the port.

auth/mux.go
```diff
@@ -0,0 +1,132 @@
...
+*/
+
+var (
+	endian           = binary.BigEndian
+	ErrBadMuxAddress = errors.New("Bad mux address")
+	ErrBadMuxHeader  = errors.New("Bad mux header")
+)
+
+const (
+	ADDRESS_BYTES   = 6
```
iancmcc: In Golang, stylistically, constants shouldn't be all caps. AddressBytes and TokenLenBytes are more appropriate.

PR #2950, opened 2016-09-09
only write ports and vhosts to zookeeper if they are enabled

smousa:


tgjonestx:
Is there a negative test-case to show that disabled vhosts/ports are NOT written to zookeeper?

PR #2956, opened 2016-09-09
Refactor: replace glog w/logrus across all domain/* packages

tgjonestx:


domain/host/host.go
```diff
@@ -188,13 +193,26 @@ func Build(ip string, rpcport string, poolid string, memory string, ipAddrs ...s
 	}
 	host, err := currentHost(ip, rpcPort, poolid)
 	if err != nil {
-		glog.V(2).Infof("currentHost failed: %s", err)
+		plog.WithError(err).WithFields(log.Fields{
+			"ip": ip,
+			"rpcPort": rpcPort,
+			"poolID": poolid,
```
iancmcc: The convention for field names we've been going with so far is all lowercase. Reduces clutter. That said, we could undo that convention if you feel strongly in the other direction.

domain/host/host.go
```diff
@@ -188,13 +193,26 @@ func Build(ip string, rpcport string, poolid string, memory string, ipAddrs ...s
...
+		}).Debug("currentHost failed")
 		return nil, err
 	}
-	glog.Infof("Building host %s (%s) with ipsAddrs: %v [%d]", host.ID, host.IPAddr, ipAddrs, len(ipAddrs))
+	plog.WithFields(log.Fields{
+		"hostID": host.ID,
+		"ipAddr": host.IPAddr,
+		"ipAddrs": ipAddrs,
+		"len(ipAddrs)": len(ipAddrs),
```
iancmcc: This shouldn't be logged here, or it should be logged at debug level. Audience for Info-level is an operator, and should only be logged after positive work is completed (e.g., `log.WithFields(logrus.Fields{"ipaddress": hostIP}).Info("Created host")`). Also, `len(ipAddrs)` is not a field we would want to publish, probably. Field names should be descriptive in human terms.

domain/host/host_test.go
```diff
@@ -56,7 +55,7 @@ func init() {
 
 	ip, err := utils.GetIPAddress()
 	if err != nil {
-		glog.Errorf("Could not determine ip %v", err)
+		plog.WithError(err).Error("Could not determine ip")
```
iancmcc: Can we use t.Error rather than Logrus here?

domain/host/hoststore.go
```diff
@@ -104,14 +103,14 @@ func HostKey(id string) datastore.Key {
 
 func convert(results datastore.Results) ([]Host, error) {
 	hosts := make([]Host, results.Len())
-	glog.V(4).Infof("Results are %v", results)
+	plog.WithField("results", results).Debug("convert results", results)
```
iancmcc: I'm not sure this is really worth logging.

domain/host/utils.go
```diff
@@ -57,7 +56,11 @@ func currentHost(ip string, rpcPort int, poolID string) (host *Host, err error)
 	host.Name = hostname
 	hostidStr, err := utils.HostID()
 	if err != nil {
-		glog.V(2).Infof("HostID failed: %s", err)
+		plog.WithError(err).WithFields(log.Fields{
+			"ip": ip,
+			"rpcPort": rpcPort,
+			"poolID": poolID,
+		}).Debug("HostID failed")
```
iancmcc: "Unable to retrieve host ID"

domain/host/utils.go
```diff
@@ -97,7 +100,11 @@ func currentHost(ip string, rpcPort int, poolID string) (host *Host, err error)
 
 	routes, err := utils.RouteCmd()
 	if err != nil {
-		glog.V(2).Infof("RouteCmd failed: %s", err)
+		plog.WithError(err).WithFields(log.Fields{
+			"ip": ip,
+			"rpcPort": rpcPort,
+			"poolID": poolID,
+		}).Debug("RouteCmd failed")
```
iancmcc: Ditto, although this is debug level so we have more leeway. But try to think of support trying to figure something out based on your message.

domain/host/host.go
tgjonestx: I think camel-case is much more readable.

domain/host/utils.go
```diff
@@ -133,7 +140,10 @@ func getIPResources(hostID string, hostIP string, staticIPs ...string) ([]HostIP
 	if err != nil {
 		return nil, err
 	}
-	glog.V(3).Infof("Interfaces on this host %s: %+v", hostID, ifacemap)
+	plog.WithFields(log.Fields{
+		"hostID": hostID,
+		"ifacemap": ifacemap,
```
iancmcc: Where you're going to be logging the same fields over and over, redefine your logger at the top. For example:
```
plog := log.WithFields(logrus.Fields{
    "host": hostID,
})
...
log.WithFields(logrus.Fields{
    "interfaces": ifacemap,
}).Debug("Found host interfaces")

File: domain/host/utils.go [hunk 8]
```diff
@@ -147,7 +157,11 @@ func getIPResources(hostID string, hostIP string, staticIPs ...string) ([]HostIP
 
 	hostIPResources := make([]HostIPResource, len(ips))
 	for i, ip := range ips {
-		glog.Infof("Checking IP '%s'", ip)
+		plog.WithFields(log.Fields{
+			"hostID": hostID,
+			"hostIP": hostIP,
+			"ip": ip,
+		}).Info("Checking IP")
```
iancmcc: Log positive work at info, code flow at debug

domain/host/utils.go
```diff
@@ -171,15 +185,15 @@ func getIPResources(hostID string, hostIP string, staticIPs ...string) ([]HostIP
...
+		plog.WithError(err).Error("Unable to read network interfaces")
 		return nil, err
 	}
 	//make a  of all ipaddresses to interface
 	ips := make(map[string]net.Interface)
 	for _, iface := range interfaces {
 		addrs, err := iface.Addrs()
 		if err != nil {
-			glog.Error("Problem reading interfaces: ", err)
+			plog.WithError(err).Error("Unable to read interface addresses")
```
iancmcc:
```
log.WithFields(logrus.Fields{
    "interface": iface.IDOrWhatever,
}).WithError(err).Error("Unable to read interface addresses")
```

domain/host/validation.go
```diff
@@ -42,21 +39,30 @@ func (h *Host) ValidEntity() error {
...
 	//TODO: what should we be validating here? It doesn't seem to work for
-	glog.V(4).Infof("Validating IPAddr %v for host %s", h.IPAddr, h.ID)
 	ipAddr, err := net.ResolveIPAddr("ip4", h.IPAddr)
 
 	if err != nil {
-		glog.Errorf("Could not resolve: %s to an ip4 address: %v", h.IPAddr, err)
+		plog.WithFields(log.Fields{
+			"hostID": h.ID,
+			"ipAddr": h.IPAddr,
+		}).WithError(err).Error("Could not resolve value to an ip4 address")
```
iancmcc: Might this be more appropriate to log at Debug or Warn level, since otherwise it'll be double-logged where ValidEntity() is called?

domain/metric.go
```diff
@@ -154,7 +154,7 @@ func (builder *MetricBuilder) Config(ID, Name, Description, Start string) (*Metr
 	//build the query body object
 	bodyBytes, err := json.Marshal(request)
 	if err != nil {
-		glog.Errorf("Failed to marshal query body: %+v", err)
+		plog.WithError(err).Error("Failed to marshal query body")
```
iancmcc: This should probably be Warn, unless it represents an immediate call to action.

smousa:
Let me know what the consensus is on whether log variables should be lower or camel, or whatever

tgjonestx:
Retest this

tgjonestx:
Retest this

tgjonestx:
retest this

domain/service/evaluate.go
```diff
@@ -429,48 +459,49 @@ func newRuntimeContext(svc *Service, instanceID int) *runtimeContext {
...
 		return err
 	}
 	if err = service.EvaluateHealthCheckTemplate(getSvc, findChild, instanceID); err != nil {
-		glog.Errorf("%+v", err)
+		logger.Error()
 		return err
 	}
 	if err = service.EvaluateEnvironmentTemplate(getSvc, findChild, instanceID); err != nil {
-		glog.Errorf("%+v", err)
+		logger.Error()
```
iancmcc: Does this work? It doesn't make a copy up above in the `WithError` call?

domain/service/json.go
```diff
@@ -58,21 +58,24 @@ func (e *ServiceEndpoint) UnmarshalJSON(b []byte) error {
...
 		//VHostList is defined, keep it and unset deprecated field if set
 		e.VHosts = nil
 		return nil
 	}
 	if len(e.VHosts) > 0 {
 		// no VHostsList but vhosts is defined. Convert to VHostsList
-		if glog.V(2) {
-			glog.Warningf("ServiceEndpoint VHosts field is deprecated, see VHostList: %#v", e.VHosts)
+		if log.GetLevel() == log.DebugLevel {
+			logger.WithField("vhosts", e.VHosts).Warning("The field named VHosts in ServiceEndpoint is deprecated, see VHostList")
```
iancmcc: What is the logic behind logging a warning if we're in debug level?

domain/servicedefinition/utils.go
```diff
@@ -116,7 +118,10 @@ func getServiceDefinition(path string) (serviceDef *ServiceDefinition, err error
 			}
 			filters, err := getFiltersFromDirectory(path + "/" + subpath.Name())
 			if err != nil {
-				glog.Errorf("Error fetching filters at "+path, err)
+				plog.WithError(err).WithFields(log.Fields{
+					"path": path,
+					"subpath": subpath.Name(),
+				}).Error("rror fetching filters")
```
iancmcc: Typo. I would suggest something like "Unable to fetch filters" rather than to log "ERROR: error"

domain/service/json.go
tgjonestx: I think the idea is that in the normal case (no debug), we will silently migrate the data for you if you have an older service definition, but if you turn on debug, it's a warning to indicate that while we upgraded the data, we really don't expect to encounter very many of these cases in practice (i.e. you have a pretty old service definition if we're migrating this stuff on the fly).

PR #2958, opened 2016-09-12
Feature/hostkeys

johnlhamilton:
host keys

### DIFF
--- a/auth/localkeys.go
+++ b/auth/localkeys.go
@@ -17,6 +17,12 @@ import (
 	"crypto"
 	"io/ioutil"
 	"os"
+	"path"
+)
+
+const (
+	DelegateKeyFileName = ".keys/delegate.keys"
+	MasterKeyFileName   = ".keys/master.keys"
 )
 
 var (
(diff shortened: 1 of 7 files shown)


cli/api/daemon.go
```diff
@@ -423,6 +424,12 @@ func (d *daemon) startMaster() (err error) {
 		}).WithError(err).Fatal("Unable to register master as host")
 	}
 
+	// Load keys if they exist, else generate them
```
iancmcc: a) Include a positive Info-level message after this is done
b) Include the file where you're trying to load/create as a field in the logger
c) Maybe distinguish in the logging between creation and loading? Might help later.

```
keylog := log.WithFields(logrus.Fields{
    "keyfile": masterKeyFile,
})
...
keylog.Info("Loaded master keys from disk")
```

auth/localkeys.go
```diff
@@ -0,0 +1,180 @@
...
+	delegateKeys = HostKeys{pub, priv}
+	return nil
+}
+
+// CreateOrLoadMasterKeys will load the master keys from disk
+//  If the file does not exist, it will generate new keys and
+//  write them to disk.
+func CreateOrLoadMasterKeys(filename string) error {
+	if _, err := os.Stat(filename); os.IsNotExist(err) {
+		os.MkdirAll(path.Dir(filename), os.ModeDir|0700)
```
iancmcc: Check the result for an error

PR #2965, opened 2016-09-14
Wire up delegate to request an authentication token on startup

iancmcc:


rpc/master/hosts_server.go
```diff
@@ -81,3 +89,52 @@ func (s *Server) FindHostsInPool(poolID string, hostReply *[]host.Host) error {
...
+		return err
+	}
+	if err := req.valid(keypem); err != nil {
+		return err
+	}
+	host, err := s.f.GetHost(s.context(), req.HostID)
+	if err != nil {
+		return err
+	}
+	signed, expires, err := auth.CreateJWTIdentity(host.ID, host.PoolID, true, true, keypem, time.Duration(1*time.Hour))
```
jafcrocker: Should we make the expiration time configurable?

cli/api/daemon.go
```diff
@@ -625,6 +625,34 @@ func (d *daemon) startAgent() error {
...
 	})
 
+	// Load delegate keys if they exist
+	delegateKeyFile := path.Join(options.IsvcsPath, auth.DelegateKeyFileName)
+	keylog := log.WithFields(logrus.Fields{
+		"keyfile": delegateKeyFile,
+	})
+
+	go func() {
+		auth.WatchForDelegateKeys(delegateKeyFile, d.shutdown)
```
jafcrocker: Should we continue to watch for changes to the key file after it has been loaded the first time?  Otherwise if we need to reissue the key we will have to restart the delegate.

PR #2967, opened 2016-09-14
CC-2640: add cli option to retrieve host public key

fraibacas:


cli/cmd/key.go
```diff
@@ -0,0 +1,60 @@
...
+		return
+	}
+	hostID := args[0]
+	if host, err := c.driver.GetHost(hostID); host == nil || err != nil {
+		fmt.Fprintln(os.Stderr, "Host not found.")
+		return
+	}
+	key, err := c.driver.GetHostPublicKey(hostID)
+	if err != nil {
+		fmt.Fprintln(os.Stderr, "Could not retrieve host's public key.")
```
jafcrocker: Should we output the error?  It may give useful information about *why* we could not retrieve the public key.

fraibacas: I tried but the error is not very self descriptive:
*Could not retrieve host's public key. %!(EXTRA *errors.errorString=invalid error <nil>, string=*

PR #2971, opened 2016-09-15
Feature/cc 2590

morr-zenoss:
This is for CC-2590 - Store only the data we need for a service in ZK

The interesting changes are in:
* node/agent.go
* node/container.go
* node/agent_test.go
* commons/docker/test/mockEventMonitor.go
* zzk/service/hoststate.go
* zzk/service/lock.go
* zzk/service/service.go
* zzk/service/servicestate.go
* zzk/service/hostregistry_test.go
* zzk/service/hoststate_test.go
* zzk/service/mocks/HostStateHandler.go
* zzk/service/servicestate_test.go
The rest of the files are formatting changes due to intelliJ running gofmt on save.


node/agent.go
```diff
@@ -322,8 +331,35 @@ func (a *HostAgent) setupContainer(masterClient master.ClientInterface, svc *ser
...
+					PortNumber:     endpoint.PortNumber,
+					PortTemplate:   endpoint.PortTemplate,
+					VirtualAddress: endpoint.VirtualAddress,
+				})
 			}
 		}
+		state.AssignedIP = assignedIP
+		state.Static = static
+	} else {
+		log.Warn("svc.endpoints was nil.")
```
smousa: I don't think this is worth warning about--in fact this is a case that will never happen

node/agent.go
```diff
@@ -825,3 +872,44 @@ func waitForSsNodes(processing map[string]chan int, ssResultChan chan stateResul
 	glog.V(0).Info("All service state nodes are shut down")
 	return
 }
+
+func (a *HostAgent) createContainer(conf *dockerclient.Config, hostConf *dockerclient.HostConfig, svcID string, instanceID int) (*docker.Container, error) {
```
smousa: Don't call it createContainer, because we already have a CreateContainer and that doesn't directly call this

node/agent.go
```diff
@@ -529,7 +556,27 @@ func (a *HostAgent) setupContainer(masterClient master.ClientInterface, svc *ser
 			Hard: 0,
 		},
 	}
-	return cfg, hcfg, nil
+
+	ctr, err := a.createContainer(cfg, hcfg, svc.ID, instanceID)
+	if err != nil {
+		logger.WithFields(log.Fields{
+			"imageUUID":  imageName,
```
smousa: is imageName the name of the image or the image hash?

node/agent.go
```diff
@@ -825,3 +872,44 @@ func waitForSsNodes(processing map[string]chan int, ssResultChan chan stateResul
...
+	ctr, err := docker.NewContainer(&opts, false, 10*time.Second, nil, nil)
+	if err != nil {
+		logger.WithError(err).Error("Could not create container")
+		return nil, err
+	}
+	logger.WithField("containerid", ctr.ID).Debug("Created a new container")
+	return ctr, nil
+}
+
+func addBindingToMap(bindsMap *map[string]string, cp, rp string) {
```
smousa: a couple of things:
1. maps are already pointers, so you don't need to pass a pointer to a pointer
2. this may make more sense as a type rather than a function.

node/container.go
```diff
@@ -133,27 +123,26 @@ func (a *HostAgent) StartContainer(cancel <-chan interface{}, svc *service.Servi
...
+		logger.WithError(err).Debug("Unable to get service")
 		return nil, nil, err
 	}
 
-	// create the container
-	opts := dockerclient.CreateContainerOptions{
-		Name:       fmt.Sprintf("%s-%d", svc.ID, instanceID),
-		Config:     conf,
-		HostConfig: hostConf,
+	// pull the service image
```
smousa: Why was this moved?

node/container.go
```diff
@@ -133,27 +123,26 @@ func (a *HostAgent) StartContainer(cancel <-chan interface{}, svc *service.Servi
 	}
 	defer masterClient.Close()
 
-	// get the container configs
-	conf, hostConf, err := a.setupContainer(masterClient, svc, instanceID, imageName)
+	svc, err := a.getService(serviceID)
```
smousa: You are making a connection to the client twice: is that necessary

node/agent.go
```diff
@@ -262,19 +261,21 @@ func manageTransparentProxy(endpoint *service.ServiceEndpoint, addressConfig *ad
 // setupContainer creates and populates two structures, a docker client Config and a docker client HostConfig structure
 // that are used to create and start a container respectively. The information used to populate the structures is pulled from
 // the service, serviceState, and conn values that are passed into setupContainer.
-func (a *HostAgent) setupContainer(masterClient master.ClientInterface, svc *service.Service, instanceID int, imageName string) (*dockerclient.Config, *dockerclient.HostConfig, error) {
+func (a *HostAgent) setupContainer(masterClient master.ClientInterface, svc *service.Service, instanceID int, imageUUID string, imageName string) (*docker.Container, *zkservice.ServiceState, error) {
```
smousa: perhaps, don't need to pass the masterClient, just the vars that the master client needs
tenantID and systemuser

node/container.go
smousa: should setup container not make calls to the masterClient?

zzk/service/hoststate.go
```diff
@@ -157,25 +157,16 @@ func (l *HostStateListener) Spawn(shutdown <-chan interface{}, stateID string) {
 
 		// load the service state node
 		sspth := path.Join("/services", serviceID, stateID)
-		ssdat := &ServiceState{}
-		if err := l.conn.Get(sspth, ssdat); err == client.ErrNoNode {
-
+		exists, err := l.conn.Exists(sspth)
```
smousa: You do need to call Get on ServiceState on first start so that it knows if it is attaching to an existing container.

smousa: But this can be called outside of the for loop

zzk/service/service.go
```diff
@@ -35,11 +36,40 @@ func (err ServiceError) Error() string {
 
 // ServiceNode is the storage object for service data
 type ServiceNode struct {
-	*service.Service
+	//*service.Service
```
smousa: delete this

zzk/service/service.go
```diff
@@ -35,11 +36,40 @@ func (err ServiceError) Error() string {
...
+		ID:            s.ID,
+		Name:          s.Name,
+		DesiredState:  s.DesiredState,
+		Instances:     s.Instances,
+		RAMCommitment: s.RAMCommitment,
+		ChangeOptions: s.ChangeOptions,
+	}
+}
+
+func (s *ServiceNode) AsService() *service.Service {
```
smousa: Where does this get called?

zzk/service/servicestate.go
```diff
@@ -127,7 +127,7 @@ func (l *ServiceListener) Spawn(shutdown <-chan interface{}, serviceID string) {
 			}
 
 			// Synchronize the number of service states
-			if _, ok := l.Sync(sDat.Locked, sDat.Service, reqs); !ok {
+			if _, ok := l.Sync(sDat.Locked, sDat.AsService(), reqs); !ok {
```
smousa: Don't use AsService here

node/agent.go
```diff
@@ -309,6 +310,14 @@ func (a *HostAgent) setupContainer(masterClient master.ClientInterface, svc *ser
 	cfg.ExposedPorts = make(map[dockerclient.Port]struct{})
 	hcfg.PortBindings = make(map[dockerclient.Port][]dockerclient.PortBinding)
 
+	state := &zkservice.ServiceState{
+		ImageID: imageUUID,
```
morr-zenoss: Done.

zzk/service/servicestate.go
morr-zenoss: Changed it to pass sdat, and modified subsequent calls to handle a ServiceNode. in  scheduler/leader.go, added a call to retrieve the full service from elastic. Per Glenn (and me), this change definitely needs a review from you for correctness.

node/agent.go
morr-zenoss: setupContainer has been moved to container.go

params for setupContainer have changed - we no longer pass masterClient.

zzk/service/service.go
morr-zenoss: Deleted.

zzk/service/service.go
morr-zenoss: Nowhere, after changes - removed.

node/agent.go
morr-zenoss: 1. fixed.
2. ?? I'm confused - what do you mean?

node/container.go
morr-zenoss: This has been refactored.

scheduler/leader.go
```diff
@@ -70,12 +70,18 @@ func Lead(shutdown <-chan interface{}, conn coordclient.Connection, cpClient dao
 // service has an address assignment the host will already be selected. If not
 // the host with the least amount of memory committed to running containers will
 // be chosen.  Returns the hostid, hostip (if it has an address assignment).
-func (l *leader) SelectHost(s *service.Service) (string, error) {
+func (l *leader) SelectHost(sn *zkservice.ServiceNode) (string, error) {
```
smousa: I think you should change StrategySelectHost to accept the ServiceNode and ServiceNode should include the fields necessary to schedule the service.

zzk/service/hoststate.go
morr-zenoss: I tried that, and it broke the code. Glenn and I backed this out and restored the Get. I'll be happy to revisit the change, but I'll need your help.

zzk/service/hoststate.go
```diff
@@ -157,17 +159,14 @@ func (l *HostStateListener) Spawn(shutdown <-chan interface{}, stateID string) {
 
 		// load the service state node
 		sspth := path.Join("/services", serviceID, stateID)
-		exists, err := l.conn.Exists(sspth)
-		if !exists {
-			logger.Debug("Service state was removed, exiting")
+		ssdat := &ServiceState{}
```
smousa: This needs to be loaded outside of the for loop, since this method is the only place where this node is ever modified

zzk/service/hoststate_test.go
```diff
@@ -33,27 +33,23 @@ var (
...
+const (
+	servicePath string = "/services/serviceid"
+	hostPath    string = "/hosts/hostid"
+	hostId      string = "hostid"
+	serviceId   string = "serviceid"
+	serviceName string = "serviceA"
+	imageId     string = "imageid"
+	containerId string = "containerid"
+)
+
```
smousa: This is kind of a nitpick, but I don't like that you set these as constants, because there may be tests in the future that may require different values for these attributes

zzk/service/hoststate_test.go
```diff
@@ -33,27 +33,23 @@ var (
...
-		Name:   "serviceA",
-		PoolID: "poolid",
-	}
-	spth := "/services/serviceid"
-	sdat := &ServiceNode{Service: svc}
-	err = conn.Create(spth, sdat)
-	c.Assert(err, IsNil)
-	err = conn.CreateDir("/hosts/hostid")
-	c.Assert(err, IsNil)
+	conn := setUpServiceAndHostPaths(c)
```
smousa: though this is a convenience, I have seen these convenience functions make these tests harder to read overall as the file gets bigger

node/agent.go
smousa: type bindsMap map[string]string

scheduler/leader.go
morr-zenoss: I could add the fields to ServiceNode, but we'd still need to get the service to populate the Endpoints for use here: https://github.com/control-center/serviced/blob/e290a689473e028c9e78d7b06e6dc5d81f5e717f/scheduler/leader.go#L97-L108

zzk/service/hoststate_test.go
morr-zenoss: I disagree. Glenn actually recommended putting the strings in constants, and I agree with his recommendation. There's already one test case that requires a different value, and I just used a different constant there. Using constants makes it clear where values are supposed to be the same (rather than just coincidentally agreeing), and prevents weird failures due to typos. Also, if we want to change the test values for some reason, we can simply change the value in one place, rather than having to pick through the source code deciding whether each instance of a string should be changed or not.

zzk/service/hoststate_test.go
morr-zenoss: I made this change not as a convenience, but to improve readability. This change reduces the amount of code that the user has to skip past to get to the interesting part of the test. It also reduces the number of places we'll need to change if we need to modify the test setup. Additionally, this change reduced the size of each test function by several lines (as mentioned above, that improves readability), and reduced the overall file size by ~120 lines.

node/container.go
```diff
@@ -420,14 +430,40 @@ func dockerLogsToFile(containerid string, numlines int) {
...
-func (a *HostAgent) setupContainer(tenantID string, svc *service.Service, instanceID int, systemUser user.User) (*dockerclient.Config, *dockerclient.HostConfig, error) {
+//ctr, state, err := a.setupContainer(tenantID, evaluatedService, instanceID, systemUser, imageUUID)
+func (a *HostAgent) setupContainer(tenantID string, svc *service.Service, instanceID int, systemUser user.User, imageUUID string) (*docker.Container, *zkservice.ServiceState, error) {
 	logger := plog.WithFields(log.Fields{
-		"tenantID": tenantID,
+		"tenantID":    tenantID,
 		"serviceName": svc.Name,
-		"serviceID": svc.ID,
-		"instanceID": instanceID,
+		"serviceID":   svc.ID,
```
morr-zenoss: done.

node/container.go
```diff
@@ -524,8 +587,8 @@ func (a *HostAgent) setupContainer(tenantID string, svc *service.Service, instan
 			}
 			logger.WithFields(log.Fields{
 				"requestedImage": requestedImage,
-				"hostPath": hostPath,
-				"containerPath": containerPath,
+				"hostPath":       hostPath,
+				"containerPath":  containerPath,
```
morr-zenoss: Done.

scheduler/leader.go
smousa: Can the ServiceNode object have a field named assigned IP that gets calculated when you call NewServiceNode?

zzk/service/hoststate_test.go
smousa: So then we have to create new constants everytime we need to create new variables?  I don't agree that is a good practice.  Alternatively, I would be okay with declaring the constants at the beginning of the functions.

zzk/service/hoststate_test.go
smousa: The thing that comes to mind for me is the facade/service* tests.  I find that those difficult to follow, and they follow the same pattern of using those consolidated calls.

facade/instance.go
```diff
@@ -222,7 +222,7 @@ func (f *Facade) getInstance(ctx datastore.Context, hst host.Host, svc service.S
 		imgLogger.Debug("Loaded service image from registry")
 
 		return imageData.UUID == imageUUID, nil
-	}(svc.ImageID, state.ImageID)
+	}(svc.ImageID, state.ImageUUID)
```
smousa: Another nitpick, but can we instead name this image hash instead of imageUUID?

node/container.go
```diff
@@ -722,3 +719,33 @@ func (a *HostAgent) createContainerConfig(tenantID string, svc *service.Service,
...
 }
+
+func (a *HostAgent) createContainer(conf *dockerclient.Config, hostConf *dockerclient.HostConfig, svcID string, instanceID int) (*docker.Container, error) {
+	logger := plog.WithFields(log.Fields{
+		"serviceid":  svcID,
+		"instanceid": instanceID,
+	})
+
+	if hostConf == nil {
+		logger.Error("Host Config passed to createContainer is nil.")
```
smousa: Does this need to be an error?

smousa: And should it return an error?

node/container.go
```diff
@@ -123,29 +124,38 @@ func (a *HostAgent) StartContainer(cancel <-chan interface{}, serviceID string,
...
+	}
+	logger.WithField("systemUser", systemUser.Name).Debug("Got system user")
 
-	// set up the container
-	ctr, state, err := a.setupContainer(masterClient, svc, instanceID, imageUUID, imageName)
+	// get the container configs
+	ctr, state, err := a.setupContainer(tenantID, evaluatedService, instanceID, systemUser, imageUUID)
 	if err != nil {
-		logger.WithError(err).Debug("Could not setup container")
+		logger.WithError(err).Debug("Unable to get service")
```
morr-zenoss: When I changed it, we were getting the service, not calling setup container. I think the mismatch is a leftover from merging with Glenn's code last night. I've put it back to the original, so it once again matches the surrounding code.

node/container.go
```diff
@@ -452,8 +495,35 @@ func (a *HostAgent) setupContainer(tenantID string, svc *service.Service, instan
...
+					PortNumber:     endpoint.PortNumber,
+					PortTemplate:   endpoint.PortTemplate,
+					VirtualAddress: endpoint.VirtualAddress,
+				})
 			}
 		}
+		state.AssignedIP = assignedIP
+		state.Static = static
+	} else {
+		log.Warn("svc.endpoints was nil.")
```
morr-zenoss: Removed else block.

zzk/service/hoststate_test.go
morr-zenoss: I originally had the constants at the beginning of the functions, but then I had the same set of constants at the beginning of each. I promoted them to the global scope when I realized they were common to the whole file.

facade/instance.go
morr-zenoss: That change was in response to a request from Glenn in an earlier review:
tgjonestx 7 days ago
To avoid more confusion in the future, let's rename zkservice.ServiceState.ImageID to zkservice.ServiceState.ImageUUID

smousa: I think ImageHash is a better name for it

node/container.go
morr-zenoss: That's a good question. 

Looking at the code, I also have a redundant version of the same check about 9 lines later (obvious now that we've factored out CreateContainerOptions). 

In the code as it is currently written, there is only one path to createContainer, and it ensures that  hostConf will be non-nil.  So, that's an argument for removing the check altogether. 

On the other hand, if future code should call the function with a nil hostConf, we'll get a panic here:  https://github.com/control-center/serviced/blob/develop/commons/docker/api.go#L108
That's an argument for leaving the check in and making it an error (and possibly returning an error). 

I guess my preference would be eliminate the redundant check, and make the first one an error (returning an error). 

I am not particularly opinionated about this, so if you have a different preference, I'll be glad to implement that.

node/agent.go
morr-zenoss: We could declare the type, but I think we'd still need the function - I was pulling out a multi-line repeated code block and making it a one-liner. If we made it a type, we'd still need a function to do the trimming, checking, and logging that was in the original code block. Were you suggesting that we make the function a method of the bindsMap type?

I've changed the param to a regular map (eliminated the pointer), and moved the function into container.go, since it belongs with createContainerConfig.

zzk/service/hoststate.go
morr-zenoss: Fixed, thanks to your help!

scheduler/leader.go
morr-zenoss: I've added the appropriate fields (CPUCommitment, HostPolicy, and AddressAssignment) to ServiceNode. 

After looking at the code some more, I modified the StrategyService struct to have a zkservice.ServiceNode, rather than a Service object. Naming the fields properly made its implementation of the strategy.ServiceConfig interface work. Since it was only being used in StrategySelectHost, it's a safe change.

I also made the change you suggested of moving the code that gets the AddressAssignment from Service.Endpoints into NewServiceNodeFromService() (in zzk/service/service.go)

zzk/service/hoststate.go
morr-zenoss: Fixed, thanks to your help this afternoon!

zzk/service/hoststate_test.go
morr-zenoss: moved the setUpServiceAndHostPaths() function to the top of the file, as we discussed.

PR #2973, opened 2016-09-15
CC-2630: Add HasChildren to ServiceDetails

cchristiezen:
Adding a property to the service details which specifies if the service has children.

### DIFF
--- a/domain/service/servicedetails_store.go
+++ b/domain/service/servicedetails_store.go
@@ -142,7 +142,7 @@ func (s *storeImpl) hasChildren(ctx datastore.Context, serviceID string) (bool, 
 		"query": map[string]interface{}{
 			"term": map[string]string{"ParentServiceID": serviceID},
 		},
-		"fields": []string{"ID", "ParentServiceID"},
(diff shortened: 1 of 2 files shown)


domain/service/servicedetails_store.go
```diff
@@ -108,12 +124,37 @@ func (s *storeImpl) GetServiceDetailsByParentID(ctx datastore.Context, parentID
...
 
 	return details, nil
 }
 
+func (s *storeImpl) hasChildren(ctx datastore.Context, serviceID string) (bool, error) {
+	searchRequest := newServiceDetailsElasticRequest(map[string]interface{}{
+		"query": map[string]interface{}{
+			"term": map[string]string{"ParentServiceID": serviceID},
+		},
+		"fields": []string{"ID", "ParentServiceID"},
```
smousa: Probably don't need to return the ParentServiceID field

smousa: Probably don't need ParentServiceID field

cchristiezen: Done

PR #2975, opened 2016-09-15
Add host authentication framework

iancmcc:
**Please do not merge this yet.**

This PR adds key-based host authentication to serviced. When a host is added to the system on the master, the master generates an RSA key pair and gives it to the operator. The operator must then go to the host that was added and use `serviced host register KEYFILE` to register that host using that key pair. 

From then on, the host will automatically request an authentication token from the master every hour (configurable). That token will be needed to make mux connections, RPC requests, and REST requests. The token is signed by the master and contains information about the host's permissions, as well as a public key specific to that host that can be used to verify the request itself.


auth/localkeys.go
```diff
@@ -0,0 +1,288 @@
...
+	dKeyCond.L.Lock()
+	defer dKeyCond.L.Unlock()
+	for delegateKeys.localPrivate == nil || delegateKeys.masterPublic == nil {
+		dKeyCond.Wait()
+	}
+}
+
+// WatchDelegateKeyFile watches the delegate key file on the filesystem and
+// updates the internal delegate keys when changes are detected.
+func WatchDelegateKeyFile(filename string, cancel chan interface{}) error {
```
smousa: this cancel channel needs to be <-chan struct{}

auth/localkeys.go
```diff
@@ -0,0 +1,288 @@
...
+	if err != nil {
+		return err
+	}
+	defer w.Close()
+	cleanDir, _ := filepath.Split(filename)
+	ops := fsnotify.Write | fsnotify.Create
+	if err := w.Add(cleanDir); err != nil {
+		return err
+	}
+	for {
```
smousa: Does this have to be synchronous?  Can this go into a goroutine, since there is no error when you reach this point.

auth/token.go
```diff
@@ -0,0 +1,104 @@
...
+	currentToken string
+	expiration   time.Time
+	cond         = &sync.Cond{L: &sync.Mutex{}}
+)
+
+func now() time.Time {
+	return time.Now().UTC()
+}
+
+func expired() bool {
```
smousa: Please comment what how this function works

auth/token.go
```diff
@@ -0,0 +1,104 @@
...
+		refresh := expiration - time.Duration(1*time.Minute)
+		select {
+		case <-done:
+			return
+		case <-time.After(refresh):
+		}
+	}
+}
+
+func updateToken(token string, expires int64) {
```
smousa: either put all the private functions above or below

auth/token.go
```diff
@@ -0,0 +1,104 @@
...
+func AuthToken() string {
+	cond.L.Lock()
+	defer cond.L.Unlock()
+	for expired() {
+		cond.Wait()
+	}
+	return currentToken
+}
+
+func TokenLoop(f TokenFunc, done chan interface{}) {
```
smousa: done should be <-chan struct{}

smousa: Add comment on what this does

cli/api/daemon.go
```diff
@@ -604,6 +618,34 @@ func (d *daemon) startAgent() error {
...
+			defer masterClient.Close()
+			token, expires, err := masterClient.AuthenticateHost(myHostID)
+			if err != nil {
+				return "", 0, err
+			}
+			return token, expires, nil
+		}
+
+		// Start authenticating
+		go auth.TokenLoop(getToken, d.shutdown)
```
smousa: is it necessary for this to be in another goroutine?

cli/api/options.go
```diff
@@ -274,6 +275,7 @@ func GetDefaultOptions(config utils.ConfigReader) Options {
 		StorageStatsUpdateInterval: config.IntVal("STORAGE_STATS_UPDATE_INTERVAL", 300),
 		SnapshotSpacePercent:       config.IntVal("SNAPSHOT_USE_PERCENT", 20),
 		ZKSessionTimeout:           config.IntVal("ZK_SESSION_TIMEOUT", 15),
+		TokenExpiration:            config.IntVal("AUTH_TOKEN_EXPIRATION", 24*60*60),
```
smousa: Change to 1 hour

container/proxy.go
```diff
@@ -263,7 +264,12 @@ func (p *proxy) prxy(local net.Conn, address addressTuple) {
 			glog.Errorf("Container address is invalid. Can't create proxy: %s", address.containerAddr)
 			return
 		}
-		remote.Write(muxHeader)
+		muxAuthHeader, err := auth.BuildMuxHeader(muxHeader)
```
smousa: need keys and tokens from the host?

pkg/serviced.default
```diff
@@ -192,3 +192,6 @@
 
 # Timeout in seconds for a zookeeper connection to timeout.  Defaults to 15.
 # SERVICED_ZK_SESSION_TIMEOUT=15
+
+# Expiration time in seconds for delegate authentication tokens.  Defaults to 1 day.
+# SERVICED_AUTH_TOKEN_EXPIRATION=86400
```
smousa: Should this default to 1 hour?

proxy/mux.go
```diff
@@ -141,13 +142,16 @@ func (mux *TCPMux) muxConnection(conn net.Conn) {
...
-	muxHeader := make([]byte, 6)
-	if _, err := conn.Read(muxHeader); err != nil {
+	// Read in the authenticated mux header
+	authMuxHeader, err := auth.ReadMuxHeader(conn)
+	if err != nil {
 		log.WithError(err).Warn("Unable to read valid mux header. Closing connection")
 		conn.Close()
 		return
 	}
+	// TODO retrieve and validate the identity of the sender
```
smousa: TODO?

rpc/master/interface.go
```diff
@@ -47,8 +47,8 @@ type ClientInterface interface {
 	// GetActiveHosts returns all active host ids or empty array
 	GetActiveHostIDs() ([]string, error)
 
-	// AddHost adds a Host
-	AddHost(h host.Host) error
+	// AddHost adds a Host; returns
```
smousa: returns...?

auth/localkeys.go
iancmcc: We discussed this; it's `interface{}` because the shutdown channel in daemon is `interface{}`

auth/localkeys.go
iancmcc: It did when there, but this function has since been extracted into a utility and rewritten. That section is now in a goroutine, because that makes it callable easier.

cli/api/daemon.go
iancmcc: Yes.

proxy/mux.go
iancmcc: Yes, this will be accomplished in the authorization stage of this feature. The current PR is for the authentication framework; nothing is rejected yet.

auth/jwt.go
```diff
@@ -0,0 +1,131 @@
...
+	if now < id.IssuedAt {
+		return ErrIdentityTokenNotValidYet
+	}
+
+	return nil
+}
+
+func (id *jwtIdentity) Expired() bool {
+	now := jwt.TimeFunc().UTC().Unix()
+	return now >= id.ExpiresAt
```
smousa: can also do now.After(id.ExpiresAt)

container/controller.go
```diff
@@ -278,8 +278,12 @@ func NewController(options ControllerOptions) (*Controller, error) {
 	// Load keys
 	keyshutdown := make(chan interface{})
 	go func() {
-		<-c.closing
+		// This is a ridiculous shutdown pattern I'll follow for the sake of
+		// not destabilizing the controller this late in a release.
+		// TODO: Make this not stupid
+		errc := <-c.closing
```
smousa: this PR might step on this change: 
https://github.com/control-center/serviced/pull/2996

facade/service.go
```diff
@@ -738,7 +738,7 @@ func (f *Facade) GetService(ctx datastore.Context, id string) (*service.Service,
 // GetEvaluatedService returns a service where an evaluation has been executed against all templated properties.
 func (f *Facade) GetEvaluatedService(ctx datastore.Context, serviceID string, instanceID int) (*service.Service, error) {
 	logger := plog.WithFields(log.Fields{
-		"serviceID": serviceID,
+		"serviceID":  serviceID,
```
smousa: should these fields be all lower case?

PR #2976, opened 2016-09-15
after host is added, show user the new host private key

rDr4g0n:
After new host is added, show a dialog with the hosts private key and some basic instructions on how to use it. Also easy to copy/pasta the key on click.

image

### DIFF
--- a/web/ui/src/Hosts/display-host-keys.html
+++ b/web/ui/src/Hosts/display-host-keys.html
@@ -1,0 +1,9 @@
+<p>Click the text area to copy the keys for host <strong>{{name}}</strong></p>
+
+<div class="keys-wrap" style="position: relative; height: 200px;">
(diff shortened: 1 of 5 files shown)


web/ui/src/Hosts/display-host-keys.html
```diff
@@ -0,0 +1,9 @@
+<p>Click the text area to copy the keys for host <strong>{{name}}</strong></p>
+
+<div class="keys-wrap" style="position: relative; height: 200px;">
+    <textarea class="keys" readonly style="height: 100%; max-width: 100%; background-color: #DDD;">{{keys}}</textarea>
+</div><br>
+
+<p>
+    These keys must be registered with the host machine <strong>{{name}}</strong> using the <span class="inline-code">serviced host register</span> command. If the keys are lost, new keys must be generated for the host.
```
iancmcc: These keys must be registered ON the host machine.

PR #2980, opened 2016-09-16
Feature/rpc auth rebased

johnlhamilton:
DO NOT MERGE

For comments only

These changes require a delegate to be authenticated and have admin access to issue any RPC calls except for the "AuthenticateHost" call.  The master is an exception and can issue RPC calls even if it is not added as a host, as long as it has the master private key.

In the future, most delegates will not have admin access, so we will need to expand the list of "NonAdminRequiredCalls", which currently is an empty list.

### DIFF
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -21,7 +21,6 @@ import (
 )
 
 const (
-	ADDRESS_BYTES   = 6
 	TOKEN_LEN_BYTES = 4
 	SIGNATURE_BYTES = 256
 )
--- a/auth/auth.go
+++ b/auth/auth.go
@@ -14,9 +14,16 @@
 package auth
 
 import (
+	"encoding/binary"
 	"errors"
 
 	"github.com/control-center/serviced/logging"
+)
+
+const (
+	ADDRESS_BYTES   = 6
+	TOKEN_LEN_BYTES = 4
+	SIGNATURE_BYTES = 256
 )
 
 var (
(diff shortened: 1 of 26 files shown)


auth/auth.go
```diff
@@ -14,11 +14,18 @@
...
 
 import (
+	"encoding/binary"
 	"errors"
 
 	"github.com/control-center/serviced/logging"
 )
 
+const (
+	ADDRESS_BYTES   = 6
```
johnlhamilton: This should probably go back into mux.go, as it is only used there.

auth/rpc.go
```diff
@@ -32,17 +32,31 @@ var (
...
 
+	// get current host token
+	signAsMaster = false
+	token, err = AuthTokenNonBlocking()
 	if err != nil {
-		return nil, err
+		// We may be an un-added master
+		token, err2 = MasterToken()
+		if err2 != nil {
+			log.WithError(err2).Debug("Unable to retrieve master token")
```
johnlhamilton: I'll add an additional debug log above the if to log the error returned from trying to get the delegate token

auth/rpc.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	// Next tokeLen bytes contain the token
+	token := string(rawHeader[offset : offset+tokenLen])
+	offset += tokenLen
+
+	// Validate the token can be parsed
+	senderIdentity, err := ParseJWTIdentity(token)
+	if err != nil || senderIdentity == nil {
+		if err == nil || senderIdentity == nil {
```
johnlhamilton: Agreed.  I will clean it up

rpc/rpcutils/authcodec.go
```diff
@@ -0,0 +1,78 @@
...
+	return a.wrappedcodec.ReadResponseBody(body)
+}
+
+func (a AuthClientCodec) Close() error {
+	return a.wrappedcodec.Close()
+}
+
+// NewClient returns a new rpc.Client to handle requests to the
+// set of services at the other end of the connection.
+func NewAuthClient(wrappedCodec rpc.ClientCodec) *rpc.Client {
```
johnlhamilton: There is already a NewAuthClientCodec.  This method returns a client, that uses an AuthClientCodec.

PR #2982, opened 2016-09-19
Feature/cc 2492.9.fix

rphillipsz:


proxy/mux.go
```diff
@@ -143,7 +143,11 @@ func (mux *TCPMux) muxConnection(conn net.Conn) {
 
 	// Read in the mux header (a 6-byte representation of a TCP address)
 	muxHeader := make([]byte, 6)
-	if _, err := conn.Read(muxHeader); err != nil {
+	if _, err := conn.Read(muxHeader); err == io.EOF {
+		log.Debug("Detected closed connection in mux.")
```
smousa: Is this strictly for purposes of logging?

web/utils.go
```diff
@@ -75,7 +75,7 @@ func GetCertFiles(certFile, keyFile string) (string, string) {
 
 // GetRemoteConnection returns a connection to a remote address
 func GetRemoteConnection(useTLS bool, export *registry.ExportDetails) (remote net.Conn, err error) {
-	isLocalAddress := IsLocalAddress(export.PrivateIP)
+	isLocalAddress := IsLocalAddress(export.HostIP)
```
smousa: Instead of making a call of IsLocalAddress, is there a way to compare the hostIDs?

proxy/mux.go
smousa: Nevermind, I answered my own question

rphillipsz: With a busy system with many clients, the log will get a continuous stream of:
WARN[0090] Unable to read valid mux header. Closing connection  address=[::]:22250 error=EOF location=mux.go:157 logger=proxy remoteaddr=x.x.x.x:xxxx

web/utils.go
rphillipsz: That's a good question. I don't know the answer.  The existing check is cheap though, since it already caches the list of local ips in init()

PR #2985, opened 2016-09-20
Use json.Number for offsets in multi-line logs. Add the fields prefix…

youngklee:
… to logstash fields

As in the logSingleLine struct, use an array of json.Number for Offsets.

Because of the logstash upgrade, each field is prefixed with "fields". This fix the issue where serviced log export won't export logs when a service is specified.


youngklee:
retest

smousa:
looking

PR #2955, opened 2016-09-09
Use a new RPC to get evaluated Service objects

tgjonestx:
Fixes CC-2591

Replace recursive RPC calls for GetService from delegate to master with a single GetEvaluatedService RPC which runs the template evaluations on the master.

PR #2986, opened 2016-09-20
fix a bug introduced by PR #2955

tgjonestx:
Fixes ZEN-24800

The bug was due to the fact that instance Service used to setup exported endpoint data was NOT the instance of Service after all of the template evaluations were done.  The first commit is simply moving code from one file to another. The second commit has the actual bug fix.


node/container.go
```diff
@@ -110,16 +115,18 @@ func (a *HostAgent) AttachContainer(state *zkservice.ServiceState, serviceID str
 // StartContainer creates a new container and starts.  It returns info about
 // the container, and an event monitor to track the running state of the
 // service.
-func (a *HostAgent) StartContainer(cancel <-chan interface{}, svc *service.Service, instanceID int) (*zkservice.ServiceState, <-chan time.Time, error) {
+// FIXME - when integrated with changes for CC-2590, the partialSvc parameter should be replaced with the handful of
+//         props we really need; e.g. ID & ImageID
+func (a *HostAgent) StartContainer(cancel <-chan interface{}, partialSvc *service.Service, instanceID int) (*zkservice.ServiceState, <-chan time.Time, error) {
```
smousa: why bother renaming the variable?

smousa: I think start container should only need to pass in the cancel, service id, and instance id

node/container.go
```diff
@@ -133,16 +140,44 @@ func (a *HostAgent) StartContainer(cancel <-chan interface{}, svc *service.Servi
 	}
 	defer masterClient.Close()
 
+	evaluatedService, err := masterClient.GetEvaluatedService(partialSvc.ID, instanceID)
```
smousa: This should be called before pull image

smousa: Get evaluated service should return the tenant id and the service

node/container.go
```diff
@@ -133,16 +140,44 @@ func (a *HostAgent) StartContainer(cancel <-chan interface{}, svc *service.Servi
...
+	if err != nil {
+		logger.WithError(err).Error("Failed to get service")
+		return nil, nil, err
+	}
+	// Update the service with the complete image name
+	evaluatedService.ImageID = imageName
```
smousa: why are you setting this value?

tgjonestx: We need to value of `imageName` to setup the container properly.  Things shuffled around some in the changeset 41a0386

node/container.go
```diff
@@ -110,39 +115,60 @@ func (a *HostAgent) AttachContainer(state *zkservice.ServiceState, serviceID str
...
 	if err != nil {
 		logger.WithError(err).Debug("Could not pull the service image")
 		return nil, nil, err
 	}
+	// Update the service with the complete image name
+	evaluatedService.ImageID = imageName
 
-	// Establish a connection to the master
-	masterClient, err := master.NewClient(a.master)
+	if len(tenantID) == 0 && len(evaluatedService.Volumes) > 0 {
```
smousa: all services must have a tenant id regardless

smousa: I don't think evaluated service should return if the tenant id is not found

rpc/master/service_server.go
```diff
@@ -65,12 +70,18 @@ func (s *Server) GetService(serviceID string, svc *service.Service) error {
...
 // GetEvaluatedService returns a service where an evaluation has been executed against all templated properties.
-func (s *Server) GetEvaluatedService(request EvaluateServiceRequest, svc *service.Service) error {
-	result, err := s.f.GetEvaluatedService(s.context(), request.ServiceID, request.InstanceID)
+func (s *Server) GetEvaluatedService(request EvaluateServiceRequest, response *EvaluateServiceResponse) error {
+	svc, err := s.f.GetEvaluatedService(s.context(), request.ServiceID, request.InstanceID)
+	if err != nil {
+		return err
+	}
+
+	tenantID, err := s.f.GetTenantID(s.context(), request.ServiceID)
```
smousa: I feel like this should be called in the facade, instead of in addition to the facade

tgjonestx: I disagree. I think the facade should provide discrete chunks of business logic. This RPC layer is combining results from one or more or those facade calls to return to a client.  I guess I don't see how this is any different from a variety of REST calls to where the web tier makes more than one call to the facade

PR #2990, opened 2016-09-20
Store identity in memory

fraibacas:


auth/auth.go
```diff
@@ -60,14 +60,18 @@ type Verifier interface {
 	Verify(message []byte, signature []byte) error
 }
 
+type DelegateIdentity interface {
```
iancmcc: I would change this to `Delegate`, move `Verifier()` to it, make a new interface called `Expires` with `Expired() bool` on it, then make `Identity` embed `Delegate`, `Expires` and `jwt.Claims`

auth/token.go
```diff
@@ -95,10 +96,27 @@ func TokenLoop(f TokenFunc, done chan interface{}) {
...
+		return nil
+	}
+	return identity
+}
+
 func updateToken(token string, expires int64) {
 	cond.L.Lock()
 	currentToken = token
+	currentIdentity = delegateIdentityFromToken(token)
 	expiration = time.Unix(expires, 0)
```
iancmcc: I think the global `expiration` should be replaced by referencing the current token's expiration.

auth/auth.go
iancmcc: Or, wait; why make this a separate interface?

PR #2991, opened 2016-09-20
Fix check for local connections so we don't mux locals

rphillipsz:


web/utils.go
```diff
@@ -73,23 +73,61 @@ func GetCertFiles(certFile, keyFile string) (string, string) {
...
-	if useTLS {
-		config := tls.Config{InsecureSkipVerify: true}
-		remote, err = tls.Dial("tcp4", remoteAddress, &config)
-	} else {
-		remote, err = net.Dial("tcp4", remoteAddress)
+	remote, err = dialer.Dial("tcp4", remoteAddress)
+
+	// Prevent a panic if we couldn't connect to the mux.
+	if err != nil {
+		return nil, err
```
tgjonestx: Seems like this method could be simplified a little to something like:
```
    var address string
    isLocalAddress := IsLocalAddress(export.HostIP)
    if isLocalAddress {
		// if the address is local return the connection
  		address = fmt.Sprintf("%s:%d", export.PrivateIP, export.PortNumber)
    } else {
 	    // set up the remote address
  	    address = fmt.Sprintf("%s:%d", export.HostIP, export.MuxPort)
    }
    return dialer.Dial("tcp4", address)
```

web/utils_test.go
```diff
@@ -0,0 +1,96 @@
...
+	return serviceAddress
+}
+
+type dialedMuxAddress struct{}
+func (e *dialedMuxAddress) Error() string {
+	return muxAddress
+}
+
+// Mock dialers that just return an error on dial based on the address given.
+type mockDialer struct{}
```
tgjonestx: This is all a little brute force. Here's how to replace this impl with the GO mocking lib we use:
```
import (
	"net"
	"github.com/stretchr/testify/mock"
)

type mockDialer struct {
	mock.Mock
}

func (_m *mockDialer) Dial(network, address string) (net.Conn, error)  {
	ret := _m.Called(d)

	var r0 net.Conn
	if rf, ok := ret.Get(0).(func(string, string) net.Conn); ok {
		r0 = rf(network, address)
	} else {
		if ret.Get(0) != nil {
			r0 = ret.Get(0).(net.Conn)
		}
	}

	var r1 error
	if rf, ok := ret.Get(1).(func(string, string) error); ok {
		r1 = rf(network, address)
	} else {
		r1 = ret.Error(1)
	}

	return r0, r1
}
```

web/utils_test.go
```diff
@@ -0,0 +1,96 @@
...
+	// Create an export to test with.
+	export := getMockExport()
+
+	_, err := getRemoteConnection(&export, &mockDialer{});
+	if err.Error() == serviceAddress {
+		c.Errorf("getRemoteConnection used a direct connection for a remote address")
+	}
+}
+
+func getMockExport() registry.ExportDetails {
```
tgjonestx: I wouldn't use the term 'mock' here if you're going to use the stretchr mock framework.  
It returns a valid instance of  `registry.ExportDetails`

web/utils_test.go
```diff
@@ -0,0 +1,96 @@
...
+	// Make sure localhost is in our ip map. hostIp isn't a real
+	// ip, so adding/removing it from the map won't have any effect
+	// on other tests.
+	ipmap[hostIp] = struct{}{}
+	defer delete(ipmap, hostIp)
+
+	// Create an export to test with.
+	export := getMockExport()
+
+	_, err := getRemoteConnection(&export, &mockDialer{});
```
tgjonestx: Given the mock I suggested above, then I think you can do something like this here:
```
var unusedConnection net.Conn
dialer := &mockDialer{}
dialer.On("Dial", "tcp4", serviceAddress).Return(unusedConnection, nil)

_, err := getRemoteConnection(&export, dialer);

c.Assert(err, IsNil)
dialer.AssertExpectations(c)
```

web/utils.go
rphillipsz: If the mux connection fails, it returns to prevent remote.Write() below from causing a panic.  We have to write the mux header on remote connections though?  Maybe I'm misreading the comment though.

smousa:
This fix is related to https://jira.zenoss.com/browse/CC-2680.  Here is the diff that I used to fix it:
```diff
diff --git a/web/reverseproxy.go b/web/reverseproxy.go
index 990aea0..52c73c8 100644
--- a/web/reverseproxy.go
+++ b/web/reverseproxy.go
@@ -73,7 +73,7 @@ func GetReverseProxy(useTLS bool, export *registry.ExportDetails) *httputil.Reve
 
        // Set the remote address based on whether the container is running on this
        // host.
-       if IsLocalAddress(export.PrivateIP) {
+       if IsLocalAddress(export.HostIP) {
                remoteAddress = fmt.Sprintf("%s:%d", export.PrivateIP, export.PortNumber)
        } else {
                remoteAddress = fmt.Sprintf("%s:%d", export.HostIP, export.MuxPort)
diff --git a/web/utils.go b/web/utils.go
index 1f9020a..dab1017 100644
--- a/web/utils.go
+++ b/web/utils.go
@@ -77,7 +77,7 @@ func GetCertFiles(certFile, keyFile string) (string, string) {
 
 // GetRemoteConnection returns a connection to a remote address
 func GetRemoteConnection(useTLS bool, export *registry.ExportDetails) (remote net.Conn, err error) {
-       isLocalAddress := IsLocalAddress(export.PrivateIP)
+       isLocalAddress := IsLocalAddress(export.HostIP)
 
        if isLocalAddress {
                // if the address is local return the connection
```

web/utils.go
```diff
@@ -141,7 +141,7 @@ func getRemoteConnection(export *registry.ExportDetails, dialer dialerInterface)
 	muxHeader, err = auth.BuildMuxHeader(muxHeader)
 	if err != nil {
 		glog.Errorf("Error building authenticated mux header. %s", err)
```
smousa: use logrus logger

rphillipsz: Yikes. this was from the develop merge.  I'll change it.

web/utils_test.go
```diff
@@ -0,0 +1,90 @@
...
+	conn   := &mocks.Conn{}
+	dialer := &mocks.Dialer{}
+	export := getExportDetails()
+	muxHeader, _ := utils.PackTCPAddress(export.PrivateIP, export.PortNumber)
+	
+	dialer.On("Dial", "tcp4", muxAddress).Return(conn, nil)
+	conn.On("Write", muxHeader).Return(0, nil)
+
+	fmt.Printf("dialer=%T\n", dialer)
+	fmt.Printf("dialer=%v\n", dialer)
```
smousa: Instead of Printf, you can use c.Logf

web/utils_test.go
```diff
@@ -0,0 +1,90 @@
...
+	defer delete(ipmap, hostIp)
+
+	var unusedConnection net.Conn
+	dialer := &mocks.Dialer{}
+	export := getExportDetails()
+
+	dialer.On("Dial", "tcp4", serviceAddress).Return(unusedConnection, nil)
+
+	fmt.Printf("dialer=%T\n", dialer)
+	fmt.Printf("dialer=%v\n", dialer)
```
smousa: Use c.Logf

web/utils_test.go
tgjonestx: Why have the Logf at all?

rphillipsz: those shouldn't be there at all.

PR #2992, opened 2016-09-20
Remove chatty warnings when browsers close muxed connections

rphillipsz:
There may be other issues in the mux causing this EOF error; I'm going to wait a bit before making any changes here.


smousa:
The vhost test is  consistently breaking with this PR

PR #2995, opened 2016-09-21
Remove SERVICED_AGENT setting altogether

spopezen:
https://jira.zenoss.com/browse/CC-2516

Simply removing SERVICED_AGENT from config file.

(Should this be removed from [serviced.service](https://github.com/control-center/serviced/blob/develop/pkg/serviced.service#L7) and [serviced.upstart](https://github.com/control-center/serviced/blob/develop/pkg/serviced.upstart#L22)?)
(Do we want to remove it from [options](https://github.com/control-center/serviced/blob/develop/cli/api/options.go#L228) entirely?)


spopezen:
Good call.  Looks like we should change default value to true at [options.go Line 228](https://github.com/control-center/serviced/blob/develop/cli/api/options.go#L228)

spopezen:
Do we want this to be configurable at all?

Would [this](https://github.com/control-center/serviced/blob/develop/cli/cmd/cmd.go#L280-L282) be better as...
Still allowing for configuration from env:
```
if options.Agent = true; os.Getenv("SERVICED_AGENT") == "0" {
    options.Agent = false
}
```
or
Making it always true:
`options.Agent = true`

spopezen:
retest this

PR #2999, opened 2016-09-22
Merge pull request #2791 from control-center/feature/CC-2327

smousa:
added zoo.cfg to isvcs resources path and bumped isvcs-zookeeper container version


smousa:
yes in v4

PR #3007, opened 2016-09-23
Fix CC-2686: panic in serviced host register

iancmcc:
A missing `return` in the code led to a correct printing of usage and then an incorrect panic when it went on to run the code anyway.

Demo:

```
[ian:~/metis/src/github.com/control-center/serviced] feature/CC-2686* 2m6s 2 ± ./serviced host register
Incorrect Usage.

NAME:
   register - Set the authentication keys to use for this host. When KEYSFILE is -, read from stdin.

USAGE:
   command register [command options] [arguments...]

DESCRIPTION:
   serviced host register KEYSFILE

OPTIONS:
   
```


iancmcc:
Sure, there you go

PR #3009, opened 2016-09-23
Add Permissions to ResourcePool

jafcrocker:


domain/pool/pool.go
```diff
@@ -30,6 +30,13 @@ type VirtualIP struct {
 	BindInterface string
 }
 
+type Permission uint
```
iancmcc: I think this type should be in a new file in the auth package.

iancmcc: Never mind.

PR #3010, opened 2016-09-23
Authenticate RPC calls

johnlhamilton:


rpc/rpcutils/authcodec.go
```diff
@@ -0,0 +1,240 @@
...
+// This implementation gets an auth header when appropriate, and writes it to the stream
+//  before letting the underlying codec send the rest of the request.
+func (a *AuthClientCodec) WriteRequest(r *rpc.Request, body interface{}) error {
+	var (
+		header []byte = []byte{}
+		err    error
+	)
+
+	if requiresAuthentication(r.ServiceMethod) {
+		header, err = a.headerBuilder.BuildHeader()
```
iancmcc: Yo, you gots to pass the request in there and sign it, or your header is replayable

auth/localkeys.go
```diff
@@ -33,8 +33,8 @@ const (
 var (
 	delegateKeys HostKeys
 	masterKeys   MasterKeys
-
-	dKeyCond = &sync.Cond{L: &sync.Mutex{}}
+	mKeyLock     sync.Mutex
```
iancmcc: This might not make a huge difference, but I recommend using a sync.RWMutex here. The writes are very infrequent, so you can usually RLock rather than Lock.

johnlhamilton:
retest this

PR #3014, opened 2016-09-26
Clean up tmpDir to avoid panic in unit tests

rphillipsz:
Failing unit tests.  It looks like the tmpDir from c.MkDir() is the same, but not cleaned up between calls.

See: http://jenkins.zendev.org/job/ControlCenter/job/develop/job/serviced-develop-pullrequests-unit/1853/console
 
Panic: Couldn't create temporary directory: mkdir /tmp/check-3784560248718450071: file exists

### DIFF
--- a/auth/localkeys_test.go
+++ b/auth/localkeys_test.go
@@ -26,6 +26,8 @@ import (
 func (s *TestAuthSuite) TestDumpLoadKeys(c *C) {
 	// We'll need a temp dir:
 	tmpDir := c.MkDir()
+	defer os.RemoveAll(tmpDir)
+
 	hostKeyFile := fmt.Sprintf("%s/host", tmpDir)
 	masterKeyFile := fmt.Sprintf("%s/master", tmpDir)
 
@@ -79,6 +81,8 @@ func (s *TestAuthSuite) TestCreateMasterKeys(c *C) {
 	auth.ClearKeys()
 	// We'll need a temp dir:
 	tmpDir := c.MkDir()
+	defer os.RemoveAll(tmpDir)
+
 	masterKeyFile := fmt.Sprintf("%s/.keys/master", tmpDir)
 	// hostKeyFile := fmt.Sprintf("%s/delegate", tmpDir)
 
--- a/auth/localkeys_test.go
+++ b/auth/localkeys_test.go
@@ -18,16 +18,19 @@ package auth_test
 import (
 	"fmt"
 	"os"
+	"math/rand"
+	"time"
 
 	"github.com/control-center/serviced/auth"
 	. "gopkg.in/check.v1"
 )
 
 func (s *TestAuthSuite) TestDumpLoadKeys(c *C) {
+	// seed rand for c.MkDir()
+	rand.Seed(time.Now().UnixNano())
 	// We'll need a temp dir:
 	tmpDir := c.MkDir()
 	defer os.RemoveAll(tmpDir)
-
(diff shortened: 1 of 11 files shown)


rphillipsz:
retest this

tgjonestx:
Should be deleting all directories created by `c.MkDir()`?

rphillipsz:
I think so. Here's the code behind it: https://github.com/go-check/check/blob/v1/check.go#L142, which creates a temp subfolder under the system temp directory.  anything you do in there is yours.  the problem comes when c.MkDir() generates the same random number later and a panic is thrown.  there were hundreds of these on the slave I checked, so I guess it's bad luck that will eventually run out as the number grows.

auth/localkeys_test.go
```diff
@@ -18,14 +18,19 @@ package auth_test
...
+	"math/rand"
+	"time"
 
 	"github.com/control-center/serviced/auth"
 	. "gopkg.in/check.v1"
 )
 
 func (s *TestAuthSuite) TestDumpLoadKeys(c *C) {
+	// seed rand for c.MkDir()
+	rand.Seed(time.Now().UnixNano())
```
smousa: Should this be in SetupSuite

dfs/backup_integration_test.go
```diff
@@ -50,7 +52,10 @@ var i int
 func (s *DFSTestSuite) BenchmarkBackup(c *C) {
 	//1012318498
 	// Create a file to play with
+	rand.Seed(time.Now().UnixNano())
```
smousa: This should be in setupsuite

volume/devicemapper/devicemapper_test.go
```diff
@@ -131,10 +131,17 @@ func (s *DeviceMapperSuite) TestDeviceMapperExcludeDirs(c *C) {
 
 	basesize1, err := units.RAMInBytes("10M")
 
-	exportdrv, err := Init(c.MkDir(), []string{fmt.Sprintf("dm.basesize=%d", basesize1)})
+	// seed rand for c.MkDir
+	rand.Seed(time.Now().UnixNano())
```
smousa: this should be in setupsuite

volume/devicemapper/devicemapper_test.go
```diff
@@ -184,8 +191,12 @@ func (s *DeviceMapperSuite) TestDeviceMapperExcludeDirs(c *C) {
 }
 
 func (s *DeviceMapperSuite) TestDeviceMapperImportBasesize(c *C) {
+	// seed rand for c.MkDir()
+	rand.Seed(time.Now().UnixNano())
```
smousa: this should be in setupsuite

volume/nfs/nfs_test.go
```diff
@@ -41,7 +43,11 @@ func (s *NFSSuite) TestNFSDriver(c *C) {
 		driver *NFSDriver
 	)
 
+	// seed rand for c.MkDir()
+	rand.Seed(time.Now().UnixNano())
```
smousa: this should be in setupsuite

volume/util_test.go
```diff
@@ -33,7 +34,9 @@ type UtilsSuite struct{}
 var _ = Suite(&UtilsSuite{})
 
 func (s *UtilsSuite) TestIsDir(c *C) {
+	rand.Seed(time.Now().UnixNano())
```
smousa: this should be in setupsuite

PR #3017, opened 2016-09-27
added option to service details to return the service ancestry

smousa:


domain/service/servicedetails.go
```diff
@@ -29,6 +29,7 @@ type ServiceDetails struct {
 	Description     string
 	PoolID          string
 	ParentServiceID string
+	Parent          *ServiceDetails `json:",omitempty"`
```
cchristiezen: Since the property is only used by one of the endpoints, it might be better to have another struct with the property (like ServiceDetailsResponse or anonymous in the web layer) and embed the ServiceDetails.

facade/interface.go
```diff
@@ -129,7 +129,7 @@ type FacadeInterface interface {
 
 	GetAllServiceDetails(ctx datastore.Context) ([]service.ServiceDetails, error)
 
-	GetServiceDetails(ctx datastore.Context, serviceID string) (*service.ServiceDetails, error)
+	GetServiceDetails(ctx datastore.Context, serviceID string, ancestors bool) (*service.ServiceDetails, error)
```
cchristiezen: Rather than have the bool parameter, have a GetServiceDetailsAncestors method that takes a service and returns the ancestors.

domain/service/servicedetails.go
smousa: I don't like that, because that implies we have to make 2 calls to get the data we need.  The purpose of this is so the function only needs to make one call.

smousa:
retest this

PR #3018, opened 2016-09-27
Fixes CC-2701: Remove restriction that import_all addresses must always go through the mux

iancmcc:


iancmcc:
While this is a good and proper change, it breaks something else that we can't fix without giving up on backwards compatibility. Thus, alas...

PR #3020, opened 2016-09-28
Add function which resets host private key to facade

jafcrocker:
* Added ResetHostKey
* Refactored code for generating host keys from AddHost into routine shared with ResetHostKey
* Added relevant unit tests

Contributes to [CC-2623](https://jira.zenoss.com/browse/CC-2623)


jafcrocker:
retest this

PR #3028, opened 2016-09-28
Feature/CC-2378

spopezen:
https://jira.zenoss.com/browse/CC-2378
Original issue reported out of date volume graph for CC ZenPack, root cause is;
The Control Center ZenPack reports stale volume information because dumpe2fs doesn't consistantly update for a mounted filesystem.

This PR fixes this by using df when the filesystem is mounted and dumpe2fs when it isn't.
A bit of extra work was required to get the "usable" blocks from dumpe2fs so the volume size stays consistent between dumpe2fs and df.

### DIFF
--- a/volume/devicemapper/stats.go
+++ b/volume/devicemapper/stats.go
@@ -208,9 +208,38 @@ func getDeviceSize(dev string) (uint64, error) {
 	return size, nil
 }
 
-// getFilesystemStats calls df to retrieve filesystem statistics.  If the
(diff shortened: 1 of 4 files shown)


spopezen:
Tested by making a request to https://localhost/storage on my local CC and observing `Status.Tenants` and comparing the values to a `df` and `dumpe2fs` for the volume.

smousa:
retest this

spopezen:
retest this

PR #3033, opened 2016-09-29
Cache service path with tenantID

tgjonestx:
The Facade tier already had code to cache tenantIDs for a given serviceID, but each time a service was scheduled, it was computing the service path for every config file on every service.  This proved to be very timing consuming with 100s of services.

This code replaces the tenantID cache with a cache of structs that contain tenantID and servicePath.  This change also required an additional call to remove cache entries when a service is removed or it's parent has changed.

Also, since we don't put the entire Service object in ZK any longer, I removed the call to fillServiceConfigs for any scheduleService operation.

### DIFF
--- a/facade/service.go
+++ b/facade/service.go
@@ -1099,7 +1099,7 @@ func (f *Facade) scheduleOneService(ctx datastore.Context, tenantID string, svc 
 		svc.DesiredState = int(desiredState)
 	}
 
-	cursvc, err := f.validateServiceUpdate(ctx, svc)
+	err := f.validateServiceStart(ctx, svc)
 	if err != nil {
 		glog.Errorf("Facade.scheduleService: Could not validate service %s (%s) for update: %s", svc.Name, svc.ID, err)
 		return err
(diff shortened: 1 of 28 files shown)


facade/service.go
```diff
@@ -237,11 +236,14 @@ func (f *Facade) updateService(ctx datastore.Context, tenantID string, svc servi
 		return err
 	}
 	glog.Infof("Updated service %s (%s)", svc.Name, svc.ID)
+
+	// FIXME - Do we really need to updateServiceConfigs everytime? can we skip this when called from schedulService?
```
smousa: yes

facade/service.go
```diff
@@ -1073,9 +1084,6 @@ func (f *Facade) scheduleService(ctx datastore.Context, tenantID, serviceID stri
 		default:
 			svc.DesiredState = int(desiredState)
 		}
-		if err := f.fillServiceConfigs(ctx, svc); err != nil {
```
smousa: You might not want to delete that

tgjonestx: Per our discussion, I'm going to replace the logic in this visitor function to schedule the service without calling `updateService()`. That way, I can leave service configs out of the equation entirely.

PR #3037, opened 2016-09-30
Cache service path with TenantID (#2)

tgjonestx:
This is a redo of #3033 because I bungled a merge/rebase in the branch for 3033.
----
The Facade tier already had code to cache tenantIDs for a given serviceID, but each time a service was scheduled, it was computing the service path for every config file on every service.  This proved to be very timing consuming with 100s of services.

This code replaces the tenantID cache with a cache of structs that contain tenantID and servicePath.  This change also required an additional call to remove cache entries when a service is removed or it's parent has changed.

Also, since we don't put the entire Service object in ZK any longer, I removed the call to fillServiceConfigs for any scheduleService operation.

### DIFF
--- a/facade/service.go
+++ b/facade/service.go
@@ -1061,7 +1061,7 @@ func (f *Facade) scheduleService(ctx datastore.Context, tenantID, serviceID stri
 			return 0, fmt.Errorf("desired state unknown")
 		}
 		if err := f.validateServiceSchedule(ctx, serviceID, autoLaunch); err != nil {
-			glog.Errorf("Facade.scheduleService: Could not validate service schedule for service %s: %s", serviceID, err)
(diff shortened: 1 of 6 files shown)


facade/service.go
```diff
@@ -1050,7 +1061,7 @@ func (f *Facade) scheduleService(ctx datastore.Context, tenantID, serviceID stri
 			return 0, fmt.Errorf("desired state unknown")
 		}
 		if err := f.validateServiceSchedule(ctx, serviceID, autoLaunch); err != nil {
-			glog.Errorf("Could not validate service schedule for service %s: %s", serviceID, err)
+			glog.Errorf("Facade.scheduleService: Could not validate service schedule for service %s: %s", serviceID, err)
```
smousa: do you need this extra logging here?

tgjonestx: Nope

PR #3038, opened 2016-09-30
Feature/facade metrics

morr-zenoss:
Add metrics code to project. An instance of Metrics is added to the datastore.Context object.
To enable metric collection and writing of metrics to the log, add the following lines to the function for which metric collection is desired (current implementation requires access to a Context object to get the metrics):
```go
ctx.Metrics().Enabled = true
defer ctx.Metrics().LogAndCleanup(ctx.Metrics().Start(<tag_string>))
```

`<tag_string>` is a string to define the metric - the method name is typical, but other strings may be used.

Timing data will be collected for that call and all instrumented child calls, and output to the log when the function returns.

To instrument a function for metrics collection, add the following to the top of the function (again, a context is required):
```go
defer ctx.Metrics().Stop(ctx.Metrics().Start(<tag_string>))
```

Because the Metrics object is a singleton on the Context object, and the Context is typically a singleton, all calls to methods using that context will go into the collected time. To facilitate isolation of metrics (for testing only, not production), we have added a GetNew() method to Context to enable getting a new instance using the configured driver. This will allow differerent contexts (and therefore,  different Metrics instances) to be passed to functions for testing.

Per a request from @tgjonestx , I've added a field, "GroupName" to the Metrics object. This field is not currently used (other than being output to the log with metrics data), but is available for future use to annotate metrics collection.

I've added a benchmark test to Metrics, which demonstrates that disabling the metrics (by not setting `Metrics.enabled` to true) is effective, as there is a ~100x time difference between a function with and without metrics collection. 

To run the benchmark test, you can cd into the metrics directory and run 
```bash
go test -bench=. -tags="unit"
```

### DIFF
--- a/metrics/metrics_test.go
+++ b/metrics/metrics_test.go
@@ -46,7 +46,6 @@ func BenchmarkLoggingOffCallingFunctionWithoutLogging(b *testing.B) {
 
 func testFunction(loggingEnabled bool, f func(m *Metrics) bool) {
 	m := NewMetrics()
-	defer m.Stop(m.Start("BenchmarkLogging"))
 	m.Enabled = loggingEnabled
 	defer m.LogAndCleanUp(m.Start("BenchmarkLogging"))
 	r := f(m) // record result to prevent compiler optimization of call to f. (See https://goo.gl/caaEOU)
--- a/metrics/metrics_test.go
+++ b/metrics/metrics_test.go
@@ -1,0 +1,65 @@
+// Copyright 2014 The Serviced Authors.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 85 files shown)


metrics/metrics_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+func BenchmarkLoggingOffCallingFunctionWithoutLogging(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		testFunction(false, functionWithoutLogging)
+	}
+}
+
+func testFunction(loggingEnabled bool, f func(m *Metrics) bool) {
+	m := NewMetrics()
+	defer m.Stop(m.Start("BenchmarkLogging"))
```
tgjonestx: Is this Start/Stop necessary given the code at line 51?

morr-zenoss: Maybe not.

morr-zenoss: I've removed the line, after confirming on my dev box that the test runs fine without it.

PR #3041, opened 2016-10-03
Refresh authentication token when delegate keys change.

iancmcc:
This also modifies the auth code to allow the functions that block waiting for a token or keys to be cancellable.

### DIFF
--- a/auth/token.go
+++ b/auth/token.go
@@ -115,7 +115,6 @@ func CurrentIdentity() Identity {
 // setting the result as the current live token, until the done channel is
 // closed.
 func TokenLoop(f TokenFunc, tokenfile string, done <-chan interface{}) {
-
 	for {
 		expires, err := RefreshToken(f, tokenfile)
 		if err != nil {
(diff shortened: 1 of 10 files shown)


auth/token.go
```diff
@@ -52,24 +52,33 @@ func RefreshToken(f TokenFunc, filename string) (int64, error) {
 		return 0, err
 	}
 	updateToken(token, time.Unix(expires, 0), filename)
-	log.WithField("expiration", expires).Debug("Received new authentication token")
+	log.WithField("expiration", expires).Info("Received new authentication token")
```
johnlhamilton: If we leave this as Info, it is probably going to show up as output on every CLI command.

makefile
```diff
@@ -160,7 +160,7 @@ FORCE:
 serviced: $(GO)
 serviced: FORCE
 	$(GO) build $(GOBUILD_FLAGS) ${LDFLAGS}
-	make govet
+	# make govet
```
jafcrocker: We need to run govet

auth/token.go
iancmcc: We need CLI commands not to show logging at all.

johnlhamilton: Opened CC-2719 to address this.

PR #3042, opened 2016-10-03
Feature/cc 2623/ui 2

rDr4g0n:
Add reset host keys button and modal

image

image

### DIFF
--- a/web/ui/src/Hosts/HostsController.js
+++ b/web/ui/src/Hosts/HostsController.js
@@ -51,7 +51,7 @@
 
                                 resourcesFactory.addHost($scope.newHost)
                                     .success(function(data, status){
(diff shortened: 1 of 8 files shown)


web/ui/src/Hosts/HostsController.js
```diff
@@ -80,52 +80,6 @@
             });
         };
```
rDr4g0n: this dialog needs to be shared between host details and host page, so it is moved to the modal package where anyone can get to it.

web/ui/src/common/modalService.js
```diff
@@ -242,8 +244,54 @@
 
             }
 
+            let displayHostKeys = function(keys, name) {
```
rDr4g0n: See, i told you i moved it to the modal package!

PR #3043, opened 2016-10-03
Optimize registry sync with Zookeeper

tgjonestx:
In timing StartService, we found approximately 25% of time was spent in `SyncServiceRegistry` in `zzk/registry/registry.go`.  That find was true for ~177 or ~326 services even though the number of public port addresses and vhosts was constant in both cases (0 and 3 respectively).

This change caches data by service ID so that each schedule-service operation does not have to examine the ZK namespace if there was nothing previously cached for a given serviceID. 

A major assumption here is that only the master can add/chg/delete these entries in zookeeper.

### DIFF
--- a/zzk/registry/registry.go
+++ b/zzk/registry/registry.go
@@ -16,14 +16,22 @@ package registry
 import (
 	"fmt"
 	"path"
+	"sync"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/control-center/serviced/coordinator/client"
 	"github.com/control-center/serviced/logging"
 )
 
-// initialize the package logger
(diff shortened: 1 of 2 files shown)


zzk/registry/registry.go
```diff
@@ -48,6 +56,22 @@ type VHostKey struct {
...
+	ServiceID string
+}
+
+// VHostCache holds cached information used to optimize the registry sync process
+type VHostCache struct {
+	Key       VHostKey
+	Value     VHost
+	Path      string
+	ServiceID string
+}
```
smousa: I think this caching layer needs to be on the facade.

PR #3047, opened 2016-10-05
cache public endpoints in facade to optimize interactions with ZK

tgjonestx:
In benchmarking the previous impl for 100s of services, approx 1/4-1/3 of the time was spent in `zzk/registry/registry.go::SyncServiceRegistry` method

This PR introduces `facade/serviceregistry_cache.go` to cache public endpoint data (public ports and vhosts) in the facade so that no ZK interactions are required for services that do not have any public endpoint data that needs to be updated in ZK.

A major assumption here is that only the master can add/chg/delete these entries in zookeeper.


smousa:
Crap, I just thought of a better fix for this

PR #3052, opened 2016-10-05
CC-2615: Authenticate CC REST API requests made from within containers using the identity token

fraibacas:
We have added a reverse proxy in the container controller that injects the rest token in the cc rest api requests.
The rest token basically wraps the auth token signing it with the delegate priv key. The CC rest api validates the rest token sender identity using the delegate's public key which is extracted from the auth token that is signed by the master.

### DIFF
--- a/auth/rest.go
+++ b/auth/rest.go
@@ -1,0 +1,55 @@
+// Copyright 2016 The Serviced Authors.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package auth
+
+import (
+	"errors"
+	"net/http"
+	"strings"
+)
+
+var (
+	ErrBadRestToken = errors.New("Could not extract auth token from REST request header")
+)
+
+func ExtractRestToken(header string) (string, error) {
+	// The expected header value is "Bearer JWT_ToKen"
+	header = strings.TrimSpace(header)
+	splitted := strings.Split(header, " ")
+	if len(splitted) >= 2 {
+		bearer := splitted[0]
+		token := splitted[len(splitted)-1]
+		if strings.ToLower(bearer) == "bearer" && len(token) > 0 {
+			return token, nil
+		}
+	}
+	return "", ErrBadRestToken
+}
+
+func ExtractRestTokenFromHeaders(h http.Header) (string, error) {
+	rawHeader := h.Get("Authorization")
+	if rawHeader != "" {
+		return ExtractRestToken(rawHeader)
+	}
+	return rawHeader, nil //Token not present
+}
+
+func ValidateRestToken(token string) (bool, error) {
+	identity, err := ParseJWTIdentity(token)
+	if err != nil {
+		return false, err
+	}
+	valid := !identity.Expired() && identity.HasAdminAccess()
+	return valid, nil
+}
--- a/auth/rest.go
+++ b/auth/rest.go
@@ -23,7 +23,7 @@ var (
 	ErrBadRestToken = errors.New("Could not extract auth token from REST request header")
 )
 
-func BuildRestToken() string {
+func BuildRestToken(req *http.Request) string {
 	return AuthToken()
 }
 
--- a/auth/rest.go
+++ b/auth/rest.go
@@ -14,8 +14,10 @@
 package auth
 
 import (
+	"bytes"
 	"crypto/sha256"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"strings"
 	"time"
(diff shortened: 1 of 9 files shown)


container/controller.go
```diff
@@ -621,6 +625,9 @@ func (c *Controller) Run() (err error) {
 	endpointExit := make(chan struct{})
 	c.endpoints.Run(endpointExit)
 
+	// Start CC Rest API Proxy
+	go c.ccApiProxy.run() // TODO how to control shutdown?
```
fraibacas: Should I do anything to shutdown?

auth/rest.go
```diff
@@ -0,0 +1,180 @@
...
+
+func (t *jwtRestToken) HasAdminAccess() bool {
+	return t.authIdentity.HasAdminAccess()
+}
+
+func (t *jwtRestToken) RestToken() string {
+	return t.restToken
+}
+
+func GetRequestHash(r *http.Request) []byte {
```
fraibacas: Do we need this? Since the token is only valid for 10 seconds we should not worry about someone replaying it

fraibacas:
Some context: We have added a reverse proxy in the container controller that injects the rest token in the cc rest api requests.
The rest token basically wraps the auth token signing it with the delegate priv key. The CC rest api validates the rest token sender identity using the delegate's public key which is extracted from the auth token that is signed by the master.

fraibacas:
Demo:
Attaching to any container we can curl to the cc proxy port and perform rest calls without passing any password or cookie:

`
(zenoss) [zenoss@fc42eb77e867 ~]$  curl -k  http://localhost:443/hosts ; echo ""
{"6f0a2b17":{"ID":"6f0a2b17","Name":"ip-10-111-23-43.zenoss.loc","PoolID":  .... output truncated
`
`
(zenoss) [zenoss@fc42eb77e867 ~]$  curl -k  -X POST http://localhost:443/login
{"Detail":"Accepted","Links":[{"Name":"Retrieve","Method":"GET","Url":"/"}]}
`

auth/rest.go
iancmcc: You need to hash the body and query args as well.

web/session.go
```diff
@@ -125,6 +125,35 @@ func loginOK(r *rest.Request) bool {
...
 }
 
+func loginWithTokenOK(r *rest.Request, token string) bool {
+	restToken, err := auth.ParseRestToken(token)
+	//validToken, err := auth.ValidateRestToken(r.Request, token)
+	if err != nil {
+		msg := "Unable to parse rest token"
+		plog.WithError(err).Info(msg)
+		return false
+	} else if !restToken.HasAdminAccess() {
```
iancmcc: This, I assume, is where we eventually do more complex authorization?

web/session.go
```diff
@@ -125,6 +125,35 @@ func loginOK(r *rest.Request) bool {
...
+		msg := "Could not login with rest token. Insufficient permissions."
+		plog.Info(msg)
```
iancmcc: Maybe debug this guy.

web/session.go
```diff
@@ -125,6 +125,35 @@ func loginOK(r *rest.Request) bool {
 	return true
 }
 
+func loginWithTokenOK(r *rest.Request, token string) bool {
+	restToken, err := auth.ParseRestToken(token)
+	//validToken, err := auth.ValidateRestToken(r.Request, token)
+	if err != nil {
+		msg := "Unable to parse rest token"
+		plog.WithError(err).Info(msg)
```
iancmcc: This is probably a debug.

web/session.go
iancmcc: Actually definitely debug all of these, and use `plog.WithField("url", r.url)`. `r.url` may not be a thing, but however you get the url from the request.

web/session.go
```diff
@@ -125,6 +125,35 @@ func loginOK(r *rest.Request) bool {
...
+	} else {
+		return true
+	}
+}
+
+func loginOK(r *rest.Request) bool {
+	token, tErr := auth.ExtractRestToken(r.Request)
+	if tErr != nil { // There is a token in the header but we could not extract it
+		msg := "Unable to extract auth token from header"
+		plog.WithError(tErr).Info(msg)
```
iancmcc: see above re: logging.

auth/rest.go
```diff
@@ -0,0 +1,180 @@
...
+	"crypto/sha256"
+	"fmt"
+	"net/http"
+	"strings"
+	"time"
+
+	jwt "github.com/dgrijalva/jwt-go"
+)
+
+type AuthTokenRetriever func() (string, error)
```
iancmcc: Is there a reason to define this as a type?

auth/rest.go
iancmcc: Yes, we need it. We should worry about everything, technically, or people who care about security holes will find it.

container/controller.go
iancmcc: we could use the `context` package here to do it, but I think don't worry about it. We'll need to implement graceful shutdown from a channel across the board later when we do wire in context.

container/cc_api_proxy.go
```diff
@@ -0,0 +1,60 @@
...
+	proxyPort int
+	ccApiPort int
+	proxy     *httputil.ReverseProxy
+}
+
+func newServicedApiProxy() *servicedApiProxy {
+	proxyPort := node.SERVICED_UI_ENDPOINT_PROXY
+	ccApiPort := node.SERVICED_UI_ENDPOINT
+	director := func(req *http.Request) {
+		req.URL.Scheme = "https"
```
iancmcc: I might be reading this wrong, but I think this all needs to switch on whether TLS is enabled for the CC UI.

auth/rest.go
fraibacas: Just to enforce a func signature. I define a var of this type and overwrite it from tests.

web/session.go
iancmcc: For this PR, yes. For the CC zenpack to connect we might need to whitelist an endpoint or two? Unless that always goes out through the host network stack rather than through our proxy.

web/session.go
```diff
@@ -130,11 +130,11 @@ func loginWithTokenOK(r *rest.Request, token string) bool {
 	//validToken, err := auth.ValidateRestToken(r.Request, token)
 	if err != nil {
 		msg := "Unable to parse rest token"
-		plog.WithError(err).Info(msg)
+		plog.WithError(err).WithField("url", r.URL.String()).Debug(msg)
```
iancmcc: This is fine, but in the future you might consider redefining plog:
```
plog := plog.WithField("url", r.URL.String())

plog.Debug(msg)
plog.Debug(othermsg)
```

web/session.go
```diff
@@ -125,6 +125,35 @@ func loginOK(r *rest.Request) bool {
 	return true
 }
 
+func loginWithTokenOK(r *rest.Request, token string) bool {
+	restToken, err := auth.ParseRestToken(token)
+	//validToken, err := auth.ValidateRestToken(r.Request, token)
```
iancmcc: This should maybe be uncommented? Something needs to validate that `token.RequestHash()` matches `GetRequestHash(r.Request)`

PR #3055, opened 2016-10-06
Remove double tree-walk on scheduleService

tgjonestx:
Fixes CC-2733

Walk the tree once to get a list of services, and then iterate over that list twice: once to validate the start schedule, and again to actually schedule the service.


facade/service.go
```diff
@@ -1141,12 +1156,6 @@ func (f *Facade) scheduleOneService(ctx datastore.Context, tenantID string, svc
 		svc.DesiredState = int(desiredState)
 	}
 
-	err := f.validateServiceStart(ctx, svc)
```
rphillipsz: This shouldn't have been here; it was only called previously if desiredState != service.SVCStop

PR #3068, opened 2016-10-10
Make path to master key readable

jafcrocker:
Because we do not have a proper serviced user group, we made the master
host key file world readable.  However, the directory in which that file
exists is not world readable, so the file cannot be read by other
processes (CLI commands in particular).  Make the directory readable.

Per Ian.

### DIFF
--- a/auth/localkeys.go
+++ b/auth/localkeys.go
@@ -193,7 +193,7 @@ func LoadMasterKeys(public crypto.PublicKey, private crypto.PrivateKey) {
 //  write them to disk.
 func CreateOrLoadMasterKeys(filename string) error {
 	if _, err := os.Stat(filename); os.IsNotExist(err) {
-		if err = os.MkdirAll(path.Dir(filename), os.ModeDir|0700); err != nil {
+		if err = os.MkdirAll(path.Dir(filename), os.ModeDir|0755); err != nil {
 			return err
 		}
 


jafcrocker:
DEMO
====
**Before**
Note that after deleting the current host, we cannot list hosts unless we are running as root.
```
jcrocker@jcrocker-dev:~/src/metis$ serviced host list
ID            Pool         Name              Addr               RPCPort      Cores      RAM        Cur/Max/Avg           Network                     Release
007f0101      default      jcrocker-dev      10.87.128.121      4979         12         94.3G      0 / 0 / 8388608T      172.17.0.0/255.255.0.0      
jcrocker@jcrocker-dev:~/src/metis$ serviced -v 3 host rm 007f0101
007f0101
jcrocker@jcrocker-dev:~/src/metis$ serviced -v 3 host list
No authentication token available
jcrocker@jcrocker-dev:~/src/metis$ sudo -E `which serviced` host list
no hosts found
```

**After**
```
jcrocker@jcrocker-dev:~/src/metis$ serviced host list
ID            Pool         Name              Addr               RPCPort      Cores      RAM        Cur/Max/Avg           Network                     Release
007f0101      default      jcrocker-dev      10.87.128.121      4979         12         94.3G      0 / 0 / 8388608T      172.17.0.0/255.255.0.0      
jcrocker@jcrocker-dev:~/src/metis$ serviced host rm 007f0101
007f0101
jcrocker@jcrocker-dev:~/src/metis$ serviced host list
no hosts found
```

PR #3070, opened 2016-10-11
Do not inject cc user/pass env variable in containers. Inject service…

fraibacas:
…d version env variable.

### DIFF
--- a/node/container.go
+++ b/node/container.go
@@ -29,7 +29,6 @@ import (
 	"github.com/control-center/serviced/commons/iptables"
 	"github.com/control-center/serviced/dfs/registry"
 	"github.com/control-center/serviced/domain/service"
-	"github.com/control-center/serviced/domain/user"
 	"github.com/control-center/serviced/rpc/master"
 	"github.com/control-center/serviced/servicedversion"
 	"github.com/control-center/serviced/utils"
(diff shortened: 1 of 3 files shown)


node/container.go
```diff
@@ -136,13 +137,13 @@ func (a *HostAgent) StartContainer(cancel <-chan interface{}, serviceID string,
...
 	evaluatedService.ImageID = imageName
 
-       // Establish a connection to the master
-       masterClient, err := master.NewClient(a.master)
-       if err != nil {
-               logger.WithField("master", a.master).WithError(err).Debug("Could not connect to the master")
-               return nil, nil, err
-       }
-       defer masterClient.Close()
+	// Establish a connection to the master
```
tgjonestx: If I'm not mistaken, I think you can remove lines 141-149, and remove all references to `systemUser`. In other words, if `systemUser` is no longer injected into the container, I believe that you will find it's not used anywhere else in this file. Removing the call to  `masterClient.GetSystemUser()` will also reduce the number of RPC calls back to the master for each service started on a delegate, which should help performance when the master is trying to start >1000 services

fraibacas: We need systemUser in line 157 to call setupContainer

fraibacas: It is not used in any of the methods. I will remove it

PR #3077, opened 2016-10-12
Feature/cc 2739

morr-zenoss:


morr-zenoss:
There is a test that fails with a data race. I'll need to fix that before reopening this PR.

PR #3078, opened 2016-10-12
Feature/cc 2711 3

rDr4g0n:
https://jira.zenoss.com/browse/CC-2711

Makes host status icon not only show if a host is active, but also if it is authenticated.

unauthedhost

With the additional of auth and host keys, it is easy to have a host that is connected, but cannot operate because its auth token is invalid. This makes it more obvious if that is the case, so that the user can resolve the issue.

### DIFF
--- a/auth/hostregistry.go
+++ b/auth/hostregistry.go
@@ -20,7 +20,8 @@ import (
 	"errors"
 	"fmt"
 	"sync"
-	"time"
+
+	jwt "github.com/dgrijalva/jwt-go"
 )
 
 var (
@@ -57,7 +58,7 @@ func (reg *HostExpirationRegistry) Expired(hostid string) (bool, error) {
 		// if it doesnt exist, I guess it's expired
 		return true, ErrMissingHost
 	}
-	now := time.Now().Unix()
+	now := jwt.TimeFunc().Unix()
 	return now >= expiration, nil
 }
 
--- a/auth/hostregistry.go
+++ b/auth/hostregistry.go
@@ -45,6 +45,13 @@ func (reg *HostExpirationRegistry) Set(hostid string, expires int64) {
 	reg.registry[hostid] = expires
 }
 
+// Remove removes a host from the expiration registry
+func (reg *HostExpirationRegistry) Remove(hostid string) {
+	reg.Lock()
+	defer reg.Unlock()
+	delete(reg.registry, hostid)
+}
+
 // IsExpired checks if a give host's auth has expired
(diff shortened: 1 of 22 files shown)


auth/hostregistry.go
```diff
@@ -0,0 +1,75 @@
...
+	// exist in the host expiration registry
+	ErrMissingHost = errors.New("Host is not present in host expiration registry")
+)
+
+// HostExpirationRegistry is a threadsafe map of
+// host id to auth expiration time. NOTE: expired hosts
+// are not removed from the registry
+type HostExpirationRegistry struct {
+	registry map[string]int64
+	sync.Mutex
```
iancmcc: Please use a `sync.RWMutex` instead. It'll add significant performance benefits in the `IsExpired` call.

auth/hostregistry_test.go
```diff
@@ -0,0 +1,62 @@
...
+	. "gopkg.in/check.v1"
+	"time"
+)
+
+func (s *TestAuthSuite) TestHostExpired(c *C) {
+	reg := auth.NewHostExpirationRegistry()
+	hostid := "fakehost"
+	expires := jwt.TimeFunc().Add(time.Minute).Unix()
+	reg.Set(hostid, expires)
+
```
iancmcc: Also add a test here for whether a nonexistent host returns expired

facade/host.go
```diff
@@ -351,6 +363,46 @@ func (f *Facade) FindReadHostsInPool(ctx datastore.Context, poolID string) ([]ho
...
+		}
+
+		status := host.HostStatus{HostID: id, MemoryUsage: service.Usage{}}
+		active, err := f.zzk.IsHostActive(h.PoolID, h.ID)
+		if err != nil {
+			continue
+		}
+		status.Active = active
+
+		authed, _ := f.hostRegistry.IsExpired(h.ID)
```
iancmcc: This seems like a peculiarly named variable

web/ui/src/common/resourcesFactory.js
```diff
@@ -69,6 +69,12 @@
                 method: GET,
                 url: id => `/api/v2/pools/${id}/hosts`
             },
+            // TODO - MOVE THIS WHEN THE V2 API
```
iancmcc: maybe open a ticket for this?

facade/host.go
rDr4g0n: I made a separate PR for this egregious error https://github.com/control-center/serviced/pull/3080

PR #3087, opened 2016-10-13
Make CC UI not rely on full application state

rDr4g0n:
This change does approximately one million things. Primarily it makes the CC UI not need the entire application state (all services, all hosts, all pools) to operate.  

This does introduce some regressions, but this code needs to start QA asap.

### DIFF
--- a/web/ui/src/Services/Service.js
+++ b/web/ui/src/Services/Service.js
@@ -1,0 +1,360 @@
+(function () {
+    'use strict';
+
+    // share angular services outside of angular context
+    let $notification, serviceHealth, $q, resourcesFactory, utils, Instance;
+
+    controlplane.factory('Service', ServiceFactory);
+
+    // DesiredState enum
+    var START = 1,
+        STOP = 0,
+        RESTART = -1;
+
+    // service types
+    var ISVC = "isvc",           // internal service
+        APP = "app",             // service with no parent
+        META = "meta",           // service with children but no startup command
+        DEPLOYING = "deploying"; // service whose parent is still being deployed
+
+    function fetch(methodName, propertyName, force) {
+        let deferred = $q.defer();
+
+        if (this[propertyName] && !force) {
+            deferred.resolve();
+            return deferred.promise;
+        }
+        // NOTE: V2 API only
+        // TODO: make sure methodName exists
+        // TODO: error callback
+        resourcesFactory.v2[methodName](this.id)
+            .then(data => {
+                // console.log(`fetched ${data.length} ${propertyName} from ${methodName} for id ${this.id}`);
+                this[propertyName] = data;
+                this.touch();
+                deferred.resolve();
+            },
+            error => {
+                deferred.reject(error);
+                console.warn(error);
+            });
+
+        return deferred.promise;
+    }
+
+    function getDescendents(descendents, service) {
+        if (service.subservices) {
+            service.subservices.forEach(svc => getDescendents(descendents, svc));
+        }
+        descendents[service.id] = service;
+        return descendents;
+    }
+
+    class Service {
+
+        constructor(model) {
+            this.subservices = [];
+            this.instances = [];
+            this.update(model);
+        }
+
+        update(model) {
+            // basically new-up with exisiting children and instances
+            this.name = model.Name;
+            this.id = model.ID;
+            this.desiredState = model.DesiredState;
+            this.model = Object.freeze(model);
+            this.evaluateServiceType();
+            this.touch();   
+        }
+
+        evaluateServiceType() {
+            // infer service type
+            this.type = [];
+            if (this.model.ID.indexOf("isvc-") !== -1) {
+                this.type.push(ISVC);
+            }
+
+            if (!this.model.ParentServiceID) {
+                this.type.push(APP);
+            }
+
+            if (this.subservices.length && !this.model.Startup) {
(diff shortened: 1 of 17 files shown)


web/ui/src/Services/Service.js
```diff
@@ -0,0 +1,361 @@
...
+                });
+            return deferred.promise;
+        }
+
+        // fast-moving state for this service and its instances
+        // note: returns a promise that resolves with a single status object
+        getStatus() {
+            let deferred = $q.defer();
+
+            resourcesFactory.v2.getServiceStatus(this.id)
```
carlcamera: This is for the current service only -- the one mentioned at the top of the page - so we can fill in all the instance tables and endpoints etc. We batch lite service status calls for all the sub-services into a single call elsewhere.

web/ui/src/Services/ServiceDetailsController.js
```diff
@@ -3,1076 +3,1107 @@
...
-            var validateVHost = function(newPublicEndpoint){
-                var name = newPublicEndpoint.name;
-
-                // if no port
-                if(!name || !name.length){
-                    return "Missing Name";
+        ["$scope", "$q", "$routeParams", "$location", "resourcesFactory",
+            "authService", "$modalService", "$translate", "$notification",
+            "$timeout", "miscUtils", "hostsFactory", "$serviceHealth", "Service",
+            "poolsFactory", "CCUIState", "$cookies", "areUIReady", "LogSearch",
```
carlcamera: I'd like to eventually remove the servicesFactory but I don't know of plans to remove poolsFactory or hostsFactory.

web/ui/src/Services/Service.js
rDr4g0n: This method is used by the services controller to get the status for the current service only. Just below it is `updateDescendentStatuses` which gets statuses for all descendant services at once.

web/ui/src/Services/ServiceDetailsController.js
```diff
@@ -3,1076 +3,1107 @@
...
+                    $scope.protocols = [];
+                    $scope.protocols.push({ Label: "HTTPS", UseTLS: true, Protocol: "https" });
+                    $scope.protocols.push({ Label: "HTTP", UseTLS: false, Protocol: "http" });
+                    $scope.protocols.push({ Label: "Other, secure (TLS)", UseTLS: true, Protocol: "" });
+                    $scope.protocols.push({ Label: "Other, non-secure", UseTLS: false, Protocol: "" });
+
+                    // default public endpoint options
+                    $scope.publicEndpoints.add = {
+                        type: "port",
+                        app_ep: $scope.currentService.exportedServiceEndpoints[0],
```
carlcamera: similar situation here. the endpoints object created here was existing code.  Since there are many tie-ins to templates, etc. I did not change the object property name.

web/ui/src/Services/ServiceDetailsController.js
rDr4g0n: after things are settled down, yeah: https://jira.zenoss.com/browse/CC-2765

web/ui/src/common/resourcesFactory.js
```diff
@@ -312,6 +312,88 @@
...
+            },
+            getServiceMonitoringProfile: {
+                method: GET,
+                url: id => `/api/v2/services/${id}/monitoringprofile`,
+            },
+            getServicePublicEndpoints: {
+                method: GET,
+                url: id => `/api/v2/services/${id}/publicendpoints?includeChildren`,
+            },
+            getServiceExportEndpoints: {
```
carlcamera: Good point and would keep the code cleaner. We can discuss.

web/ui/src/common/resourcesFactory.js
```diff
@@ -312,6 +312,88 @@
...
+            },
+            getServiceIpAssignments: {
+                method: GET,
+                url: id => `/api/v2/services/${id}/ipassignments?includeChildren`,
+            },
+            getServiceMonitoringProfile: {
+                method: GET,
+                url: id => `/api/v2/services/${id}/monitoringprofile`,
+            },
+            getServicePublicEndpoints: {
```
rDr4g0n: Yeah, we need it for the top level stuff, like the old behavior, and having it for the others is ok too.

web/ui/src/common/resourcesFactory.js
rDr4g0n: We could, but right now we always want all children to make sure it behaves like the old code did. When the need arises, we can easily add the second arg to choose not to get all children

web/ui/src/Services/ServiceDetailsController.js
rDr4g0n: dooooonne!

web/ui/src/Services/ServiceDetailsController.js
```diff
@@ -3,1076 +3,1107 @@
...
+                                actions: [
+                                    {
+                                        role: "cancel",
+                                        label: "close"
+                                    }, {
+                                        classes: "btn-primary",
+                                        label: "refresh",
+                                        icon: "glyphicon-repeat",
+                                        action: function () {
+                                            var textarea = this.$el.f
```
rDr4g0n: Can you clarify this comment? Cant quite figure out what youre referring to.

PR #3094, opened 2016-10-14
CC-2772 UILite Instance status icons disappear

carlcamera:
Fix for the disappearing status icons in instance tables. Also fixes hover-bubbles that cannot be inspected continuously to observe changes in dependent process status.

### DIFF
--- a/web/ui/src/Instances/Instance.js
+++ b/web/ui/src/Instances/Instance.js
@@ -17,14 +17,14 @@
         }
 
         update(data) {
-            // the instance model data comes in with health and
-            // memory stats, so use that to do an initial instace
-            // status update
             this.model = Object.freeze(data);
             this.id = buildStateId(this.model.HostID, this.model.ServiceID, this.model.InstanceID);
             this.resources = {
                 RAMCommitment: utils.parseEngineeringNotation(data.RAMCommitment)
             };
+            // the instance model data comes in with health and
+            // memory stats, so use that to do an initial instace
+            // status update
             this.updateState({
                 HealthStatus: data.HealthStatus,
                 MemoryUsage: data.MemoryUsage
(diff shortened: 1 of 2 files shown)


web/ui/src/Services/Service.js
```diff
@@ -141,9 +141,18 @@
         fetchInstances() {
             let deferred = $q.defer();
             resourcesFactory.v2.getServiceInstances(this.id)
-                .then(data => {
-                    // TODO - dont blow away existing instances
-                    this.instances = data.map(i => new Instance(i));
+                .then(results => {
+                    // console.log(`fetched ${data.length} instances for ${this.id}`);
```
rDr4g0n: remove this log message please

web/ui/src/Instances/Instance.js
```diff
@@ -12,20 +12,22 @@
...
-                RAMCommitment: utils.parseEngineeringNotation(instance.RAMCommitment)
-            };
+        constructor(data) {
+            this.update(data);
+        }
 
+        update(data) {
             // the instance model data comes in with health and
             // memory stats, so use that to do an initial instace
             // status update
```
rDr4g0n: this comment belongs with `updateState` a few lines below

web/ui/src/Services/Service.js
```diff
@@ -141,9 +141,18 @@
...
+                    results.forEach(data => {
```
rDr4g0n: Please comment here that you are merging instances to prevent anything bounce to the instances array from updating.

web/ui/src/Services/Service.js
carlcamera: Done.

web/ui/src/Services/Service.js
carlcamera: Added the following
{noformat}
// new-ing instances will cause UI bouncing and force rebuilding
// of the popover. To minimize UI churn update/merge status info
// into exisiting instance ojects  
{noformat}

PR #3106, opened 2016-10-17
CC-2785 Scheduler should not schedule services on a host that isn't registered

johnlhamilton:


scheduler/leader.go
```diff
@@ -75,18 +77,37 @@ func (l *leader) SelectHost(sn *zkservice.ServiceNode) (string, error) {
...
-	if len(hosts) == 0 {
+	if len(reghosts) == 0 {
 		plog.Debug("Scheduler is shutting down")
 		return "", errors.New("scheduler is shutting down")
 	}
 
+	// filter out hosts that have not been authenticated
+	hosts := []host.Host{}
+	for _, h := range reghosts {
+		hlogger := logger.WithField("HostID", h.ID)
```
smousa: log fields need to be lowercase

scheduler/scheduler.go
```diff
@@ -31,6 +32,10 @@ import (
 
 var plog = logging.PackageLogger()
 
+var (
+	ErrNoAuthenticatedHosts = errors.New("No authenticated hosts found")
```
smousa: error messages should not be capitalized

PR #3109, opened 2016-10-17
CC-2773: Show the confirmation modal with total number of child services when start/stop is clicked

iancmcc:


makefile
```diff
@@ -160,7 +160,7 @@ FORCE:
 serviced: $(GO)
 serviced: FORCE
 	$(GO) build $(GOBUILD_FLAGS) ${LDFLAGS}
-	make govet
+	# make govet
```
rDr4g0n: is this intentional?

web/ui/src/Services/ServiceDetailsController.js
```diff
@@ -324,34 +324,39 @@
...
-                        $scope.modal_confirmSetServiceState(service, state, childCount);
-
-                        // if no children, just start the service
-                    } else {
-                        $scope.setServiceState(service, state);
-                    }
+                    var targetState = serviceStateChangeTargets[state];
+                    resourcesFactory.v2.getDescendantCounts(service.id)
+                      .success(function (data, status) {
+                        var childCount = data.auto[targetState] || 0;
```
rDr4g0n: I think a comment here about what `auto` refers to will be useful.

web/ui/src/Services/ServiceDetailsController.js
```diff
@@ -343,7 +343,28 @@
...
+                            if (data.auto) {
+                              childCount += data.auto["0"] || 0;
+                            }
+                            break;
+                          case "restart":
+                          case "stop":
+                            // When stopping or restarting, we care about
+                            // running services that are either manual or
+                            // autostart
+                            if (data.auto) {
```
rDr4g0n: oh, good catch :D

PR #3110, opened 2016-10-18
CC-2778, add Application property to getServiceAddressAssignmentDetails

ecifreo:
CC-2778, add Application property to getServiceAddressAssignmentDetails v2 API
so the UI can show the endpoints's application name.


smousa:
@jplouis this field isn't for display and the other endpoint calls use Application

PR #3112, opened 2016-10-18
Add logging configuration for logri

jafcrocker:
- Add and load logri configuration files, one for Server and one for CLI.  (The difference is motivated by differing default severity levels.)
- Map from existing glog controls (i.e., _-v_ command line parameter, SIGUSR1 handling) to logri.
- Add a LogControl interface to facilitate unit tests and update existing tests.
- Add new unit tests around logri logging configuration.

Fixes [CC-2761](https://jira.zenoss.com/browse/CC-2761)
See related https://github.com/zenoss/zendev/pull/307


cli/cmd/cmd.go
```diff
@@ -376,37 +394,47 @@ func getEndpoint(options config.Options) string {
...
-	if ctx.IsSet("logtostderr") {
-		glog.SetToStderr(ctx.GlobalBool("logtostderr"))
+	go logControl.WatchConfigFile(options.LogConfigFilename)
+
+	// Set glog verbosity.  Map the glog verbosity level onto logrus levels as best
+	// we can, so that the verbosity for both can be controlled with a single argument.
+	if ctx.IsSet("v") {
+		verbosity := ctx.GlobalInt("v")
+		logControl.SetVerbosity(verbosity)
+		logControl.SetLevel(logrus.DebugLevel - logrus.Level(verbosity))
```
johnlhamilton: If verbosity is a large number, what happens here?

cli/cmd/cmd.go
```diff
@@ -376,37 +394,47 @@ func getEndpoint(options config.Options) string {
...
-		logrus.SetLevel(logrus.WarnLevel)
-		logri.SetLevel(logrus.WarnLevel)
+func setLogging(options *config.Options, ctx *cli.Context, logControl logging.LogControl) error {
+	// Rather than immediately returning from the function on error, keep track
+	// of the errors and continue.  This allows us to process all arguments,
+	// start watchers, etc.
+	var errors []error
+
+	if err := logControl.ApplyConfigFromFile(options.LogConfigFilename); err != nil {
+		errors = append(errors, err)
```
jafcrocker: If we were planning on maintaining the glog controls permanently then I would say "yes, we need to report all of the errors somehow".  However, since all of these controls are going away with glog, the controls are very rarely used, and the behavior is the same is before these changes (i.e., only reports the first error), I think it is acceptable.  

If you think we need to report them all, I would probably return a list of errors to the calling routine and then report all of them from there.  Something like this:
https://github.com/golang/appengine/blob/master/errors.go#L22-L25

cli/cmd/cmd.go
jafcrocker: Agreed; this is a problem.

PR #3121, opened 2016-10-19
Fix race condition in cli/cmd unit test

jafcrocker:
This fixes the integration tests.

Previously, the assertion that all expectations were met (i.e., all routines which were expected to be called were actually called) would fail for the "WatchConfigFile" routine.  This was due to the fact that the routine was called in a goroutine, so it may not have been called by the time it was checked.

This change adds a channel, calls a send on the channel when the mock routine is called, and waits (with a timeout) on a receive for the channel.  This ensures that the routine runs before we check the calls.

### DIFF
--- a/cli/cmd/cmd_logging_test.go
+++ b/cli/cmd/cmd_logging_test.go
@@ -19,12 +19,14 @@ import (
 	"errors"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/Sirupsen/logrus"
 	mockapi "github.com/control-center/serviced/cli/api/apimocks"
 	"github.com/control-center/serviced/domain/host"
 	mocklog "github.com/control-center/serviced/logging/mocks"
 	"github.com/control-center/serviced/utils"
+	"github.com/stretchr/testify/mock"
 
 	. "gopkg.in/check.v1"
 )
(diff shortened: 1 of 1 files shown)


jafcrocker:
The integration tests were broken prior to this change.  e.g., 
```
----------------------------------------------------------------------
FAIL: cmd_logging_test.go:77: CmdLoggingSuite.Test_LoadsConfigFile_Error

✅	ApplyConfigFromFile(string)
❌	WatchConfigFile(string)
cmd_logging_test.go:90:
    s.log.AssertExpectations(c)
/workspace/workspace/ControlCenter/develop/serviced-develop-merge-unit-integration/gopath/src/github.com/control-center/serviced/vendor/github.com/stretchr/testify/mock/mock.go:375:
    t.Errorf("FAIL: %d out of %d expectation(s) were met.\n\tThe code you are testing needs to make %d more call(s).\n\tat: %s", len(expectedCalls)-failedExpectations, len(expectedCalls), failedExpectations, assert.CallerInfo())
... Error: FAIL: 1 out of 2 expectation(s) were met.
	The code you are testing needs to make 1 more call(s).
	at: [cmd_logging_test.go:90]

✅	GetHosts()
```

PR #3123, opened 2016-10-19
cloning a service should filter BuiltIn monitoring configs

smousa:


domain/service/service.go
```diff
@@ -275,6 +275,32 @@ func CloneService(fromSvc *Service, suffix string) (*Service, error) {
 		svc.Volumes[idx].ResourcePath += suffix
 	}
 
+	// CC-2750: filter invalid monitoring profile data
```
tgjonestx: Unit tests?

PR #3127, opened 2016-10-20
Refactor "pipe" to take a function with no args

jafcrocker:
Previously, cmd.pipe() took as arguments a function with a variadic string parameter, and a variadic string parameter.  The function was called with the strings.

After this change, the pipe function (which has been renamed to captureStdout) takes a single parameterless function instead.  This prevents the proliferation of pipe functions, each taking different
arguments (e.g., pipeAPI).  To pass parameters to the function, a closure can be formed from an anonymous function.

Also created a captureStderr which operates in a parallel fashion, and refactored pipeStderr to take a single parameterless function and use captureStderr.

This design makes it easy to capture the output streams from an arbitrary function.

### DIFF
--- a/cli/cmd/cmd_logging_test.go
+++ b/cli/cmd/cmd_logging_test.go
@@ -89,10 +89,8 @@ func (s *CmdLoggingSuite) Test_LoadsConfigFile_Error(c *C) {
 			watch <- struct{}{}
 		})
 	s.api.On("GetHosts").Return([]host.Host{}, nil)
-	f := func(args ...string) { s.Run(env, args...) }
-	g := func(args ...string) { pipeStderr(f, args...) }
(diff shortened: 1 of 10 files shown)


jafcrocker:
DEMO
-----
Tests still run
```
jcrocker@jcrocker-dev:~/src/metis/src/github.com/control-center/serviced$ ./serviced-tests.py --cover --unit --package ./cli/...
ok      github.com/control-center/serviced/cli/api      0.110s  coverage: 13.9% of statements
?       github.com/control-center/serviced/cli/api/apimocks     [no test files]
?       github.com/control-center/serviced/cli/api/mocks        [no test files]
ok      github.com/control-center/serviced/cli/cmd      0.235s  coverage: 58.9% of statements
```

PR #3130, opened 2016-10-20
CC-2770: New Isvcs pages

cchristiezen:
New Internal Services pages to work with the new v2 endpoints.


web/api_isvcs.go
```diff
@@ -0,0 +1,210 @@
...
+
+	for _, service := range getISVCS() {
+		if service.ID == isvcs.InternalServicesISVC.ID {
+			results = append(results, getParent())
+		} else {
+			results = append(results, getService(service))
+		}
+	}
+
+	w.WriteJson(results)
```
rDr4g0n: `web.writeJSON` is used inconsistently throughout the `web` package, but maybe that should be used here instead of directly calling the `ResponseWriter`'s method.

web/api_isvcs.go
```diff
@@ -0,0 +1,210 @@
...
+			w.WriteJson(getService(service))
+			return
+		}
+	}
+
+	writeJSON(w, "Internal Service Not Found.", http.StatusNotFound)
+}
+
+func getInternalServiceInstances(w *rest.ResponseWriter, r *rest.Request, ctx *requestContext) {
+	id, err := url.QueryUnescape(r.PathParam("id"))
```
rDr4g0n: `getInternalServiceInstances` plural but appears to only return 1 instance?

PR #3133, opened 2016-10-21
Remove last v1 api thingies from ServiceDetailsController

rDr4g0n:
There were a few little bits of thingies that used the super expensive v1 endpoints in the ServiceDetailsController that needed to be removed:

* pools are now loaded when "edit service" is clicked instead of on page load
* host names are looked up from the (small/fast) host status object and cached
* a few TODOs were resolved for moving v1/v2 calls around

### DIFF
--- a/web/ui/src/Pools/PoolDetailsController.js
+++ b/web/ui/src/Pools/PoolDetailsController.js
@@ -171,7 +171,6 @@
 
         // update/refresh current pools data
         function update(){
-            // TODO - use v2 endpoint once its in
             return resourcesFactory.getPool($scope.params.poolID)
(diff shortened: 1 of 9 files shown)


web/ui/src/Pools/PoolDetailsController.js
```diff
@@ -171,7 +171,6 @@
 
         // update/refresh current pools data
         function update(){
-            // TODO - use v2 endpoint once its in
             return resourcesFactory.getPool($scope.params.poolID)
```
rDr4g0n: Its ok to use v1 endpoints to get the fully detailed object for a specific pool

PR #3136, opened 2016-10-21
CC-2798: Refactor the mux and RPC to use a common auth header protocol.

iancmcc:
This fixes the problem by prepending any authentication header request with a magic number, so it can be rejected early, as well as a protocol version so we can eventually modify the protocol while being backwards-compatible.
It also prepends RPC requests with their own magic number for similar reasons.


auth/header.go
```diff
@@ -0,0 +1,373 @@
...
+// WriteTo satisfies the io.WriterTo interface. It builds an auth header
+// according to the spec, signs it, and writes it to the specified writer.
+func (h *authHeaderWriterTo) WriteTo(w io.Writer) (n int64, err error) {
+
+	var (
+		signedContent bytes.Buffer
+		bytesWritten  int64
+	)
+
+	// MAGIC VERSION | ALLLEN SIGNATURE TIMESTAMP TOKEN PAYLOADLEN PAYLOAD
```
smousa: COMMENT

auth/header.go
```diff
@@ -0,0 +1,373 @@
...
+	payload, err := readPayload(r)
+	if err != nil {
+		return err
+	}
+	return &AuthHeaderError{e, payload}
+}
+
+// readPayload reads a payload length and the accompanying payload from
+// a reader.
+func readPayload(r io.Reader) ([]byte, error) {
```
smousa: redundant

auth/rpc.go
```diff
@@ -15,172 +15,122 @@ package auth
...
 type RPCHeaderParser interface {
-	ParseHeader(header []byte, request []byte) (Identity, error)
+	ReadHeader(io.Reader) (Identity, []byte, error)
 }
 type RPCHeaderBuilder interface {
-	BuildHeader([]byte) ([]byte, error)
+	WriteHeader(io.Writer, []byte, bool) error
 }
 
 type RPCHeaderHandler struct{}
```
smousa: Lint

PR #3143, opened 2016-10-24
Feature/cc 2825

rDr4g0n:
Fixing https://jira.zenoss.com/browse/CC-2825

When stop, start, or restart are clicked for a service, a request is made to the backend to count the number of descendants that will be affected by the change. However, if the network connection between the UI and serviced is slow, then the user can end up waiting for a number of seconds. The user needs immediate feedback that something is happening so that they know their click was received. 

This change does a few things to give the user immediate feedback on the click:

1) Shows a "one moment please" modal while waiting on the child count
2) Shows a modal even if the number of affected services is only 1  

modal
(note that this gif shows the best case scenario where the descendant count loads within 200ms. In some cases that loading dialog may be visible for many seconds)

To support these features, a few changes needed to be made

* A new "unclosable" modal feature was added so that the "one moment please" modal cant just be closed, resulting in unexpected behavior
* When modals are displayed one-after-another, the UI flashes in a yucko way, so the bootstrap modal background was removed and replaced with our own modal background, which lets us control how it shows/hides more closely
* some codemirror textarea issues popped up in modals with this change, so those were tightened up
* most of this code needs to be shared with the Apps page, so it was moved into a shared location


web/ui/src/Services/ServiceDetailsController.js
```diff
@@ -329,81 +329,35 @@
                 // children and ask the user to choose to start all
                 // children or only the top service
                 $scope.clickRunning = function (service, state) {
```
rDr4g0n: this all gets moved into `Service.js` as a class method

web/ui/src/Services/ServiceDetailsController.js
```diff
@@ -329,81 +329,35 @@
...
+                    let onStartService = function(modal){
+                        // the arg here explicitly prevents child services
+                        // from being started
+                        $scope.setServiceState(service, state, true);
+                        modal.close();
+                    };
+                    let onStartServiceAndChildren = function(modal){
+                        $scope.setServiceState(service, state);
+                        modal.close();
+                    };
```
rDr4g0n: This is all moved into `modalService.js` as a shared modal

web/ui/src/Services/ServiceDetailsController.js
```diff
@@ -495,6 +449,7 @@
                                 ],
                                 onShow: function () {
                                     modalScope.codemirrorRefresh = true;
+                                    modalScope.$apply();
```
rDr4g0n: it takes a tick or 2 for this view to update, so update it asap

tgjonestx:
Retest this

tgjonestx:
retest this

rDr4g0n:
retest this

PR #3147, opened 2016-10-25
Pass mux/rpc config options down to the container controller

tgjonestx:
Fixes CC-2816

Includes changes to pass the following to serviced container controller:
* MUX port
* flag for mux-disable-tls
* RPC port
* flag for rpc-disable-tls

### DIFF
--- a/auth/token.go
+++ b/auth/token.go
@@ -198,7 +198,7 @@ func LoadTokenFile(tokenfile string) error {
 	// No need to handle expires or save file, because we're loading from the file rather
 	// than re-requesting authentication tokens
 	updateToken(string(data), zerotime, "")
-	log.Infof("Updated authentication token from disk")
+	log.Debug("Updated authentication token from disk")
 	return nil
 }
 
--- a/auth/token.go
+++ b/auth/token.go
@@ -90,6 +90,22 @@ func AuthTokenNonBlocking() (string, error) {
 	return currentToken, nil
 }
 
+// WaitForAuthToken blocks until the authentication token is defined.
+func WaitForAuthToken(cancel <-chan interface{}) <-chan struct{} {
+	ch := make(chan struct{})
+	go func() {
+		for currentToken == "" {
+			select {
+			case <-cond.Wait():
+			case <-cancel: // Receive from nil channel never returns, so this is fine
+			}
+		}
+		close(ch)
+	}()
+	return ch
+}
+
+
 // MasterToken() generates a new token with an empty host and pool ID and the master's public key,
 //  signed by the master's private key.  This will return an error if there is no master private
 //  key available (i.e. if we are not the master)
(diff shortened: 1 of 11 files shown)


auth/token.go
```diff
@@ -164,14 +164,12 @@ func WatchTokenFile(tokenfile string, done <-chan interface{}) error {
 	})
 
 	loadToken := func() {
-		data, err := ioutil.ReadFile(tokenfile)
+		err := LoadTokenFile(tokenfile)
 		if err != nil {
 			log.WithError(err).Warn("Unable to load authentication token from file. Continuing to watch for changes")
```
iancmcc: You're double-logging here.

auth/token.go
```diff
@@ -188,6 +186,22 @@ func WatchTokenFile(tokenfile string, done <-chan interface{}) error {
...
 }
 
+func LoadTokenFile(tokenfile string) error {
+	log := log.WithFields(logrus.Fields{
+		"tokenfile": tokenfile,
+	})
+
+	data, err := ioutil.ReadFile(tokenfile)
+	if err != nil {
+		log.WithError(err).Warn("Unable to load authentication token from file.")
```
iancmcc: I believe you should replace this line with `return err`

container/controller.go
```diff
@@ -61,6 +61,10 @@ var (
 	ErrInvalidHostID = errors.New("container: invalid host id")
 	// ErrNoServiceEndpoints is returned if we can't fetch the service endpoints
 	ErrNoServiceEndpoints = errors.New("container: unable to retrieve service endpoints")
+	// ErrNoDelegateKey is returned if the key file can not be read
+	ErrNoDelegateKey = errors.New("container: no delegate key file")
```
iancmcc: Do we really need a separate error type here? Can't we just log appropriately?

container/controller.go
```diff
@@ -287,6 +292,17 @@ func NewController(options ControllerOptions) (*Controller, error) {
 		errc <- nil
 	}()
 
+	// Load the delegate keys and auth tokens first so there's no race btwn starting the watcher routines
```
iancmcc: Yeah, I guess this is fine now. Before, services would be able to be scheduled to a host before it had authenticated, so the watching was necessary.

container/controller.go
```diff
@@ -287,6 +292,17 @@ func NewController(options ControllerOptions) (*Controller, error) {
...
+	if err := auth.LoadDelegateKeysFromFile(containerDelegateKeyFile); err != nil {
+		glog.Errorf("Unable to load delegate keys from file %s: %s", containerDelegateKeyFile, err)
+		return nil, ErrNoDelegateKey
+	}
+	if err := auth.LoadTokenFile(containerTokenFile); err != nil {
+		glog.Errorf("Unable to token file %s: %s", containerTokenFile, err)
+		return nil, ErrNoAuthToken
+	}
+
 	go auth.WatchDelegateKeyFile(containerDelegateKeyFile, keyshutdown)
```
iancmcc: The first thing these do is an immediate load; was the point just to block until we have a token/key? because we have a simple way to do that already.

tgjonestx:
retest this

PR #3148, opened 2016-10-25
Do not authenticate the host unless we need to make an RPC call

tgjonestx:
Fixes CC-2844

The HA resource agent was timing out when trying to check some CC config with `serviced config` in cases where the CC master had already been stopped. This was because the CLI always issued an RPC to get a valid auth token, even for calls like `serviced config` and `serviced version` which perform no actual network I/O in and of themselves. This change moves the authenticateHost call so it only occurs for CLI calls that trigger an RPC.

### DIFF
--- a/cli/api/api.go
+++ b/cli/api/api.go
@@ -16,17 +16,20 @@ package api
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"runtime/pprof"
 	"strings"
 	"time"
 
 	"github.com/Sirupsen/logrus"
+	"github.com/control-center/serviced/auth"
 	"github.com/control-center/serviced/config"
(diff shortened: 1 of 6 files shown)


cli/api/api.go
```diff
@@ -135,6 +140,7 @@ func (a *api) connectMaster() (master.ClientInterface, error) {
 		if err != nil {
 			return nil, fmt.Errorf("could not create a client to the master: %s", err)
 		}
+		a.authenticateHost()
```
jplouis: No logging or returning the error?

tgjonestx: Well, kind of a catch-22. It's expected to fail in the case of `serviced host add ...` for adding the master host because there's no key file to start with, and add-host will work without it.

PR #3152, opened 2016-10-26
CC-2765:  App page changes.

cchristiezen:
This changes removes the use of the servicesFactory, hostsFactory, and poolsFactory from the apps page and the deployment wizard.

### DIFF
--- a/web/ui/src/Hosts/HostsController.js
+++ b/web/ui/src/Hosts/HostsController.js
@@ -129,7 +129,7 @@
 
             modalScope.newHost = {
                 port: $translate.instant('placeholder_port'),
-                PoolID: this.arrayEmpty(this.pools) ? "" : this.pools[0].id
+                PoolID: this.utils.arrayEmpty(this.pools) ? "" : this.pools[0].id
             };
 
             areUIReady.lock();
@@ -234,10 +234,6 @@
         updateHostsInView(data) {
             this.hostsInView = data;
             this.refreshHostStatuses();
-        }
-
-        arrayEmpty(array) {
-            return typeof array === "undefined" || array === null || array.length <= 0;
         }
     }
 
--- a/web/ui/src/Hosts/HostsController.js
+++ b/web/ui/src/Hosts/HostsController.js
@@ -129,7 +129,7 @@
 
             modalScope.newHost = {
                 port: $translate.instant('placeholder_port'),
-                PoolID: this.utils.arrayEmpty(this.pools) ? "" : this.pools[0].id
+                PoolID: utils.arrayEmpty(this.pools) ? "" : this.pools[0].id
             };
 
             areUIReady.lock();
--- a/web/ui/src/Hosts/HostsController.js
+++ b/web/ui/src/Hosts/HostsController.js
@@ -5,7 +5,7 @@
     "use strict";
 
     let resourcesFactory, authService, $modalService, $translate, $notification,
-        areUIReady, $interval, servicedConfig, utils, Host, log;
+        areUIReady, $interval, servicedConfig, utils, Host, log, Pool;
 
     class HostsController {
 
@@ -82,7 +82,7 @@
         refreshPoolIds() {
             resourcesFactory.v2.getPools()
                 .success(data => {
-                    this.poolIds = data.map(result => result.ID).sort();
+                    this.pools = data.map(result => new Pool(result)).sort((first, second) => first.id.localeCompare(second.id));
                 })
                 .error(data => {
                     $notification.create("Unable to load pools.", data.Detail).error();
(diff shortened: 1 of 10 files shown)


web/ui/src/Hosts/HostsController.js
```diff
@@ -237,15 +237,15 @@
         }
 
         arrayEmpty(array) {
-            return typeof array !== "undefined" && array !== null && array.length > 0;
+            return typeof array === "undefined" || array === null || array.length <= 0;
         }
```
rDr4g0n: `arrayEmpty` should be a util function, not a method.

web/ui/src/Services/AppsController.js
```diff
@@ -292,34 +267,108 @@
...
-        function fetchInternalServiceInstances(app) {
-            return resourcesFactory.v2.getInternalServiceInstances(app.id).then( data => {
-                app.instances = data;
+
+        function refreshApps() {
+            return fetchApps()
+                .then(fillOutApps)
+                .then(fetchServicesHealth)
+                .then(fetchInternalServicesHealth)
+                .then(() => touch());
```
rDr4g0n: need a `catch` in here or this beautiful promise chain will be the source of infinite sadness.

web/ui/src/Services/AppsController.js
```diff
@@ -292,34 +267,108 @@
...
+                    var services = results[1];
+                    services.forEach(s => {
+                        $scope.apps.push(s);
+                    });
+                });
+        }
+
+        function fetchChildEndpoints(app) {
+            return resourcesFactory.v2.getServiceChildPublicEndpoints(app.id).then(data => {
+                app.publicEndpoints = data;
```
rDr4g0n: In `Service.js`, public endpoints get their parent's `DesiredState` attached so that the publicEndpointLinkDirective knows if the link should be active or not. That is probably needed here.

`Service.js` also adds `Value` but that appears to only be used when listing exported service endpoints for adding public endpoints, so it can probably be skipped.

web/ui/src/Services/AppsController.js
```diff
@@ -292,34 +267,108 @@
...
+            });
+
+            return $q.all(promises);
+        }
+
+        function fetchServices() {
+            let deferred = $q.defer();
+            resourcesFactory.v2.getTenants().then( data => {
+                var services = data.map(s => new Service(s));
+                deferred.resolve(services);
```
rDr4g0n: above 2 lines need to be indented one more or something... its not quite right around here.

web/ui/src/Services/AppsController.js
```diff
@@ -292,34 +267,108 @@
...
+                    app.fetchAllStates();
+                }
+            });
+        }
+
+        function fetchInternalService() {
+            let deferred = $q.defer();
+            resourcesFactory.v2.getInternalServices()
+                .then(data => {
+                    let parentIndex = data.findIndex(i => !i.Parent);
```
rDr4g0n: looks like IE doesnt support `findIndex` :(

web/ui/src/Services/AppsController.js
```diff
@@ -292,34 +267,108 @@
...
+                    if (parentIndex > -1) {
+                        var internalServices = new InternalService(data[parentIndex]);
+                        internalServices.fetchInstances();
```
rDr4g0n: Do we need to wait till `fetchInstances` finishes?

web/ui/src/Hosts/HostsController.js
cchristiezen: Moved to utils.

web/ui/src/Services/AppsController.js
cchristiezen: Added catch to write to console.

web/ui/src/Services/AppsController.js
cchristiezen: When the status is updated, there is a call to the fetchAllStates method on the Service object which sets the desiredState of the endpoint to its parent.

web/ui/src/Services/AppsController.js
cchristiezen: Changed to find.

web/ui/src/Services/AppsController.js
cchristiezen: This call was actually unnecessary.  The fetchInstances call has been moved into a promise chain.

web/ui/src/Hosts/HostsController.js
```diff
@@ -129,7 +129,7 @@
 
             modalScope.newHost = {
                 port: $translate.instant('placeholder_port'),
-                PoolID: this.arrayEmpty(this.pools) ? "" : this.pools[0].id
+                PoolID: this.utils.arrayEmpty(this.pools) ? "" : this.pools[0].id
```
rDr4g0n: you prrroooooobably want `utils` instead of `this.utils` :>

cchristiezen: Fixed.

PR #3158, opened 2016-10-27
Feature/cc 2862

rphillipsz:


pkg/serviced.default
```diff
@@ -1,194 +1,247 @@
...
+# The default path for configuration files.  If not set, this will be the path given for SERVICED_HOME
+# followed by /etc
+# SERVICED_ETC_PATH=/opt/serviced/etc
 
-# Set the supported TLS ciphers for MUX connections
-# SERVICED_MUX_TLS_CIPHERS=TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA,TLS_RSA_WITH_AES_128_CBC_SHA,TLS_RSA_WITH_AES_256_CBC_SHA,TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
+# Set the driver type on the master for the distributed file system (rsync/btrfs/devicemapper)
+# SERVICED_FS_TYPE=devicemapper
+
+# The default home directory.
```
iancmcc: This doesn't seem quite descriptive enough. It's the directory relative to which serviced will put all its stuff.

pkg/serviced.default
```diff
@@ -1,194 +1,247 @@
...
+# Max size of Logstash data to keep in gigabytes
+# SERVICED_LOGSTASH_MAX_SIZE=10
 
-# Set to true to allow use of loopback files (instead of thin pools) with devicemapper for serviced storage.
-# NOTE: This is not recommended for production use.
-# SERVICED_ALLOW_LOOP_BACK=false
+# Set the address to which to send logs
+# SERVICED_LOG_ADDRESS={{SERVICED_MASTER_IP}}:5042
+
+# Set the loglevel for serviced
```
iancmcc: Maybe a description here? "2 is debug"

rphillipsz: This was an existing entry/comment.  I didn't review them (just added the missing ones).. I'll try to update this one to give more information.

rphillipsz: On the other hand.. the logging descriptions in general are in the other ticket.  Does this apply to both glog and the package logger? Would it be better to leave this until we've updated the descriptions on logging completely?

iancmcc:
It does indeed apply to both, but you're right, don't worry about it if you
don't have to. I should've looked at just the 2nd commit like you said :)

PR #3162, opened 2016-10-28
Fixes for logstash configuration on the master

tgjonestx:
Fixes CC-2875

Several changes here:
1. Add locking to prevent possible races with different template requests trying to update the configuration at the same time
2. On startup and on RestoreServiceTemplate, wait till all templates have been processed so that the logstash isvcs is only restarted once
3. On startup, always build a new file, but only restart logstash if the contents actually change. 

The last item covers the original problem reported in CC-2875 because when a new version of CC is installed, we overwrite the existing `logstash.conf` with a new version that has an empty filter section. If the customer didn't install anything else, then the config file would never get updated because the startup process would not find any new templates (though, if an RM rpm was also installed at the same time, then the system would heal itself). Now it rebuilds the config file in the background on every startup, whether there are new RM templates or not, and only updates the config if the file contents change


facade/logstash.go
```diff
@@ -11,26 +11,73 @@
...
+	}
+	err = writeLogstashConfiguration(templates)
+	if err == ErrLogstashUnchanged {
+		return nil
+	} else if err != nil {
+		glog.Errorf("Could not write logstash configuration: %s", err)
+		return err
+	} 
+
+	glog.V(2).Infof("Starting logstash container")
```
ecifreo: Is this the "old style" logging, or the one that should be used going forward?

tgjonestx: That is old school. As a lot of this PR is copy/paste from older code, I didn't take the time to retrofit the logging.   Moving forward, the plan is to systematically retrofit existing code to fit the standards and remove glog from our tool chain in favor of logrus and logri (the hierarchical logging extension for logrus written by Ian).

PR #3163, opened 2016-10-28
CC-2869: Fix flare message

cchristiezen:
When the requests take longer than the polling interval on the Pools page, the UI now sends a warning to the console.

### DIFF
--- a/web/ui/src/Pools/PoolsController.js
+++ b/web/ui/src/Pools/PoolsController.js
@@ -67,7 +67,7 @@
                     deferred.resolve();
                 })
                 .error(data => {
-                    let message = data.Detail || data || "";
+                    let message = (data && data.Detail) || "";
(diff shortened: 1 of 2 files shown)


web/ui/src/Pools/PoolsController.js
```diff
@@ -67,7 +67,8 @@
                     deferred.resolve();
                 })
                 .error(data => {
-                    $notification.create("Unable to load pools.", data.Detail).error();
+                    let message = data.Detail || data || "";
```
rDr4g0n: if `data` is undefined here, trying to access `data.Detail` is going to throw an error. you may want to change this to match the other fix in this pr: ` let message = (data && data.Detail) || "";`

cchristiezen:
retest this

PR #3165, opened 2016-10-28
CC-2850 Instance not restarted if container dies during wan outtage

johnlhamilton:
Fixed a race that prevented the zk listener from restarting when a WAN outtage ends.  Now, if a container dies during a WAN outtage, it will be restarted when the WAN comes back.

### DIFF
--- a/node/agent.go
+++ b/node/agent.go
@@ -79,8 +79,8 @@ type HostAgent struct {
 	mount                []string             // each element is in the form: dockerImage,hostPath,containerPath
 	currentServices      map[string]*exec.Cmd // the current running services
 	mux                  *proxy.TCPMux
-	muxport              string               // the mux port to serviced (default is 22250)
-	useTLS               bool                 // true if TLS should be enabled for MUX
+	muxport              string // the mux port to serviced (default is 22250)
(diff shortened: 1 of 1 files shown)


node/agent.go
```diff
@@ -407,8 +407,8 @@ func (a *HostAgent) Start(shutdown <-chan interface{}) {
 		case <-startExit:
 			glog.Infof("Host Agent restarting")
 			close(unregister)
-			unregister = make(chan interface{})
 			rwg.Wait()
+			unregister = make(chan interface{})
```
smousa: why is moving this the source of contention?

smousa: Could we be covering up another race condition?

johnlhamilton: We close unregister channel and then wait for the goroutine on line 366 to exit.  If we reassign the unregister variable to a fresh channel at the wrong time, the goroutine on 366 never exits. If we happen to close and then reassign unregister while the RegisterHost call on line 371 is executing, AND that call returns an error, then we will loop back into 371 passing in an unclosed channel that will never get closed.

smousa: Can you add a comment, so someone else doesn't accidently misplace it?

PR #3170, opened 2016-10-31
CC-2898: services wont start if the service state node is deleted fro…

fraibacas:
…m zk

### DIFF
--- a/zzk/service/hostregistry.go
+++ b/zzk/service/hostregistry.go
@@ -178,6 +178,15 @@ func (h *HostRegistryListener) Spawn(cancel <-chan interface{}, hostid string) {
 		}
 		isRunning := len(ch) > 0
 
+		eventLogger := plog.WithFields(log.Fields{
+			"poolid":    h.poolid,
+			"hostid":    hostid,
+			"isOnline":  isOnline,
+			"isLocked":  isLocked,
+			"isRunning": isRunning,
+		})
+		eventLogger.Debug("Waiting for host event")
+
 		if isOnline {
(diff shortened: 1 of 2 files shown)


zzk/service/hoststate.go
```diff
@@ -185,6 +185,16 @@ func (l *HostStateListener) Spawn(shutdown <-chan interface{}, stateID string) {
 			logger.WithError(err).Error("Could not watch host state")
 			return
 		}
+		// set up a listener on the service state node
+		ssdat = &ServiceState{}
+		ssevt, err := l.conn.GetW(sspth, ssdat, done)
```
fraibacas: Can I use the same "done" channel?

smousa: yes

smousa: use ExistsW, instead of GetW because, we only want an event if the node is deleted, not if it is updated.

zzk/service/hostregistry.go
```diff
@@ -178,6 +178,15 @@ func (h *HostRegistryListener) Spawn(cancel <-chan interface{}, hostid string) {
 		}
 		isRunning := len(ch) > 0
 
+		eventLogger := plog.WithFields(log.Fields{
+			"poolid":    h.poolid,
+			"hostid":    hostid,
+			"isOnline":  isOnline,
+			"isLocked":  isLocked,
+			"isRunning": isRunning,
```
smousa: use lowercase defining field names for log messages.

PR #3177, opened 2016-10-31
updates to make acceptance tests run on xenial

rDr4g0n:


rDr4g0n:
still needs some fixin

PR #3178, opened 2016-10-31
Feature/cc 2903

rphillipsz:


acceptance/ui/features/pages/hosts.rb
```diff
@@ -7,7 +7,7 @@ class Hosts < SitePrism::Page
 
     section :navbar, NavBarSection, ".navbar-collapse"
 
-    element :addHost_button, ".add-host-button"
+    element :addHost_button, "[ng-click=\"hostsVM.clickAddHost()\"]"
```
rDr4g0n: `.add-host-button` is a better solution than matching on something like angular function binding. Also, that change was coupled with an html change: https://github.com/control-center/serviced/commit/3cec80c73d080289c4d4b4912cbd3f865cee0f5d

PR #3189, opened 2016-11-03
Feature/acceptance fix6

rDr4g0n:
deploywiz needs templates, pools, and hosts to do its work, but doesnt always wait for all 3 to load up. This moves all those requests to one place and waits till all 3 are done before it updates the wizard's state.


web/ui/src/DeployWiz/DeployWizard.js
```diff
@@ -28,15 +28,6 @@
             return !$scope.dockerLoggedIn;
         };
```
rDr4g0n: the following get moved into an `init` function

PR #3190, opened 2016-11-03
delete old backup directory when recovering es logstash isvc

smousa:


isvcs/es.go
```diff
@@ -157,7 +157,12 @@ func initElasticSearch() {
 }
 
 func recoverES(path string) error {
-	return os.Rename(path, path+"-backup")
+	recoveryPath := path + "-backup"
+
+	if _, err := os.Stat(recoveryPath); err == nil {
+		os.RemoveAll(recoveryPath)
+	}
+	return os.Rename(path, recoveryPath)
```
tgjonestx: Seems like we should have some logging here (or the caller) to tell the administrator what happened.

smousa: there is logging (see ticket)

PR #3191, opened 2016-11-03
Return a smaller struct for calls to GetServicesHealth

ecifreo:


domain/service/servicehealth.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func BuildServiceHealth(svc Service) (*ServiceHealth) {
+	svch := &ServiceHealth{
+		ID: svc.ID,
+		Name: svc.Name,
+		PoolID: svc.PoolID,
+		Instances: svc.Instances,
+		DesiredState: svc.DesiredState,
+		HealthChecks: svc.HealthChecks,
```
tgjonestx: This is copying the reference to the map, rather than making a copy the map itself.  To be safe, the code should be something like:
```
for key, value := range svc.HealthChecks {
    svch.HealthChecks[key] = value
}
```

domain/service/servicehealth_store.go
```diff
@@ -0,0 +1,124 @@
...
+import (
+	"strings"
+	"errors"
+
+	"github.com/control-center/serviced/datastore"
+	"github.com/control-center/serviced/datastore/elastic"
+)
+
+func (s *storeImpl) GetServiceHealth(ctx datastore.Context, svcId string) (*ServiceHealth, error) {
+	defer ctx.Metrics().Stop(ctx.Metrics().Start("storeImpl.GetServiceHealth"))
```
tgjonestx: Please change `"storeImpl.GetServiceHealth"` to `"ServiceStore.GetServiceHealth"` to be more consistent with the other metrics on storeImpl

domain/service/servicehealth_store.go
```diff
@@ -0,0 +1,124 @@
...
+		err := results.Next(&sh)
+		if err != nil {
+			return nil, err
+		}
+		s.fillHealthVolatileInfo(&sh)
+		health = append(health, sh)
+	}
+
+	// no errors but also no results for given id
+	return nil, nil
```
tgjonestx: The comment and the return is wrong; needs to be
```
return health, nil
```

domain/service/servicehealth_store.go
```diff
@@ -0,0 +1,124 @@
...
+		var health ServiceHealth
+		err = results.Next(&health)
+		if err != nil {
+			return nil, err
+		}
+		s.fillHealthVolatileInfo(&health)
+		return &health, nil
+	}
+
+	// no errors but also no results for given id
```
tgjonestx: For consistency with the similar methods on this class, this case should return `ErrNoSuchEntity`. See the latest version of `GetServiceDetails()` on develop for an example.

PR #3197, opened 2016-11-03
Replace client-side service search with a backend version

iancmcc:


domain/service/servicedetails_store.go
```diff
@@ -142,6 +143,56 @@ func (s *storeImpl) GetServiceDetailsByParentID(ctx datastore.Context, parentID
...
+		if err != nil {
+			return nil, err
+		}
+
+		if hasChildren, err := s.hasChildren(ctx, d.ID); err == nil {
+			d.HasChildren = hasChildren
+		} else {
+			return nil, err
+		}
+
```
tgjonestx: Need to call `s.fillDetailsVolatileInfo(&d)` before append() to get the correct value of DesiredState

domain/service/servicedetails_store.go
```diff
@@ -142,6 +143,56 @@ func (s *storeImpl) GetServiceDetailsByParentID(ctx datastore.Context, parentID
 	return details, nil
 }
 
+// GetServiceDetailsByIDOrName returns the service details for any services
+// whose serviceID matches the query exactly or whose names contain the query
+// as a substring
+func GetServiceDetailsByIDOrName(ctx datastore.Context, query string) ([]ServiceDetails, error) {
```
tgjonestx: Please add the following so that we can measure this method:
```
defer ctx.Metrics().Stop(ctx.Metrics().Start("ServiceStore. GetServiceDetailsByIDOrName"))
```

cli/cmd/service.go
```diff
@@ -697,108 +647,41 @@ func cmdSetTreeCharset(ctx *cli.Context, config utils.ConfigReader) {
...
 
 	// more than one match, display a dialog
-	var svcpath func(string) string
-	svcpath = func(serviceID string) string {
-		svc := svcmap[serviceID]
-		if svc.ParentServiceID == "" {
-			return path.Join(svc.DeploymentID, svc.Name)
-		} else {
-			return path.Join(svcpath(svc.ParentServiceID), svc.Name)
+	var svcpath func(*service.ServiceDetails) string
```
smousa: Does this still work?

facade/service.go
```diff
@@ -2140,3 +2141,72 @@ func (f *Facade) CountDescendantStates(ctx datastore.Context, serviceID string)
 	}, "descendantStatus")
 	return result, nil
 }
+
+// ResolveServicePath resolves a service path (e.g., "infrastructure/mariadb")
+// to zero or more service details with their ancestry populated.
+func (f *Facade) ResolveServicePath(ctx datastore.Context, svcPath string) ([]service.ServiceDetails, error) {
```
smousa: logging?

smousa: Also, I think you need to include the metrics thing: @tgjonestx am I right?

cli/cmd/service.go
iancmcc: Yes.

PR #3206, opened 2016-11-04
More scheduling optimization

tgjonestx:
More optimizations in support of CC-2509. The two main themes (primarily in c0d78ea)

* do not pass arrays of Service objs; pass arrays of pointers to Service objs
* When updating ZK for manual scheduling, group the updates by pool, so we only have to check the ZK path "/pool/<poolName>/services" once for a given pool instead of once for each service (e.g. replaces 2000 of those checks with 100 checks in the 100-pools scenario).

### DIFF
--- a/domain/service/servicehealth.go
+++ b/domain/service/servicehealth.go
@@ -54,7 +54,6 @@ func BuildServiceHealth(svc Service) *ServiceHealth {
 		HealthChecks: make(map[string]health.HealthCheck),
 	}
 
-	sh.HealthChecks = make(map[string]health.HealthCheck)
 	for key, value := range svc.HealthChecks {
 		sh.HealthChecks[key] = value
 	}
--- a/domain/service/servicehealth.go
+++ b/domain/service/servicehealth.go
@@ -54,6 +54,7 @@ func BuildServiceHealth(svc Service) *ServiceHealth {
 		HealthChecks: make(map[string]health.HealthCheck),
 	}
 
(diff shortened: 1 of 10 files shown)


facade/zkapi.go
```diff
@@ -66,33 +71,52 @@ func zkr_SyncServiceRegistry(ctx datastore.Context, conn client.Connection, requ
...
+// synced in zookeeper
+func (zk *zkf) UpdateServices(ctx datastore.Context, tenantID string, svcs []*service.Service, setLockOnCreate, setLockOnUpdate bool) error {
+	defer ctx.Metrics().Stop(ctx.Metrics().Start("zzk.UpdateServices"))
 
-	// get the pool-based connection to update the service
-	poolconn, err := getLocalConnection(ctx, path.Join("/pools", svc.PoolID))
-	if err != nil {
-		logger.WithError(err).Debug("Could not acquire a pool-based connection to update the service in zookeeper")
-		return err
+	servicesByPool := make(map[string]*[]*service.Service)
```
iancmcc: Is the pointer to the array necessary?

domain/service/servicehealth.go
```diff
@@ -54,6 +54,7 @@ func BuildServiceHealth(svc Service) *ServiceHealth {
 		HealthChecks: make(map[string]health.HealthCheck),
 	}
 
+	sh.HealthChecks = make(map[string]health.HealthCheck)
```
tgjonestx: It was necessary before I merged your recent changes :-). Will remove.

PR #3221, opened 2016-11-07
Feature/acceptance fix10

rDr4g0n:
improves stability of acceptance tests

* bumps versions of all ruby gems
* removes unneeded phantomjs/poltergeist and capybara-webkit deps
* puts a test that doesnt always clean up well at the end of a scenario (so it doesnt affect the others)
* uses a different method to fill out `input`s because the other method is unreliable
* puts all these changes into 1.1.3 version of the capybara image


rDr4g0n:
retest this

PR #3223, opened 2016-11-08
Reduce backup/restore polling frequency to 1/sec

ecifreo:


tgjonestx:
I think one second is still too fast. We could easily get by with something like once every 3-5 seconds

PR #3233, opened 2016-11-10
CC-2983 export endpoints in service definitions for collector services cause "duplicate endpoint found" if more than one collector is defined at install time

johnlhamilton:
Evaluate templates before validating endpoints

Fixes [CC-2983](https://jira.zenoss.com/browse/CC-2983)

### DIFF
--- a/facade/service.go
+++ b/facade/service.go
@@ -623,7 +623,6 @@ func (f *Facade) validateServiceDeployment(ctx datastore.Context, parentID strin
 		}
 		// Evaluate templated endpoints
 		if err = f.evaluateEndpointTemplates(ctx, svc); err != nil {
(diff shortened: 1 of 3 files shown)


facade/service.go
```diff
@@ -621,6 +621,12 @@ func (f *Facade) validateServiceDeployment(ctx datastore.Context, parentID strin
 			glog.Errorf("Could not create service %s: %s", sd.Name, err)
 			return nil, err
 		}
+		// Evaluate templated endpoints
+		if err = f.evaluateEndpointTemplates(ctx, svc); err != nil {
+			glog.Errorf("Could not evaluate endpoint templates for service %s with parent %s: %s", svc.Name, svc.ParentServiceID, err)
```
jafcrocker: Per the [logging guidelines](https://sites.google.com/a/zenoss.com/engineering/home/faq/logging-guidelines), we should not both log and return an error.

PR #3241, opened 2016-11-11
Fix pending deployment status reporting

jafcrocker:
This change addresses some code debt in the existing _pending deployment status reporting_ approach.
There were several problems that were addressed:
1) The use of ad-hoc mutexes made the code brittle and difficult to reason about.
2) The storing of "lastStatus" in the per-deployment map meant that querying from multiple clients would have incorrect results.  This was not critical in the current application, but having an efficient pattern for this is desirable for other domains (e.g., backup status)
3) Storing the per-deployment data in a global variable was not amenable to unit testing.
4) The busy-wait loop checking for changed status is inefficient and inelegant, given the robust synchronization techniques in go.

The new approach is for the status query to take two parameters - a timeout, and the last status reported.  If the current status is the same as the last status, the server will wait for the status to change up to the timeout. 

Fixes [CC-3076](https://jira.zenoss.com/browse/CC-3076)

### DIFF
--- a/facade/pendingdeployment.go
+++ b/facade/pendingdeployment.go
@@ -1,0 +1,122 @@
+// Copyright 2014 The Serviced Authors.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package facade
+
+import (
+	"errors"
+	"sync"
+
+	"github.com/control-center/serviced/utils"
+)
+
+var (
+	ErrPendingDeploymentConflict = errors.New("template deployment id conflict")
+)
+
+// PendingDeployment represents a template deployment which is has been initiated
+// but has not yet completed.  It is synchronized, allowing access from multiple
+// goroutines.  In particular clients can register for asynchronous notification of
+// changes to its "status".
+type PendingDeployment struct {
+	status       utils.ValueChangePublisher
+	templateID   string
+	poolID       string
+	deploymentID string
+	templateName string
+	mutex        sync.RWMutex
+}
+
+// NewPendingDeployment returns a new PendingDeployment object.
+func NewPendingDeployment(deploymentID, templateID, poolID string) PendingDeployment {
+	return PendingDeployment{
+		status:       utils.NewValueChangePublisher("Starting"),
+		templateID:   templateID,
+		deploymentID: deploymentID,
+		poolID:       poolID,
+		mutex:        sync.RWMutex{},
+	}
+}
+
+// UpdateStatus updates the status of a PendingDeployment.
+func (d *PendingDeployment) UpdateStatus(status string) {
+	d.mutex.Lock()
+	defer d.mutex.Unlock()
+	d.status.Set(status)
+}
+
+// SetTemplateName sets the template name of a PendingDeployment.
+func (d *PendingDeployment) SetTemplateName(name string) {
+	d.mutex.Lock()
+	defer d.mutex.Unlock()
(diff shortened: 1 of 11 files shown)


facade/pendingdeployment.go
```diff
@@ -0,0 +1,122 @@
...
+	return &PendingDeploymentMgr{
+		deployments: make(map[string]*PendingDeployment),
+		mutex:       sync.RWMutex{},
+	}
+}
+
+// NewPendingDeployment allocates a new PendingDeployment with the given attributes
+// and associates it with the PendingDeploymentMgr
+func (dm *PendingDeploymentMgr) NewPendingDeployment(deploymentID, templateID, poolID string) *PendingDeployment {
+	deployment := NewPendingDeployment(deploymentID, templateID, poolID)
```
smousa: Should this be after the mutex, to prevent a race condition of two threads overwriting the same pending deployment key?

facade/pendingdeployment.go
```diff
@@ -0,0 +1,122 @@
...
+	return dm.deployments[deploymentID]
+}
+
+// DeletePendingDeployment removes the PendingDeployment associated with the
+// given deploymentID from the PendingDeploymentMgr.
+func (dm *PendingDeploymentMgr) DeletePendingDeployment(deploymentID string) {
+	dm.mutex.Lock()
+	delete(dm.deployments, deploymentID)
+	dm.mutex.Unlock()
+}
```
smousa: Please add unit tests

facade/servicetemplate.go
```diff
@@ -357,19 +335,19 @@ func (f *Facade) DeployService(ctx datastore.Context, poolID, parentID string, o
 		}
 	}
 
-	return f.deployService(ctx, tenantID, svc.ID, svc.DeploymentID, poolID, overwrite, svcDef)
+	var statusUpdater = func(status string) {}
```
smousa: Is this supposed to be a noop?

utils/valuechangepublisher.go
```diff
@@ -0,0 +1,54 @@
...
+		value:  initialValue,
+		mutex:  sync.RWMutex{},
+		notify: make(chan interface{}),
+	}
+}
+
+// Set closes the current channel notifying current subscribers of a change
+// and stores the value.
+func (v *ValueChangePublisher) Set(value interface{}) {
+	if value == v.value {
```
smousa: this might be a race condition.  You may want to RLock this

jafcrocker: Yes.  Changing...

facade/pendingdeployment.go
```diff
@@ -0,0 +1,122 @@
...
+// SetTemplateName sets the template name of a PendingDeployment.
+func (d *PendingDeployment) SetTemplateName(name string) {
+	d.mutex.Lock()
+	defer d.mutex.Unlock()
+	d.templateName = name
+}
+
+// GetStatus returns the current status of a PendingDeployment and a
+// channel which will be closewhen the status changes.
+func (d *PendingDeployment) GetStatus() (string, <-chan interface{}) {
```
jafcrocker: Yes, changed.

facade/pendingdeployment.go
```diff
@@ -0,0 +1,122 @@
...
+
+package facade
+
+import (
+	"sync"
+
+	"github.com/control-center/serviced/utils"
+)
+
+// Pending deployment represents a template deployment which is has been initiated
```
jafcrocker: Changed

facade/servicetemplate.go
jafcrocker: Yes.  The facade.DeployService routine is about adding a single service to an existing service tree. We don't monitor the status of those operations, so there's no need to update the status.

Actually, I don't _think_ we use DeployService any more.  We could possibly remove it.  It was part of the old zenpack installation process, but now we use MigrateService via the service migration library.

smousa: I believe we do call DeployService when we are creating services as a result of installing a zenpack.  MigrateService and UpdateService does not handle service creation.

jafcrocker:
retest this

facade/pendingdeployment.go
jafcrocker: Per your suggestion, I made NewPendingDeployment return an error in the case of a duplicate template id

PR #3245, opened 2016-11-17
Escape endpoint, vhost name in on validation error

glujan:
[CC-2671](https://jira.zenoss.com/browse/CC-2671)


glujan:
I'm not sure if I need additional function to sanitize input or I can use Angular to do this

PR #3248, opened 2016-11-17
CC-3010 Remove app-level health check icon

ed-:


ed-:
Hold off on this one until after 1.2.1 is cut

ed-:
This also needs changes to acceptance tests

PR #3255, opened 2016-11-28
Don't shut down containers when host state listener shuts down

johnlhamilton:
Fixes CC-3021

### DIFF
--- a/node/agent.go
+++ b/node/agent.go
@@ -394,7 +394,7 @@ func (a *HostAgent) Start(shutdown <-chan interface{}) {
 		// 1) has a connection
 		// 2) its node is registered
 		// 3) receives signal to shutdown or breaks
-		hsListener := zkservice.NewHostStateListener(a, a.hostID)
+		hsListener := zkservice.NewHostStateListener(a, a.hostID, shutdown)
 
 		startExit := make(chan struct{})
 		go func() {
--- a/node/agent.go
+++ b/node/agent.go
@@ -347,6 +347,9 @@ func (a *HostAgent) Start(shutdown <-chan interface{}) {
 	unregister := make(chan interface{})
 	stop := make(chan interface{})
 
+	// Create the host state listener here, so it keeps its state
+	hsListener := zkservice.NewHostStateListener(a, a.hostID, shutdown)
+
 	for {
 		// handle shutdown if we are waiting for a zk connection
 		var conn coordclient.Connection
@@ -394,8 +397,6 @@ func (a *HostAgent) Start(shutdown <-chan interface{}) {
 		// 1) has a connection
 		// 2) its node is registered
 		// 3) receives signal to shutdown or breaks
-		hsListener := zkservice.NewHostStateListener(a, a.hostID, shutdown)
-
 		startExit := make(chan struct{})
 		go func() {
 			defer close(startExit)
--- a/zzk/service/hoststate.go
+++ b/zzk/service/hoststate.go
@@ -207,97 +207,7 @@ func (l *HostStateListener) Spawn(cancel <-chan interface{}, stateID string) {
 		}
 
 		// set the state of this instance
-		setInstanceState := func() bool {
-			l.mu.Lock()
-			defer l.mu.Unlock()
-
-			// Don't do anything if we are shutting down
-			select {
-			case <-l.shutdown:
-				return false
-			default:
-			}
-			// attach to the container if not already attached
-			if containerExit == nil {
-				containerExit, err = l.handler.AttachContainer(ssdat, serviceID, instanceID)
-				if err != nil {
-					logger.WithError(err).Error("Could not attach to container")
-					l.cleanUpContainer(stateID)
-					return false
-				}
-				l.setExistingState(stateID, ssdat, containerExit)
-			}
-
-			switch hsdat.DesiredState {
-			case service.SVCRun:
-				if containerExit == nil {
-					// container is detached because it doesn't exist
-					ssdat, containerExit, err = l.handler.StartContainer(l.shutdown, serviceID, instanceID)
-					if err != nil {
-						logger.WithError(err).Error("Could not start container")
-						l.cleanUpContainer(stateID)
-						return false
-					}
-
-					// set the service state
-					l.setExistingState(stateID, ssdat, containerExit)
-					logger.Debug("Started container")
-
-					if err := l.updateServiceStateInZK(ssdat, req); err != nil {
-						logger.WithError(err).Error("Could not set state for started container")
-						return false
-					}
-
-					return true
-				} else if ssdat.Paused {
-					// resume paused container
-					if err := l.handler.ResumeContainer(serviceID, instanceID); err != nil {
-						logger.WithError(err).Error("Could not resume container")
-						l.cleanUpContainer(stateID)
-						return false
-					}
-
-					// update the service state
-					ssdat.Paused = false
-					l.setExistingState(stateID, ssdat, containerExit)
-
-					if err := l.updateServiceStateInZK(ssdat, req); err != nil {
-						logger.WithError(err).Error("Could not set state for resumed container")
-						return false
-					}
-
-					logger.Debug("Resumed paused container")
-				}
-			case service.SVCPause:
-				if containerExit != nil && !ssdat.Paused {
-					// container is attached and not paused, so pause the container
-					if err := l.handler.PauseContainer(serviceID, instanceID); err != nil {
-						logger.WithError(err).Error("Could not pause container")
-						l.cleanUpContainer(stateID)
-						return false
-					}
-
-					// update the service state
-					ssdat.Paused = true
-					l.setExistingState(stateID, ssdat, containerExit)
-					if err := l.updateServiceStateInZK(ssdat, req); err != nil {
-						logger.WithError(err).Error("Could not set state for resumed container")
-						return false
-					}
-
-					logger.Debug("Paused running container")
-				}
-			case service.SVCStop:
-				// shut down the container and clean up nodes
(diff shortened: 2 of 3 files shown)


zzk/service/hoststate.go
```diff
@@ -50,19 +51,32 @@ type HostStateHandler interface {
 	PauseContainer(serviceID string, instanceID int) error
 }
 
+type HostStateThread struct {
```
smousa: This should be private, you may even want to just embed this into HostStateListener, i.e.
```
threads map[string]struct{
data *ServiceState
exited <-chan time.Time
}
```

zzk/service/hoststate.go
```diff
@@ -79,20 +93,38 @@ func (l *HostStateListener) Ready() error {
...
-// Done removes the ephemeral node from the host registry
+// Done implements zzk.Listener
 func (l *HostStateListener) Done() {
 }
 
 // PostProcess implements zzk.Listener
-func (l *HostStateListener) PostProcess(p map[string]struct{}) {}
+// This is always called after all threads have been spawned
+func (l *HostStateListener) PostProcess(p map[string]struct{}) {
+	// We are running all of the containers we are supposed to, now
```
smousa: Should this function be under a single lock?  Seems like you can hit some map inconsistencies

zzk/service/hoststate.go
```diff
@@ -79,20 +93,38 @@ func (l *HostStateListener) Ready() error {
...
+		}
+	}
+}
 
 // Spawn listens for changes in the host state and manages running instances
-func (l *HostStateListener) Spawn(shutdown <-chan interface{}, stateID string) {
+func (l *HostStateListener) Spawn(cancel <-chan interface{}, stateID string) {
 	logger := plog.WithFields(log.Fields{
-		"hostid":  l.hostID,
+		"hostID":  l.hostID,
```
smousa: needs to be in all lowercase

zzk/service/hoststate.go
```diff
@@ -119,192 +151,264 @@ func (l *HostStateListener) Spawn(shutdown <-chan interface{}, stateID string) {
...
-		if err := l.handler.StopContainer(serviceID, instanceID); err != nil {
-			logger.WithError(err).Error("Could not stop container")
-		} else if containerExit != nil {
-			// wait for the container to exit
-			time := <-containerExit
-			logger.WithField("terminated", time).Debug("Container exited")
-		}
+	// reattach to orphaned container
+	ssdat, containerExit := l.getExistingState(stateID)
+	// TODO: Should we go ahead and pull the service state node from zk and make sure it matches?
```
smousa: No, because this is the only place the service state node will ever be modified.

zzk/service/hoststate.go
```diff
@@ -119,192 +151,264 @@ func (l *HostStateListener) Spawn(shutdown <-chan interface{}, stateID string) {
...
 
 		// attach to the container if not already attached
 		if containerExit == nil {
 			containerExit, err = l.handler.AttachContainer(ssdat, serviceID, instanceID)
 			if err != nil {
-
 				logger.WithError(err).Error("Could not attach to container")
+				// TODO: there is a problem with docker here. How do we handle
+				// this behavior?  For now just shut down and clean up nodes
+				l.cleanUpContainer(stateID)
```
smousa: I think it is correct to shut down and clean up in event of a problem with docker.

node/agent.go
```diff
@@ -394,7 +394,7 @@ func (a *HostAgent) Start(shutdown <-chan interface{}) {
 		// 1) has a connection
 		// 2) its node is registered
 		// 3) receives signal to shutdown or breaks
-		hsListener := zkservice.NewHostStateListener(a, a.hostID)
+		hsListener := zkservice.NewHostStateListener(a, a.hostID, shutdown)
```
jplouis: This needs to be outside of the for loop to keep state between any zzk failures.

zzk/service/hoststate.go
```diff
@@ -119,192 +151,264 @@ func (l *HostStateListener) Spawn(shutdown <-chan interface{}, stateID string) {
...
+		InstanceID: instanceID,
+	}
+	if err := DeleteState(l.conn, req); err != nil {
+		logger.WithError(err).Warn("Could not delete state for stopped container")
+		return
+	}
+}
+
+func (l *HostStateListener) watchForShutdown() {
+	<-l.shutdown
```
smousa: This may cause some contention if there is a Spawn managing the node

zzk/service/hoststate.go
```diff
@@ -119,192 +151,264 @@ func (l *HostStateListener) Spawn(shutdown <-chan interface{}, stateID string) {
...
+func (l *HostStateListener) updateServiceStateInZK(data *ServiceState, req StateRequest) error {
+	return UpdateState(l.conn, req, func(s *State) bool {
+		s.ServiceState = *data
+		return true
+	})
+}
+
+func (l *HostStateListener) cleanUpContainer(stateID string) {
+	logger := plog.WithFields(log.Fields{
+		"hostID":  l.hostID,
```
smousa: lowercase

zzk/service/hoststate.go
```diff
@@ -119,192 +151,264 @@ func (l *HostStateListener) Spawn(shutdown <-chan interface{}, stateID string) {
...
+}
+
+func (l *HostStateListener) updateServiceStateInZK(data *ServiceState, req StateRequest) error {
+	return UpdateState(l.conn, req, func(s *State) bool {
+		s.ServiceState = *data
+		return true
+	})
+}
+
+func (l *HostStateListener) cleanUpContainer(stateID string) {
```
smousa: I feel like cleanup should retain the lock from the beginning of the call

zzk/service/hoststate.go
jplouis: You can make it package private if you want but I like the explicit struct.

zzk/service/hoststate.go
```diff
@@ -79,20 +94,40 @@ func (l *HostStateListener) Ready() error {
...
+// This is always called after all threads have been spawned
+func (l *HostStateListener) PostProcess(p map[string]struct{}) {
+	// We are running all of the containers we are supposed to, now
+	// shut down any containers we are not supposed to be running
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	stateIDs := l.getExistingStateIDs()
+	for _, s := range stateIDs {
+		if _, ok := p[s]; !ok {
+			l.cleanUpContainer(s)
```
jplouis: What about some logging for this scenario?

smousa:
I think we only want to cache threads that are detached.  I don't think we should store active thread data.

zzk/service/hoststate.go
```diff
@@ -107,13 +107,21 @@ func (l *HostStateListener) PostProcess(p map[string]struct{}) {
 	// shut down any containers we are not supposed to be running
 	l.mu.Lock()
 	defer l.mu.Unlock()
-	stateIDs := l.getExistingStateIDs()
+	stateIDs := l.getExistingThreadStateIDs()
+	orphanedStates := make([]string, len(stateIDs))
```
smousa: you shouldn't initialize this with a size--append instead

zzk/service/hoststate.go
```diff
@@ -243,24 +245,23 @@ func (l *HostStateListener) setInstanceState(containerExit <-chan time.Time, ssd
...
 	if containerExit == nil {
 		containerExit, err = l.handler.AttachContainer(ssdat, serviceID, instanceID)
 		if err != nil {
 			logger.WithError(err).Error("Could not attach to container")
-			l.cleanUpContainer(stateID)
+			l.cleanUpContainers([]string{stateID}, true)
 			return nil, false
 		}
-		l.setExistingState(stateID, ssdat, containerExit)
+		l.mu.Lock()
```
smousa: You may want to consider wrapping this in function.

zzk/service/hoststate.go
```diff
@@ -270,12 +271,20 @@ func (l *HostStateListener) setInstanceState(containerExit <-chan time.Time, ssd
...
 			if err != nil {
 				logger.WithError(err).Error("Could not start container")
-				l.cleanUpContainer(stateID)
+				l.cleanUpContainers([]string{stateID}, true)
 				return nil, false
 			}
 
 			// set the service state
-			l.setExistingState(stateID, ssdat, containerExit)
+			l.mu.Lock()
```
smousa: This may need to be wrapped in a function

zzk/service/hoststate.go
```diff
@@ -287,13 +296,20 @@ func (l *HostStateListener) setInstanceState(containerExit <-chan time.Time, ssd
...
 				logger.WithError(err).Error("Could not resume container")
-				l.cleanUpContainer(stateID)
+				l.cleanUpContainers([]string{stateID}, true)
 				return nil, false
 			}
 
 			// update the service state
 			ssdat.Paused = false
-			l.setExistingState(stateID, ssdat, containerExit)
+			l.mu.Lock()
```
smousa: same ^^

zzk/service/hoststate.go
```diff
@@ -307,13 +323,21 @@ func (l *HostStateListener) setInstanceState(containerExit <-chan time.Time, ssd
...
 				logger.WithError(err).Error("Could not pause container")
-				l.cleanUpContainer(stateID)
+				l.cleanUpContainers([]string{stateID}, true)
 				return nil, false
 			}
 
 			// update the service state
 			ssdat.Paused = true
-			l.setExistingState(stateID, ssdat, containerExit)
+			l.mu.Lock()
```
smousa: same ^^

PR #3259, opened 2016-11-28
Add new env variable for configuring timeout for GetRunningServices

ecifreo:
RPC call. It times out in large environments.

### DIFF
--- a/dao/client/cp_client.go
+++ b/dao/client/cp_client.go
@@ -30,6 +30,8 @@ package client
 // --------------------------------------------------------------------------------------------------
 
 import (
+	"os"
+	"strconv"
 	"time"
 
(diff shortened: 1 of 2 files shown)


dao/client/cp_client.go
```diff
@@ -58,6 +61,14 @@ func NewControlClient(addr string) (s *ControlClient, err error) {
 	s = new(ControlClient)
 	s.addr = addr
 	s.rpcClient = client
+
+	s.timeout = 10*time.Second
+	timeoutvalue := os.Getenv("SERVICED_GETSERVICES_TIMEOUT")
```
tgjonestx: Typically, we add options like this to the command line as well. Was there discussion about specifically leaving this one off of the command line args?

PR #3269, opened 2016-11-30
Fix acceptance test partial text fill-ins.

rphillipsz:
It's likely that the Angular bindings weren't complete when Capybara was filling in the fields, resulting in a race and occasional partially populated fields.  This combines the angular ng-if to not show the form until the binding is there with a capybara find() to wait for the form elements to ensure the binding is complete before entering the text fields.

ref: https://goo.gl/y58srz


tgjonestx:
👍

PR #3287, opened 2016-12-08
CC UI Resource Pools page generates too many requests and causes serviced cpu usage to go up

spopezen:
https://jira.zenoss.com/browse/CC-3025

The commits in this PR fix CC-3025 by creating/maintaining a cache to get the pools from instead of querying elastic.


spopezen:
Reopening when I have unit tests

PR #3288, opened 2016-12-09
CC-2365 Return more results from ElasticSearch

glujan:
As of now, it defaults to just 10 results

[Jira](https://jira.zenoss.com/browse/CC-2365)


glujan:
retest this

PR #3289, opened 2016-12-12
Make the call to /api/v2/pools less demanding

spopezen:
https://jira.zenoss.com/browse/CC-3025

This PR fixes CC-3025 by creating a pool cache to be used for ReadPools.

Due to the multiple ways poolCache can be invalidated and the cost of updating a stale poolCache, the best way I could think to manage this cache was by having a simple dirty flag that's set when the cache will be stale, and have a func to refresh the cache as a parameter to GetPools from cache, which is called when GetPools is called if the cache is dirty.

Unit tests were setup to verify that the cache is getting correctly invalidated and subsequently updated.

### DIFF
--- a/facade/pool_cache.go
+++ b/facade/pool_cache.go
@@ -1,0 +1,54 @@
+// Copyright 2016 The Serviced Authors.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


facade/pool_cache.go
```diff
@@ -0,0 +1,63 @@
...
+		pools: make([]pool.ReadPool, 0),
+	}
+}
+
+// GetPools caches the result of getPoolsFunc if the cache is dirty
+// then returns the cached ReadPools.
+func (pc *poolCache) GetPools(getPoolsFunc GetPoolsFunc) ([]pool.ReadPool, error) {
+
+	var err error
+	if pc.dirty {
```
smousa: this needs to be locked as well.

PR #3291, opened 2016-12-15
CC services endpoint returns MonitoringProfile data for top level ser…

fraibacas:
…vices
Put back the check we had in 1.1.x : https://github.com/control-center/serviced/blob/support/1.1.x/web/service_internal_metrics.go#L51
Dont return monitoring profile for services without a command


fraibacas:
retest this

PR #3296, opened 2016-12-19
Fix filesystem-used metric

jafcrocker:
Previously, the value stored in 'storage.filesystem.used.<TENANT_ID>' metric was calculated as the difference between the total bytes for the filesystem and the available bytes; however that is incorrect because it does not take into account blocks reserved for the super-user (by default 5%).

After this change the metric will contain the correct value as returned by 'df'.

This change is related to [CC-3106](https://jira.zenoss.com/browse/CC-3106)


jafcrocker:
After the change, notice that there is a roughly 5% gap between the _total_ and _used_ + _available_ metrics.
image

PR #3297, opened 2016-12-20
Handle SIGPIPE

youngklee:
Handle SIGPIPE so that restarting journald won't crash serviced


cli/api/daemon.go
```diff
@@ -346,8 +346,18 @@ func (d *daemon) run() (err error) {
 	}
 
 	signalC := make(chan os.Signal, 10)
-	signal.Notify(signalC, syscall.SIGINT, syscall.SIGTERM, syscall.SIGHUP)
-	sig := <-signalC
+	signal.Notify(signalC, syscall.SIGINT, syscall.SIGTERM, syscall.SIGHUP, syscall.SIGPIPE)
+
+    var sig os.Signal
```
fraibacas: The indentation seems off

PR #3301, opened 2016-12-29
CC-3148 Applications - No input filtering on the "RAM Requested" text…

romanbakaleyko:
… field


romanbakaleyko:
backporting https://github.com/control-center/serviced/pull/3253/files

PR #3303, opened 2016-12-29
CC-3096: Stop running services when emergency shutdown triggered

johnlhamilton:
Added a facade method that will stop a service and all child services in the order specified by their EmergencyShutdownLevel.  It will also set EmergencyShutdown to true for all affected services.

### DIFF
--- a/facade/facade.go
+++ b/facade/facade.go
@@ -80,7 +80,8 @@ type Facade struct {
 	hostRegistry  auth.HostExpirationRegistryInterface
 	deployments   *PendingDeploymentMgr
 
-	isvcsPath string
+	isvcsPath              string
+	serviceRunLevelTimeout time.Duration
 }
 
 func (f *Facade) SetZZK(zzk ZZK) { f.zzk = zzk }
@@ -123,3 +124,5 @@ func (f *Facade) SetHostExpirationRegistry(hostRegistry auth.HostExpirationRegis
 }
 
 func (f *Facade) SetDeploymentMgr(mgr *PendingDeploymentMgr) { f.deployments = mgr }
+
+func (f *Facade) SetServiceRunLevelTimeout(t time.Duration) { f.serviceRunLevelTimeout = t }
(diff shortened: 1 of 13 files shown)


facade/facade.go
```diff
@@ -80,7 +80,8 @@ type Facade struct {
 	hostRegistry  auth.HostExpirationRegistryInterface
 	deployments   *PendingDeploymentMgr
 
-	isvcsPath string
+	isvcsPath              string
+	serviceRunLevelTimeout time.Duration
```
jafcrocker: _No change requested_, but this parameter doesn't fit in the facade from a logical perspective.  To me that's an indication that the shutdown sequencing code does not belong in the facade, as we discussed earlier.

facade/service.go
```diff
@@ -1213,18 +1213,86 @@ func (f *Facade) scheduleService(ctx datastore.Context, tenantID, serviceID stri
...
+				nextBatch := []*service.Service{}
+				nextBatchIDs := []string{}
+				for _, svc := range svcs {
+					// Set EmergencyShutdown to true and update the database
+					svc.EmergencyShutdown = true
+					uerr := f.updateService(ctx, tenantID, *svc, false, false)
+					if uerr != nil {
+						errToReturn = uerr
+						logger.WithField("service", svc.ID).WithError(uerr).Error("Failed to update database with EmergencyShutdown")
+					}
```
jafcrocker: I think we should set the emergency shutdown (ES) flag on all services in a separate loop before we begin shutting down services.  Otherwise there is a (potentially persistent) race condition in which a top level service is in ES but its children are not.

facade/service.go
```diff
@@ -1213,18 +1213,86 @@ func (f *Facade) scheduleService(ctx datastore.Context, tenantID, serviceID stri
...
+
+					currentLevel := svc.EmergencyShutdownLevel
+					if currentLevel == previousLevel {
+						nextBatch = append(nextBatch, svc)
+						nextBatchIDs = append(nextBatchIDs, svc.ID)
+					} else {
+						// Schedule this batch
+						levelLogger := logger.WithField("level", previousLevel)
+						levelLogger.Info("Shutting down all services at current emergency shutdown level")
+						a, serr := scheduleServices(f, nextBatch, ctx, tenantID, serviceID, desiredState)
```
jafcrocker: The services in the batch with ES level 0 need to be ordered (and batched) by reverse StartLevel.

PR #3304, opened 2017-01-04
CC-2379: Implement least squares algo

iancmcc:


commons/statistics/ols.go
```diff
@@ -0,0 +1,28 @@
...
+
+	if n == 0 {
+		return
+	}
+
+	var sumx, sumy, sumxx, sumxy float64
+
+	for _, p := range series {
+		x, y := p.X(), p.Y()
+		sumx += x
```
jafcrocker: For numerical accuracy, we should subtract either the mean or the first value from the X and Y samples, then add those back in to _b_ before returning.

commons/statistics/ols.go
```diff
@@ -0,0 +1,28 @@
+package statistics
+
+type Point interface {
+	X() float64
+	Y() float64
+}
+
+func LeastSquares(series []Point) (m float64, b float64) {
```
jafcrocker: I suggest this function take a "Points" interface, rather than a slice of Points.  The X and Y functions would take an index as a parameter.

PR #3310, opened 2017-01-09
Fix 'no child processes' error matching

rphillipsz:
https://jira.zenoss.com/browse/CC-2768
https://zenoss.zendesk.com/agent/tickets/123608

### DIFF
--- a/utils/nfs.go
+++ b/utils/nfs.go
@@ -3,15 +3,18 @@ package utils
 import (
 	"fmt"
 	"os/exec"
+	"regexp"
 
 	"github.com/zenoss/glog"
 )
+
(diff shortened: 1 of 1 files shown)


utils/nfs.go
```diff
@@ -3,15 +3,18 @@ package utils
 import (
 	"fmt"
 	"os/exec"
+	"regexp"
 
 	"github.com/zenoss/glog"
 )
 
+var noChildProcessesRegex = regexp.MustCompile("^wait(id)?: no child processes$")
```
tgjonestx: What was the error message seen at the customer site?

tgjonestx: nvm - I found the two messages in CC-2768

PR #3311, opened 2017-01-10
IMP-30: Could not set configurations to service zenimpactstate

olst:
There are two options to resolve IMP-30:
1) completely removing the serviced warning for empty config files (after discussion with architects), or 
2) adding some meaningless content to the impact config file. e.g., whitespace, or a comment.

This PR is for 1)
OR
https://github.com/zenoss/ZenPacks.zenoss.Impact/pull/414 is for 2)

### DIFF
--- a/facade/serviceconfig.go
+++ b/facade/serviceconfig.go
@@ -15,6 +15,7 @@ package facade
 
 import (
 	"errors"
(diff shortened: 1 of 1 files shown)


olst:
retest this

PR #3312, opened 2017-01-10
fixes CC-3147: Resource Pools - No validation on error messages on th…

HIMix:
…e interface text input
original fix: #3260


HIMix:
This should be rechecked and retested after the port for CC-2671 will be merged to support branch.

HIMix:
retest this please

PR #3330, opened 2017-01-18
ZEN-25802 zenossdbpack logging not captured

ssOlexBaiko:
fix update the cron command to redirect output to a log file that is located under /opt/serviced/log/zenossdbpack at master.
[Jira](https://jira.zenoss.com/browse/ZEN-25802)

### DIFF
--- a/pkg/serviced-zenossdbpack
+++ b/pkg/serviced-zenossdbpack
@@ -1,3 +1,8 @@
 #!/bin/sh
 
-${SERVICED:=/opt/serviced/bin/serviced} service run zope zenossdbpack
+if [ ! -d /opt/serviced/log/zenossdbpack ]; then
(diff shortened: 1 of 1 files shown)


pkg/serviced-zenossdbpack
```diff
@@ -1,3 +1,6 @@
 #!/bin/sh
 
-${SERVICED:=/opt/serviced/bin/serviced} service run zope zenossdbpack
\ No newline at end of file
+mkdir -p /opt/serviced/log/zenossdbpack
```
tgjonestx: There's no need to do this every time.  Should be something like:
```
if [ ! -d /opt/serviced/log/zenossdbpack ]; then
    mkdir -p /opt/serviced/log/zenossdbpack
    chmod 1777 /opt/serviced/log/zenossdbpack
fi 
```

PR #3331, opened 2017-01-18
CC-3194: Container metrics are sporadic

spopezen:
https://jira.zenoss.com/browse/CC-3194

Samples were getting a reference to the same map, causing missing and sporadic data.
The changes in this PR create a map for each metric in each registry.


spopezen:
retest this

PR #3334, opened 2017-01-19
CC-3108 "Use "Run-level" property to order normal service startup and shutdown"

johnlhamilton:
Introduced a service state manager to manage a queue of service schedule requests and execute them in batches according to start level and emergency shutdown level.


domain/service/service.go
```diff
@@ -120,30 +120,6 @@ type Service struct {
 	datastore.VersionedEntity
 }
```
spopezen: Note: This was [moved to servicestatemanager.go](#diff-c8e81cdb91f28c802e17dec1fc06248d)

facade/dfs.go
```diff
@@ -554,6 +552,15 @@ func (f *Facade) Snapshot(ctx datastore.Context, serviceID, message string, tags
...
 	}
+	// Pause the services that need pausing in a batch
+	if _, err := scheduleServices(f, servicesToPause, ctx, tenantID, service.SVCPause, false); err != nil {
+		glog.Errorf("Could not pause services for snapshot: %s", err)
+		return "", err
+	}
+
+	defer scheduleServices(f, servicesToPause, ctx, tenantID, service.SVCRun, false)
+
+	// Wait for the paused services to reach the paused state (and other services to reach stopped)
```
spopezen: Note: Previously, the services were getting individual requests to pause and defer a return to previous state (which would always be run), this was changed to do a batch pause and defer a batch run.

facade/facade.go
```diff
@@ -126,4 +128,4 @@ func (f *Facade) SetHostExpirationRegistry(hostRegistry auth.HostExpirationRegis
 
 func (f *Facade) SetDeploymentMgr(mgr *PendingDeploymentMgr) { f.deployments = mgr }
 
-func (f *Facade) SetServiceRunLevelTimeout(t time.Duration) { f.serviceRunLevelTimeout = t }
+//func (f *Facade) SetServiceRunLevelTimeout(t time.Duration) { f.serviceRunLevelTimeout = t }
```
spopezen: Remove this?

facade/service.go
```diff
@@ -1180,7 +1181,9 @@ func (f *Facade) ScheduleService(ctx datastore.Context, serviceID string, autoLa
 	mutex := getTenantLock(tenantID)
 	mutex.RLock()
 	defer mutex.RUnlock()
-	return f.scheduleService(ctx, tenantID, serviceID, autoLaunch, synchronous, desiredState, false, false)
+	// Pause requests are always expedited
+	emergency := desiredState == service.SVCPause
+	return f.scheduleService(ctx, tenantID, serviceID, autoLaunch, synchronous, desiredState, false, emergency)
 }
```
spopezen: I don't think we need to make Pause an emergency anymore since when we Snapshot, we're calling lockTenant, which calls Wait, which waits until there is nothing left in the queue.  (probably no negative side effects though)

facade/service.go
```diff
@@ -1245,129 +1253,72 @@ func (f *Facade) scheduleService(ctx datastore.Context, tenantID, serviceID stri
 		return 0, err
 	}
 
-	serviceScheduler := func() (int, error) {
```
spopezen: Note: This was how we handled emergency shutdown previously - it's now performed in servicestatemanager.

facade/service.go
```diff
@@ -1503,14 +1454,39 @@ func (f *Facade) ListTenants(ctx datastore.Context) ([]string, error) {
...
+
+		return f.ScheduleService(ctx, request.ServiceID, request.AutoLaunch, request.Synchronous, service.SVCRun)
+	}
+
+	if request.Synchronous {
+		return forceRestart()
+	} else {
+		go forceRestart()
+		return 0, nil
+	}
```
spopezen: Note: This is a "forcedRestart" (synchronous stop, then a start) in order to balance services while observing startLevel, which should be a separate function (or removed?) when rolling restarts is implemented.

facade/dfs.go
johnlhamilton: Need to do the same for Rollback

facade/service.go
```diff
@@ -1228,7 +1231,11 @@ func (f *Facade) scheduleService(ctx datastore.Context, tenantID, serviceID stri
 
 	// Build a list of services to be scheduled
 	svcs := []*service.Service{}
+	var svcIDs []string
 	visitor := func(svc *service.Service) error {
+		if svc.ID != serviceID && svc.Launch == commons.MANUAL && !emergency {
```
johnlhamilton: If it is an emergency shutdown, we want to shut it down even if Launch is set to MANUAL.

cli/api/daemon.go
```diff
@@ -111,6 +112,7 @@ type daemon struct {
 	tokenExpiration  time.Duration
 
 	facade *facade.Facade
+	ssm    *servicestatemanager.BatchServiceStateManager
```
johnlhamilton: Easy enough to do, but there are two arguments against that:
1. The facade member is the impl, not facade.FacadeInterface
2. ssm.SetFacade, ssm.Start, and ssm.Stop are not currently part of the interface, since a different implementation may not need those (or may have a different set of Facade methods it needs).  Of course, that's not actually going to happen ever.

johnlhamilton: Oops, Argument 2 isn't really valid here.  We could still have the interface here as long as NewBatchServiceStateManager returns the impl and not the interface.

scheduler/servicestatemanager/servicestatemanager.go
```diff
@@ -0,0 +1,947 @@
...
+	sync.RWMutex
+	Facade                 Facade
+	ctx                    datastore.Context
+	ServiceRunLevelTimeout time.Duration
+	TenantQueues           map[string]map[service.DesiredState]*ServiceStateQueue
+	TenantShutDowns        map[string]chan<- int
+}
+
+func NewCancellableService(svc *service.Service) CancellableService {
+	cancel := make(chan interface{})
```
johnlhamilton: Unfortunately, that eventually gets passed to zzk.WaitService, which requires a chan interface{}.

PR #3343, opened 2017-01-26
Feature/cc 3214

smousa:


zzk/service/hoststate_test.go
```diff
@@ -1455,6 +1456,211 @@ func (t *ZZKTest) TestHostStateListener_Shutdown_Spawn(c *C) {
...
+		ImageUUID:   imageId,
+		Paused:      false,
+		Started:     time.Now(),
+	}
+	var retShutdown <-chan interface{} = shutdown
+	containerExit := make(chan time.Time, 1)
+	var retExit <-chan time.Time = containerExit
+
+	handler.On("AttachContainer", mock.AnythingOfType("*service.ServiceState"), serviceId, 1).Return(nil, nil)
+	handler.On("StartContainer", retShutdown, serviceId, 1).Return(ssdat, retExit, nil)
```
johnlhamilton: Can we put a .Once() on this so it will fail if StartContainer is called later during the restart?

zzk/service/hoststate.go
```diff
@@ -302,6 +307,25 @@ func (l *HostStateListener) setInstanceState(containerExit <-chan time.Time, ssd
...
+
+		// set the host state
+		if err := UpdateState(l.conn, req, func(s *State) bool {
+			if s.DesiredState == service.SVCRestart {
+				s.DesiredState = service.SVCRun
+				return true
+			}
+			return false
+		}); err != nil {
+			logger.WithError(err).Error("Could update desired state to restart container")
```
johnlhamilton: Type-o  Could NOT update....

node/container.go
```diff
@@ -168,6 +168,63 @@ func (a *HostAgent) StartContainer(cancel <-chan interface{}, serviceID string,
...
+
+	// look up the container to get the image
+	ctrName := fmt.Sprintf("%s-%d", serviceID, instanceID)
+	ctr, err := docker.FindContainer(ctrName)
+	if err == docker.ErrNoSuchContainer {
+		// container has been deleted so we will pull when the container starts
+		// again.
+		logger.Debug("Container not found")
+		return nil
+	}
```
smousa: I copied from ResumeContainer, though I can fix that for both

zzk/service/hoststate.go
smousa: Thanks, I will change that.

node/container.go
```diff
@@ -168,6 +168,63 @@ func (a *HostAgent) StartContainer(cancel <-chan interface{}, serviceID string,
 	return state, ev, nil
 }
 
+// RestartContainer asynchronously pulls the latest image of a running
+// container before stopping the service.  After the service has stopped, the
+// listener will be notified by the event monitor.
+func (a *HostAgent) RestartContainer(cancel <-chan interface{}, serviceID string, instanceID int) error {
```
smousa: Good catch, fixed.

node/container.go
```diff
@@ -168,6 +168,66 @@ func (a *HostAgent) StartContainer(cancel <-chan interface{}, serviceID string,
...
+			}
+			logger.Debug("Pulled image")
+			break
+		}
+
+		// set the container to stop; ctr.Stop() stops the container by
+		// container id and not name, so if the container was stopped or
+		// deleted before the pull is successful, then this will just be a
+		// no-op.
+		if err := ctr.Stop(45 * time.Second); err != nil {
```
jplouis: Stopping the container will cause it to be restarted by the delegate, correct? If so can you note it in the comments as the restart is a side effect of the ctr.Stop

PR #3348, opened 2017-01-27
CC-3199: Implement rolling restarts

spopezen:
https://jira.zenoss.com/browse/CC-3199

Adds rolling restarts and makes it the default restart behavior.
Also, adds ability to wait on and restart a single instance of a service.


spopezen:
Retest this

facade/zkapi.go
```diff
@@ -704,6 +764,61 @@ func (zk *zkf) StopServiceInstances(ctx datastore.Context, poolID, serviceID str
...
+			}
+			return false
+		}); err != nil {
+			logger.WithError(err).Warn("Could not restart service instance")
+			return err
+		}
+		logger.Debug("Set service instance to restart")
+	} else {
+		logger.Warning("Could not restart service instance, host is not online")
+		return ErrHostDoesNotExist
```
johnlhamilton: Is there an ErrHostOffline? If not, lets add one

facade/service.go
```diff
@@ -1336,21 +1336,50 @@ func (f *Facade) ScheduleServiceBatch(ctx datastore.Context, svcs []*service.Ser
...
-		// shutdown all service instances
-		if err := f.zzk.StopServiceInstances(ctx, svc.PoolID, svc.ID); err != nil {
+	if desiredState == service.SVCRestart {
+		//  place-holder ready func until we figure out how best to determine if we should move on to the next instance
+		if err := f.rollingRestart(ctx, svc, func() bool {
+			time.Sleep(5 * time.Second)
+			return true
+		}); err != nil {
+			glog.Errorf("Facade.updateDesiredState: Could not perform rolling restart for service %s (%s): %s", svc.Name, svc.ID, err)
 			return err
```
johnlhamilton: Maybe we should fall back to a hard restart in this case.  If a host is offline and the user hits "restart", they won't get any feedback unless they look at the logs.  I don't know, its worth discussing.

zzk/service/service.go
```diff
@@ -404,3 +404,67 @@ func WaitService(cancel <-chan struct{}, conn client.Connection, poolID, service
...
+		select {
+		case <-cancel:
+			return nil
+		default:
+		}
+	}
+
+	// otherwise, wait for the event for the instance
+	select {
+	case event := <-ev:
```
johnlhamilton: MonitorState blocks until the changes we are waiting on have occurred.  I don't think we need to wait for another event.

zzk/service/service_test.go
```diff
@@ -97,3 +97,88 @@ func (t *ZZKTest) TestWaitService(c *C) {
...
+
+	// create the state
+	req := StateRequest{
+		PoolID:     "poolid",
+		HostID:     "hostid",
+		ServiceID:  "serviceid",
+		InstanceID: instanceID,
+	}
+
+	timer.Reset(time.Second)
```
johnlhamilton: Why are we waiting again here?

facade/service.go
```diff
@@ -1336,21 +1336,50 @@ func (f *Facade) ScheduleServiceBatch(ctx datastore.Context, svcs []*service.Ser
...
+			case <-time.After(30 * time.Second):
+				close(cancel)
+			}
+		}()
+		err := f.zzk.WaitInstance(ctx, svc, instanceID, service.SVCRun, cancel)
+		if err != nil {
+			return err
+		}
+		// Check if we're ready to move on to the next instance on an interval
+		for !ready() {
```
spopezen: As discussed, ready() returning true indicates that a health check passes or another condition is met where we don't want to wait on this instance anymore.

zzk/service/service_test.go
spopezen: WaitInstance isn't going to return until UpdateState finishes, then it will `close(done)`.  We want to see this happen, but within some short time period, so we fail if there is a bug in WaitInstance where it isn't returning in a timely manner.

spopezen:
Based on some discussion, progress on this story is blocked until new states and/or some state caching in the facade exists.  Closing pull request until then.

PR #3351, opened 2017-02-01
CC-3069 "Warn/Show Error when authorization fails due to clock drifts between hosts"

johnlhamilton:
On host authentication, we now fail authentication if the timestamp of the request is +/- auth.ExpirationDelta of the master's current time.  This will flip the "Authenticated" icon in the UI to red, and will log an error on the master about re-syncing the clocks.


johnlhamilton:
retest this

PR #3352, opened 2017-02-01
Feature/cc 3200 state change notify

mayathegreat:
* added service states
* modified state changes to reflect new additional "Instance State"
* test cases added for the above


node/container.go
```diff
@@ -367,9 +379,11 @@ func (a *HostAgent) pullImage(logger *log.Entry, cancel <-chan interface{}, imag
 }
 
 // monitorContainer tracks the running state of the container.
+// runs when the container dies
 func (a *HostAgent) monitorContainer(logger *log.Entry, ctr *docker.Container) <-chan time.Time {
 	ev := make(chan time.Time, 1)
 	ctr.OnEvent(docker.Die, func(_ string) {
+		a.instanceCache.SetKey(ctr.Name, StateStopped)
```
smousa: get rid of this--should be calling setInstanceState

node/container.go
```diff
@@ -56,6 +56,7 @@ func (a *HostAgent) StopContainer(serviceID string, instanceID int) error {
 		return err
 	}
 
+	a.setInstanceState(serviceID, instanceID, StateStopping)
```
smousa: Where is the declaration of this function?

smousa:
I helped with this, so someone else please review

node/container.go
```diff
@@ -367,9 +400,12 @@ func (a *HostAgent) pullImage(logger *log.Entry, cancel <-chan interface{}, imag
 }
 
 // monitorContainer tracks the running state of the container.
+// runs when the container dies
 func (a *HostAgent) monitorContainer(logger *log.Entry, ctr *docker.Container) <-chan time.Time {
 	ev := make(chan time.Time, 1)
 	ctr.OnEvent(docker.Die, func(_ string) {
+		a.setInstanceState(StateStopped)
+		a.instanceCache.SetKey(ctr.Name, StateStopped)
```
johnlhamilton: What is instanceCache?

zzk/service/state.go
```diff
@@ -577,8 +635,24 @@ func DeleteState(conn client.Connection, req StateRequest) error {
 		logger.Debug("No state to delete on host")
 	}
 
-	// Delete the service instance
 	sspth := path.Join(basepth, "/services", req.ServiceID, req.StateID())
+
+	// Delete the current state (status)
```
johnlhamilton: Is this necessary, since we are deleting the parent node below?

node/container.go
```diff
@@ -56,6 +77,7 @@ func (a *HostAgent) StopContainer(serviceID string, instanceID int) error {
 		return err
 	}
 
+	a.setInstanceState(serviceID, instanceID, StateStopping)
 	err = ctr.Stop(45 * time.Second)
 	if _, ok := err.(*dockerclient.ContainerNotRunning); ok {
 		logger.Debug("Container already stopped")
```
smousa: That will be handled by the event monitor code

node/container.go
```diff
@@ -65,6 +87,7 @@ func (a *HostAgent) StopContainer(serviceID string, instanceID int) error {
 		return err
```
smousa: The event monitor receives an event when the container exits and will update the state accordingly

zzk/service/state.go
smousa: Yes, because if the node is not there, you don't want to add it to the transaction, otherwise it will error

node/container.go
```diff
@@ -65,6 +87,7 @@ func (a *HostAgent) StopContainer(serviceID string, instanceID int) error {
...
 	}
 
+	a.setInstanceState(serviceID, instanceID, StateStopped)
```
johnlhamilton: Given Summer's comments above, this should NOT be in a defer

zzk/service/state.go
johnlhamilton: In that case, should we return an error (if the node is missing) or just continue and not include it in the transaction?

node/container.go
smousa: This does not need to be set here

node/container.go
mayathegreat: added

node/container.go
mayathegreat: was a remnant of an older plan. Removed.

node/container.go
```diff
@@ -65,6 +87,7 @@ func (a *HostAgent) StopContainer(serviceID string, instanceID int) error {
 		return err
 	}
 
+	a.setInstanceState(serviceID, instanceID, zkservice.StateStopped)
```
smousa: Remove this line, this will be handled by the event monitor
https://github.com/control-center/serviced/pull/3352/files#diff-fbf6fa650618db48c5e4bcb7ad4b84d4R410

mayathegreat: done

PR #3355, opened 2017-02-06
CC-3199 "Add support for rolling restart"

johnlhamilton:
The default restart behavior is now to restart multi-instance services one instance at a time.  We wait for an instance to restart and pass healthchecks before restarting the next one, with a timeout specified by SERVICED_RUN_LEVEL_TIMEOUT.

To perform the old-style restart (stopping all instances before restarting them), the user can use the new CLI option:
`serviced service restart --rebalance SERVICE`

### DIFF
--- a/facade/service.go
+++ b/facade/service.go
@@ -42,6 +42,7 @@ import (
 	"github.com/control-center/serviced/domain/service"
 
 	"github.com/control-center/serviced/utils"
+	"sync"
 )
 
 const (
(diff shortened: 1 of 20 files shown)


facade/service.go
```diff
@@ -1230,6 +1231,31 @@ func (f *Facade) clearEmergencyStopFlag(ctx datastore.Context, tenantID, service
 	return cleared, nil
 }
 
+// countServices will count how many unique services in the given list (including children) satisfy the check function
+func (f *Facade) countServices(ctx datastore.Context, serviceIDs []string, check func(*service.Service) bool) (int, error) {
+	alreadyChecked := make(map[string]bool)
```
smousa: Instead of a bool, this could be a map[string]struct{}, and you would be checking the presence of key in the struct instead of checking the bool value.  The advantage of this is that the struct{} type takes up 0 bytes, while the bool type takes > 0 bytes.
```go
checked := make(map[string]struct{})
a, ok := checked[name]
//ok represents the presence of the element in the map
```

facade/service.go
```diff
@@ -1368,21 +1411,139 @@ func (f *Facade) ScheduleServiceBatch(ctx datastore.Context, svcs []*service.Ser
...
+		"serviceid":    svc.ID,
+		"servicename":  svc.Name,
+		"desiredstate": desiredState,
+	})
+	if desiredState == service.SVCRestart && svc.DesiredState == int(service.SVCStop) {
+		// If we are already stopped or stopping, just start it back up
+		desiredState = service.SVCRun
+	}
+
+	if desiredState == service.SVCRestart {
```
smousa: should this be in an else if block?

facade/service.go
```diff
@@ -1368,21 +1411,139 @@ func (f *Facade) ScheduleServiceBatch(ctx datastore.Context, svcs []*service.Ser
...
+			// If we got cancelled, bail:
+			select {
+			case <-cancelWait:
+				errC <- nil
+				return
+			default:
+			}
+
+			// Then for the start
+			errC <- f.zzk.WaitInstance(ctx, svc, instanceID, service.SVCRun, cancelWait)
```
smousa: is it possible for SVCRun to never reach this state?  For example a case where the service is instead stopped?

facade/service.go
```diff
@@ -1368,21 +1411,139 @@ func (f *Facade) ScheduleServiceBatch(ctx datastore.Context, svcs []*service.Ser
...
+				close(cancelWait)
+			}
+		}()
+
+		// We need to start the wait in a goroutine before calling restart, so we can be sure to catch the first
+		// transition from running to stopped
+		errC := make(chan error)
+		go func() {
+			// First wait for the stop
+			err := f.zzk.WaitInstance(ctx, svc, instanceID, service.SVCStop, cancelWait)
```
smousa: instead of WaitInstance, I would propose that you monitor the instance for the ContainerID to change, because the first request would be a get and then the second request would be a watch

smousa: And you could call the get outside of the scope of this goroutine

facade/service.go
johnlhamilton: No, if the if statement above passes we still want to go into the else below (with a new desiredState of SVCRun)

facade/service.go
johnlhamilton: Yeah, that could happen.  We should probably implement a cancel mechanism that could be initiated by the service state manager when a new state change request comes in.  I think we will still have to deal with this even when we switch to monitoring the ContainerID instead of WaitInstance, correct?

smousa: Yes

PR #3356, opened 2017-02-07
Feature/cc 3229

smousa:
DO NOT MERGE


smousa:
moving to feature branch

PR #3360, opened 2017-02-08
CC-2076 Set permissions and group for binaries and default config

glujan:
[Jira](https://jira.zenoss.com/browse/CC-2076)

### DIFF
--- a/makefile
+++ b/makefile
@@ -158,12 +158,15 @@ $(GOBIN):
 
 $(GOBIN)/serviced: $(GOSOURCEFILES) | $(GOBIN)
 	$(GO) build $(GOBUILD_FLAGS) ${LDFLAGS} -o $@ .
+	@chmod 750 $@
 
 $(GOBIN)/serviced-controller: $(GOSOURCEFILES) | $(GOBIN)
 	$(GO) build $(GOBUILD_FLAGS) ${LDFLAGS} -o $@ ./serviced-controller
+	@chmod 750 $@
 
 $(GOBIN)/serviced-storage: $(GOSOURCEFILES) | $(GOBIN)
 	$(GO) build $(GOBUILD_FLAGS) ${LDFLAGS} -o $@ ./tools/serviced-storage
+	@chmod 750 $@
 
 .PHONY: serviced
 serviced: $(GOBIN)/serviced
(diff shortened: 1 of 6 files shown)


jpeacock-zenoss:
I can't tell for certain, but does this change ensure that everything in `/opt/serviced` has `root:serviced` ownership?

pkg/deb/postinstall
```diff
@@ -2,3 +2,9 @@
 mkdir -p /var/log/serviced
 chmod 1777 /var/log/serviced
 
+for BINARY in "serviced" "serviced-controller" "serviced-storage"; do
+    chgrp serviced /opt/serviced/bin/${BINARY}
+done
+
+chgrp serviced /etc/default/serviced
+chgrp serviced /opt/serviced
```
jpeacock-zenoss: There's a loop above that changes three specific files to have `root:serviced` ownership, but it seems inconsistent to have different ownership in the `/opt/serviced` directory.

Changing the above line to `chgrp -R serviced /opt/serviced` will switch everything over to the `serviced` group and eliminate the need for the earlier loop.

glujan: You are right, it makes sense to recursively change group. And permissions are already set in `makefile`.

glujan:
@tgjonestx Could you have a look at this as well? Does this PR make sense to you?

tgjonestx:
@glujan - A few comments

* I don't see where the permissions for `/etc/default/serviced` are set to `640`
  * We don't want users who are not in the `serviced` group to view the configuration
* I don't see where the permissions for `/opt/serviced` are set to `750`
  * We don't want users who are not in the `serviced` group to list any directory contents below `/opt/serviced`
  * We don't want user who are not in the `serviced` group to be able to execute the binaries in `/opt/serviced/bin` 
* I don't see where the `root` user is added to the `serviced` group
  * Technically speaking this is not required to achieve our security goals, but I think it's a nice-to-have to make that `serviced` group somewhat "self documenting"; i.e. these are the users allowed to use serviced and by default, only root is allowed

Also, did you trying building and installing the RPM to verify that the group and permissions are all set as expected?

makefile
```diff
@@ -410,7 +413,8 @@ $(install_DIRS): FORCE
 				exit 1 ;\
 			fi ;\
 		done ;\
-	done
+	done ;\
+	chown 750 $(_DESTDIR)${prefix}
```
glujan: Here permissions are set to `/opt/serviced/` directory

glujan:
Retest this

glujan:
retest

PR #3361, opened 2017-02-08
CC-3105: Thresholds for Emergency Shutdown

spopezen:
https://jira.zenoss.com/browse/CC-3105

These are some thresholds that have provided reasonable predictions in calculations and visualizations.


spopezen:
Retest this

PR #3368, opened 2017-02-14
CC-3239 Show service and instance statuses in UI separate from healthchecks

johnlhamilton:


scheduler/servicestatemanager/servicestatemanager.go
```diff
@@ -269,6 +289,112 @@ func (s *BatchServiceStateManager) removeTenant(tenantID string) error {
...
+			<-thread.Done
+			delete(s.currentStateWaits, svc.ID)
+		}
+	}
+
+	finalState := service.DesiredToCurrentFinalState(desiredState, emergency)
+	currentState := service.DesiredToCurrentTransitionState(desiredState, emergency)
+
+	// Set current state to the transition state
+	s.updateServiceCurrentState(currentState, svc.ID)
```
spopezen: I think we need to lock around this? (and function header comment that explains it needs to be locked around, or just lock in the function)

johnlhamilton: The calling method should (and does) get the lock.  I added a comment on both methods stating that the lock is required before calling.

PR #3374, opened 2017-02-16
Set the states of emergency stopped services to stopped as the emerge…

youngklee:
…ncy shutdown flag is cleared


youngklee:
retest

PR #3377, opened 2017-02-16
CC-3279: Service status not reflective of instance when restarting

spopezen:
https://jira.zenoss.com/browse/CC-3279

These changes set the status of our successfully and completely restarted services earlier, so we don't wait for other services in the batch.


facade/service.go
```diff
@@ -1450,6 +1448,8 @@ func (f *Facade) rollingRestart(ctx datastore.Context, svc *service.Service, tim
 
 	// Build the service health object to use for getting instance health
 	svch := service.BuildServiceHealth(*svc)
+	var punctualLock sync.RWMutex
+	punctualInstances := make([]struct{}, 0)
```
johnlhamilton: why not punctualInstances := 0 ?

spopezen: I don't know

PR #3378, opened 2017-02-16
CC-3267 "No "pending restart" state on instances"

johnlhamilton:
During a rolling restart, when we begin restarting a service, instances that have not yet restarted will show up as "Pending restart" in both the UI and CLI.

### DIFF
--- a/scheduler/servicestatemanager/servicestatemanager_unit_test.go
+++ b/scheduler/servicestatemanager/servicestatemanager_unit_test.go
@@ -1931,7 +1931,7 @@ func (s *ServiceStateManagerSuite) TestServiceStateManager_StartShutdown(c *C) {
 	}
 }
 
-func (s *ServiceStateManagerSuite) TestServiceStateManager_queueLoop(c *C) {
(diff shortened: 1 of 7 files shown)


scheduler/servicestatemanager/servicestatemanager_unit_test.go
```diff
@@ -2025,6 +2024,8 @@ func (s *ServiceStateManagerSuite) TestServiceStateManager_queueLoop(c *C) {
 	done := make(chan struct{})
 	go func() {
 		wg.Wait()
+		// Wait for all services to get scheduled so we don't randomly fail assertExpectations below
+		s.serviceStateManager.WaitScheduled("tenant1", "A", "D", "G", "H")
```
spopezen: Is this test function testing anything that TestServiceStateManager_WaitScheduled doesn't now?

johnlhamilton: Nope, I'll get rid of it.

PR #3385, opened 2017-02-21
silently (re)try to updateHost, fix CC-3262

corntoole:
Fixes regression in CC-3262 by not blocking delegate when it fails to update the Master, say for instance if things are locked down during a backup on the Master. It does no kicking off a goroutine that keeps trying to updateHost until no errors are returned.

### DIFF
--- a/cli/api/daemon.go
+++ b/cli/api/daemon.go
@@ -849,7 +849,7 @@ func (d *daemon) startAgent() error {
 								case <-d.shutdown:
 									return
 								default:
-									time.Sleep(5)
+									time.Sleep(5 * time.Second)
 							}
 							err = masterClient.UpdateHost(updatedHost)
 						}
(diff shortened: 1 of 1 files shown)


cli/api/daemon.go
```diff
@@ -841,10 +841,21 @@ func (d *daemon) startAgent() error {
...
-					return "" // Try again
+					log.WithError(err).Warn("Unable to update master with delegate host information. Retrying silently")
+					go func(masterClient *master.Client, updatedHost host.Host) {
+						err := errors.New("")
+						for err != nil {
+							select {
+								case <-d.shutdown:
+									return
+								default:
+									time.Sleep(5)
```
tgjonestx: I think you want `time.Sleep(5 * time.Second)`; otherwise it's only sleeping 5 nanoseconds

PR #3394, opened 2017-02-23
tighten permissions on some directories post install

tgjonestx:
The goal is to prevent users who are not in the new `serviced` group from looking at serviced files or executing serviced binaries.

### DIFF
--- a/pkg/deb/postinstall
+++ b/pkg/deb/postinstall
@@ -1,6 +1,8 @@
 
 mkdir -p /var/log/serviced
-chmod 1777 /var/log/serviced
+chgrp serviced /var/log/serviced
+chmod 1770 /var/log/serviced
 
 chgrp serviced /etc/default/serviced
 chgrp -R serviced /opt/serviced
+chmod 750 /opt/serviced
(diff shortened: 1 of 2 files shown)


pkg/deb/postinstall
```diff
@@ -1,6 +1,8 @@
 
 mkdir -p /var/log/serviced
-chmod 1777 /var/log/serviced
+chgrp serviced /var/log/serviced
```
smousa: does this group need to be hardcoded to a specific GID for containers?

tgjonestx: IDK. The problem we're trying to solve is that users logged into a CC node, who are not in the `serviced` group, should not be able to execute the CC CLI or look at CC config data.

So far, the containerized services seem to have been running fine after this change with the exception of ZEN-26903 which is under investigation

tgjonestx: @smousa - I tested a 5.2.1 install without any of these changes, and the files under `/opt/serviced/var/volumes/tenantID` are owned by a variable number of UID/GID - only some were owned by `root:root`. I also checked a 5.2.2 install with a preliminary set of these changes, and the permissions on `/opt/serviced/var/volumes/tenantID` were varied in the same way. So I don't think having the `serviced-controller` binary with `root:serviced` permissions will have any impact on the ownership of files in the DFS created by various services from their respective containers.

PR #3395, opened 2017-02-23
tighten permissions on some directories post install

tgjonestx:
The goal is to prevent users who are not in the new serviced group from looking at serviced files or executing serviced binaries.


tgjonestx:
Closing per discussion with Rajan. Will backout other changes related to CC-3067 on the support/1.2.x branch until we resolve apparent side-effects like ZEN-26903

PR #3396, opened 2017-02-23
silently retry to updateHost, fix CC-3262 (#3385)

corntoole:
* silently (re)try to updateHost, fix CC-3262
  * sleep for 5 secs


corntoole:
My mistake, this PR shouldn't have been made. The fix version for the CC-3262 is 1.3.0.

PR #3398, opened 2017-02-23
CC-3275 Service state manager detects "cancel" scenarios and handles them appropriately

johnlhamilton:
Relaxed locking around the current state updates to allow more immediate updating of current state when a pending state is cancelled.

Use the channel in CancellableService to cancel "-ing" states as best we can, especially for "restarting".

### DIFF
--- a/facade/service.go
+++ b/facade/service.go
@@ -1365,7 +1365,7 @@ func scheduleServices(f *Facade, svcs []*service.Service, ctx datastore.Context,
 	return len(svcs), nil
 }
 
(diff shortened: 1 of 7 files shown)


facade/service.go
```diff
@@ -1476,6 +1488,16 @@ func (f *Facade) rollingRestart(ctx datastore.Context, svc *service.Service, tim
 		}
 	}
 
+	nextInstance := 0
+	defer func() {
```
spopezen: Could use a comment...

PR #3400, opened 2017-02-24
CC-3285: Use mimmax instead of sum for storage aggregator

spopezen:
https://jira.zenoss.com/browse/CC-3285

It is possible for a fail-over to occur in HA but still have some metrics being pushed, which was causing an emergency shutdown when using the sum aggregator.
(note; the commit comment incorrectly states min)

This required updating to latest isvcs, as it has the latest central query, which added the aggregator for mimmax.


spopezen:
Retest this

smousa:
Don't run r*test this!  Find the broken build and fix manually!!!

spopezen:
I'll remember that, next time.

PR #3410, opened 2017-03-01
Revert "CC-3285: Use mimmax instead of sum for storage aggregator"

spopezen:
Reverts control-center/serviced#3400

Need to figure out how the graphs got killed...


spopezen:
Found the problem, fixing properly, will open another pr when that fix is in.

PR #3412, opened 2017-03-02
Feature/cc 2421

morr-zenoss:
Ready for merge once approved by reviewers.
Changes are for CC-2421

Thanks to Summer for her review and good feedback. She also addressed most of the concerns I had earlier. 

Thanks!


cli/api/backup.go
```diff
@@ -29,13 +30,17 @@ func (a *api) Backup(dirpath string, excludes []string) (string, error) {
...
 	}
-
 	var path string
 	req := dao.BackupRequest{
 		Dirpath:              dirpath,
 		SnapshotSpacePercent: config.GetOptions().SnapshotSpacePercent,
 		Excludes:             excludes,
 	}
+
+	est := dao.BackupEstimate{}
```
smousa: Can you move this call into the rpc package?  We shouldn't be creating new calls in the dao layer.

dao/model.go
```diff
@@ -125,3 +125,12 @@ type BackupRequest struct {
 	SnapshotSpacePercent int
 	Excludes             []string
 }
+
+type BackupEstimate struct {
```
smousa: Again, move this to the RPC package

morr-zenoss: Same comment as above - definitely a good idea, but would prefer to wait until post-Alderaan due to existing dependencies.

cli/api/backup.go
morr-zenoss: I tried earlier, but there were dependencies (BackupRequest, specifically) that caused problems for the other backup functions here when I tried to move it. Per our discussion, I'd like to hold off on moving this code until Alderaan is released.

dao/model.go
morr-zenoss: Same as other instance - prefer to hold off on move until after Alderaan, due to conflicts with existing calls.

PR #3416, opened 2017-03-03
Add separate tolerances before and after the token valid duration

achow-zenoss:
1. The handling of clock drift is separated out from the expiration tolerance
2. Starting the token refresh earlier
3. Fix valid checks against the allowed clock drift


auth/jwt.go
```diff
@@ -111,7 +111,8 @@ func (id *jwtIdentity) Valid() error {
 	}
 
 	now := jwt.TimeFunc().UTC().Unix()
-	if now < id.IssuedAt {
+	// provide tolerance for clockdrifting slower
+	if now < id.IssuedAt-int64(ClockDriftDelta.Seconds()) {
```
johnlhamilton: 3 Questions:
1. Does this ever get called?
2. If it does, should we use the ClockDriftDelta here or the ExpirationDelta?  This is checking the token's expiration, so I think we should use ExpirationDelta
3. I think we should use the same tolerance in the other direction for the id.Expired() method below

auth/token.go
```diff
@@ -23,10 +23,14 @@ import (
...
-	// considered expired. This should help avoid expiration races when server
-	// times don't match
+	// expirationDelta and ClockDriftDelta are margins of error during
+	// which a token should be considered expired.
+	// Together they help avoid expiration races when server/host times don't match
 	ExpirationDelta = 10 * time.Second
+	ClockDriftDelta = 5 * time.Minute
+
+	// refresh
+	RefreshAhead = 5 * time.Minute
```
johnlhamilton: Can we get separate comments on each of these explaining exactly what they control:  ClockDrift gives leeway on request expiration, ExpirationDelta gives leeway on token expiration, and RefreshAhead tells the delegate to refresh its auth token well in advance of expiration.

auth/jwt.go
johnlhamilton: Answering the first question:  Yes, we pass this object as a jwt.Claims, which requires a "Valid" method.

auth/token.go
achow-zenoss: Yes, it should be separated. Thanks.

auth/jwt.go
achow-zenoss: 2. This is a check of the token at the client side on the validity of the token it has. This particular part is to check if the token has started to become valid (not expired) So, using clockdriftdelta should be correct.
3. The code below should check the expiration with expire.. I agree that we probably need to add the possible timedriftdelta to this as well.

auth/token.go
```diff
@@ -23,10 +23,17 @@ import (
...
-	// considered expired. This should help avoid expiration races when server
-	// times don't match
-	ExpirationDelta = 10 * time.Second
+	// before - Not valid ..
+	// |---------------------!--------------------|---------------------|
+	// |<- ClockDriftDelta ->|<- token duration ->|<- ClockDriftDelta ->|
+	// after  - Expired ..
+
+	// ClockDriftDelta provides the tolerance for clock drift
+	// between master and hosts when a token is considered valid
```
johnlhamilton: should be 'when a token or request' is considered valid.

smousa:
Please pull develop

PR #3426, opened 2017-03-07
Update host IPs for CC-3182

cchristiezen:
At start up, when the host information is retrieved update the IPs of the host to pick up any changes like the interface name.


smousa:
Why not cherry-pick from develop?

smousa:
Do not merge into support at this time.  Not until after 1.3.0 is released.

PR #3427, opened 2017-03-07
Update IPs for host for CC-3346

cchristiezen:
At start up, when the host information is retrieved update the IPs of the host to pick up any changes like the interface name.


cchristiezen:
retest this

smousa:
Please don't use ret*st this to rerun jenkins tests, either fix the broken tests or rerun the broken build jobs

PR #3433, opened 2017-03-09
Feature/revert revert cc2421

morr-zenoss:
Re-adding changes for CC-2421 and CC-3319, with modifications:

Changed parameters BACKUP_ESTIMATED_COMPRESSION and BACKUP_MIN_OVERHEAD to 1.0 and 0, respectively. This has the effect of assuming no compression, for a worst-case scenario.

Added --force flag to command line. This allows taking of a backup if worst-case projects not enough room, but a reasonable human guess (say, based on historical backup data) says that a backup is likely to work. Using the --force flag does carry the risk of locking up serviced if there is indeed not enough room to take a backup.

Updated unit tests - added example tests for new command-line options.


cli/cmd/backup.go
```diff
@@ -67,7 +71,7 @@ func (c *ServicedCli) cmdBackup(ctx *cli.Context)  {
 			return
 		} else {
 			if ! backupSpace.AllowBackup {
-				fmt.Fprintf(os.Stderr, "Unable to backup: estimated space required (%s) exceeds space available (%s) on %s\n", backupSpace.EstimatedString, backupSpace.AvailableString, backupSpace.BackupPath)
+				fmt.Fprintf(os.Stdout, "Unable to backup: estimated space required (%s) exceeds space available (%s) on %s\n", backupSpace.EstimatedString, backupSpace.AvailableString, backupSpace.BackupPath)
```
ecifreo: Should be a log statement

cli/cmd/backup.go
```diff
@@ -77,8 +81,8 @@ func (c *ServicedCli) cmdBackup(ctx *cli.Context)  {
 		return
 	}
 	// do backup
-	if path, err := c.driver.Backup(args[0], ctx.StringSlice("exclude")); err != nil {
-		fmt.Fprintln(os.Stderr, err)
+	if path, err := c.driver.Backup(args[0], ctx.StringSlice("exclude"), ctx.Bool("force")); err != nil {
+		fmt.Fprintln(os.Stdout, err)
```
ecifreo: Another log statement

cli/cmd/backup.go
morr-zenoss: The message is logged in filesystem.go: https://github.com/control-center/serviced/blob/4412af56901bdbdc601df095ada55d35dfc5631f/dao/elasticsearch/filesystem.go#L111-L116. This message is for the user.

cli/cmd/backup.go
morr-zenoss: The message is logged in filesystem.go: https://github.com/control-center/serviced/blob/4412af56901bdbdc601df095ada55d35dfc5631f/dao/elasticsearch/filesystem.go#L109-L111. The code here is for the user.

PR #3434, opened 2017-03-13
Feature/cc 3382

morr-zenoss:
Maintenance ticket for CC-2421. 

CC-2421 changes were reverted last week due to concerns about implementation. This change reverts those reversions, and adds in changes that were made to address concerns. Those changes are:
* estimate backup size assuming no compression
* add -force flag to cli to allow for taking a backup even when estimate indicates there may not be enough space. This could be useful in situations where the compressed backup fits, but the uncompressed size is larger than the available space.
* improve cli tests to validate new functionality


smousa:
I feel like the UI should indicate that the backup estimate is uncompressed

PR #3444, opened 2017-03-15
Do not force services without dfs access to pause during snapshot

smousa:


facade/dfs.go
```diff
@@ -683,16 +684,34 @@ func (f *Facade) Snapshot(ctx datastore.Context, serviceID, message string, tags
...
+		// only pause services that have access to the dfs
+		hasDFS, ok := poolmap[svc.PoolID]
+		if !ok {
+			p := &pool.ResourcePool{}
+			err := f.poolStore.Get(ctx, pool.Key(svc.PoolID), p)
+			if err != nil {
+				logger.WithField("poolid", svc.PoolID).WithError(err).Error("Could not get resource pool")
+				return "", err
+			}
+			hasDFS = (p.Permissions & pool.DFSAccess) > 0
```
johnlhamilton: You can use p.HasDfsAccess()

PR #3450, opened 2017-03-20
Feature/CC-3182

cchristiezen:
Update IP information for host on startup.

### DIFF
--- a/test_lib.sh
+++ b/test_lib.sh
@@ -24,7 +24,7 @@ fi
 
 export START_TIMEOUT=300
 DIR="$(cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd)"
-DEFAULT_INTERFACE=$(ip route | awk '/default/ { print $5 }')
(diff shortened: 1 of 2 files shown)


test_lib.sh
```diff
@@ -24,7 +24,8 @@ fi
 
 export START_TIMEOUT=300
 DIR="$(cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd)"
-IP=$(ip addr show docker0 | grep -w inet | awk {'print $2'} | cut -d/ -f1)
+DEFAULT_INTERFACE=$(ip route | awk '/default/ { print $5 }')
```
rphillipsz: I found that both of these can return multiple addresses (see https://github.com/zenoss/zenoss-deploy/blob/develop/common/appliance/standard/addhostname.sh#L37) and cause problems running commands later when you expect just one value.  head -1 to use the first?

cchristiezen: Changed to use head -1

PR #3457, opened 2017-03-22
Fixes CC-3425 "Panic shutting down serviced during snapshot"

johnlhamilton:
These changes add several protections against the issue in the ticket:
1.) We check the channel to make sure it isn't already closed before closing it.
2.) We always remove the "current state wait" thread from the list after closing it, even when shutting down
3.) We guard against performing certain operations if the service state manager isn't running yet, or has been shut down.

### DIFF
--- a/scheduler/servicestatemanager/servicestatemanager.go
+++ b/scheduler/servicestatemanager/servicestatemanager.go
@@ -41,6 +41,10 @@ var (
 	ErrMissingQueue = errors.New("No queue found for tenant ID and desired state")
 	// ErrWaitTimeout occurs when we timeout waiting for a batch of services to reach the desired state
 	ErrWaitTimeout = errors.New("Timeout waiting for services")
+	// ErrNotRunning occurs when you try to perform an operation on a service state manager that is not running
+	ErrNotRunning = errors.New("Service state manager is not running")
+	// ErrAlreadyStarted occurs when you try to start a service state manager that has already been started
+	ErrAlreadyStarted = errors.New("Service state manager has already been started")
 )
 
 // ServiceStateManager provides a way to organize and manage services before scheduling them
@@ -122,9 +126,9 @@ type BatchServiceStateManager struct {
 	ServiceRunLevelTimeout time.Duration
 	TenantQueues           map[string]map[service.DesiredState]*ServiceStateQueue
 	TenantShutDowns        map[string]chan<- int
-
-	currentStateLock  sync.Mutex
-	currentStateWaits map[string]CurrentStateWait
+	currentStateLock       sync.Mutex
+	currentStateWaits      map[string]CurrentStateWait
+	shutdown               chan struct{}
 }
 
 type CurrentStateWait struct {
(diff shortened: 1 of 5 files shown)


johnlhamilton:
Currently, CurrentStateWait.Cancel() is always called under a currentStateLock, but I can add a lock in the CurrentStateWait struct to better protect against that.  I should do the same to CancellableService.

smousa:
It looked potentially racy, but if you have an outer lock, it is probably ok

scheduler/servicestatemanager/servicestatemanager.go
```diff
@@ -73,6 +73,7 @@ type ServiceStateChangeBatch struct {
 
 // CancellableService is a service whose scheduling may be canceled by a channel
 type CancellableService struct {
+	sync.Mutex
```
smousa: Is there a reason you are embedding this lock?

johnlhamilton: These objects get passed around a lot, and aren't always cancelled from a BatchServiceStateManager member method, so the alternative would be to have a package-level lock, which makes me nervous about potential deadlock.

smousa: I ask because, why is it embedded and not just an attribute on the object?

johnlhamilton: Good point, I will change them to private attributes.

scheduler/servicestatemanager/servicestatemanager.go
```diff
@@ -66,21 +66,21 @@ type ServiceStateManager interface {
...
 
 // CancellableService is a service whose scheduling may be canceled by a channel
 type CancellableService struct {
-	sync.Mutex
 	*service.Service
-	cancel chan interface{}
-	C      <-chan interface{}
+	cancel     chan interface{}
+	C          <-chan interface{}
+	cancelLock sync.Mutex
```
smousa: Need to make this a pointer, otherwise it will be a pass by value and be almost useless.

scheduler/servicestatemanager/servicestatemanager.go
```diff
@@ -111,7 +111,7 @@ func (s ByReverseStartLevel) Less(i, j int) bool {
 
 // ServiceStateQueue is a queue with a queueLoop processing it's batches
 type ServiceStateQueue struct {
-	sync.RWMutex
+	lock         sync.RWMutex
```
smousa: Needs to be a pointer

scheduler/servicestatemanager/servicestatemanager.go
```diff
@@ -121,27 +121,27 @@ type ServiceStateQueue struct {
 // BatchServiceStateManager schedules batches of services to start/stop/restart according to
 //  StartLevel
 type BatchServiceStateManager struct {
-	sync.RWMutex
+	lock                   sync.RWMutex
```
smousa: Needs to be a pointer

PR #3465, opened 2017-03-24
CC-3439: Set proper defaults and ranges

spopezen:
https://jira.zenoss.com/browse/CC-3439

Also setting the aggregator on the TSDB query to "mimmax", as that is what we use in CC.

### DIFF
--- a/web/ui/shutdownvis/src/app.js
+++ b/web/ui/shutdownvis/src/app.js
@@ -4,7 +4,7 @@ import {buildLegend} from "./legend";
 import {buildShutdownList} from "./shutdownlist";
 import {buildGraph} from "./graph";
 
-// colors come in pairs, one for actual data, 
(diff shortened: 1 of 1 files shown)


web/ui/shutdownvis/src/app.js
```diff
@@ -60,11 +60,9 @@ function fetchDiskUsage(config){
     let metrics = getDiskUsageMetrics(tenantIDs),
         tsdbURL = buildTSDBQuery(url, metrics, start, end);
 
-    // load up mock data
-    //tsdbURL = "available.json";
-
     return get(tsdbURL)
         .catch(e => {
+            alert("Unable to fetch TSDB data");
```
rDr4g0n: Only the devil uses alert messages. Are you the devil, son?

PR #3470, opened 2017-03-31
Use isvcs image v57, fix kibana.

ecifreo:


smousa:
This ticket doesn't have a fix version for 1.3.x

PR #3473, opened 2017-03-31
Add a pre-upgrade check for emergency shutdown

rphillipsz:
Added a pre-upgrade check to prevent upgrading serviced if this would place the system into Emergency Shutdown mode.
1) Issue a warning if the metadata size is under 1% of the data size.
2) Fail the upgrade if the data free is under the STORAGE_MIN_FREE setting.
3) Fail the upgrade if the metadata free is under 2% of the STORAGE_MIN_FREE setting.
4) Fail the upgrade if a tentant free is under the STORAGE_MIN_FREE setting.
5) Fail the upgrade if an application is deployed but there's no tenant mounted to check.
6) Add instructions for backing out of the failed upgrade
7) Add a link to the CC upgrade guide
8) Add instructions to override the check and install anyway

echeck

### DIFF
--- a/pkg/rpm/preupgrade
+++ b/pkg/rpm/preupgrade
@@ -1,0 +1,356 @@
+##############################################################################
+#
+# Copyright (C) Zenoss, Inc. 2017, all rights reserved.
+#
+# This content is made available according to terms specified in
+# License.zenoss under the directory where your Zenoss product is installed.
+#
+##############################################################################
+#
+# Checks the free space for the thinpool/meta/tenants prior to upgrading
+# serviced.  Gives a message and fails if the upgrade would place the
+# system into an emergency shutdown mode.
+#
+# Note that these are all bash scripts, so the python has been inserted
+# inline.
+#
+##############################################################################
+
+python <<EOF
+import os
+import re
+import subprocess
+import sys
+
+
+SERVICED_CONFIG="/etc/default/serviced"
+C_RESET  = '\033[0m'
+C_BOLD   = C_RESET + '\033[1m'
+C_WARN   = C_RESET + '\033[93m'
+C_ERR    = C_RESET + '\033[91m'
+
+# Parse a given file for key=value settings and return them as data.
+def parse(input_file):
+    config = {}
+    if os.path.isfile(input_file):
+        with open(input_file, 'r') as fd:
+            contents = fd.read()
+            for line in contents.split("\n"):
+                if "=" in line and not line.strip().startswith("#"):
+                    data = line.split("=", 1)
+                    config[data[0].strip()] = data[1].strip()
+    return config
+
+# Parses the serviced config file for values. Note that all values
+# are returned as strings.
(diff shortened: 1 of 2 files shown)


pkg/rpm/preupgrade
```diff
@@ -0,0 +1,356 @@
...
+        err += "\n%sThe metadata minimum threshold is calculated as 2 percent of the SERVICED_STORAGE_MIN_FREE value.%s" % \
+            (C_RESET, C_ERR)
+    if len(stats['tenants']):
+        for tenant in stats['tenants']:
+            if tenant['free'] < stats['data_min_free']:
+                err += "\nError: The tenant volume %s available space (%s) is under the minimum threshold (%s)" % \
+                    (tenant['id'], bytesize(tenant['free']), bytesize(stats['data_min_free']))
+    else:
+        # Only return an error here if an app has been deployed.
+        if app_is_deployed():
```
tgjonestx: This check is problematic for a couple of reasons. First, the check assumes that serviced is running during the upgrade; the user should be able to upgrade CC even if it's stopped.   

If the machine has just been rebooted and serviced hasn't been started, then no DM device will be active, and nothing will be mounted.  The same is true in the case of HA on the secondary (passive) node - there will be no DM device active so nothing will be mounted.

pkg/rpm/preupgrade
```diff
@@ -0,0 +1,356 @@
...
+    return data
+
+
+def get_tenant_stats():
+    """
+    Uses df to get free space for mounted tenant volumes (if any). df -h shows
+    base 1024 stats, so we need to parse these appropriately.
+    """
+    data = { 'tenants': [] }
+    cmd = "df -h --output=avail,target | grep /opt/serviced/var/volumes 2>/dev/null || true"
```
tgjonestx: This is not going to provide the expected output when executed on an HA system.
On the primary(active) node, our recommended size for this partition is 1GB.

On the secondary (passive) node of an HA cluster, the directory will exist, but it will not be mounted on a device, so most likely the df command will report the size of the root partition.

pkg/rpm/preupgrade
rphillipsz: The app_is_deployed() check should return false if serviced isn't running, skipping the tenant check in that case.

pkg/rpm/preupgrade
rphillipsz: I'm not sure.. this is getting the mounted volumes, and grepping for those with "/opt/serviced/var/volumes". If nothing's mounted, that list will be empty rather than somehow defaulting to the root partition?

tgjonestx: Fair enough.  I think the bottom line is that we need to run this on a HA deployment, especially when the cluster has been stopped per our standard instructions. In those cases, there are other checks that might potentially fail (e.g. the LV/VG will not be active)

tgjonestx:
We need to test this with HA systems.

PR #3478, opened 2017-04-04
ZEN-27266: add host IP addresses to filebeat config file

rderkach:
When running multiple remote hubs in resource pools, it is very
difficult to distinguish the source among the hub log records.
Add IP addresses of hosts to filebeat config file, which will
give ability to filter in Kibana by host ip.

[JIRA&DEMO](https://jira.zenoss.com/browse/ZEN-27266)

### DIFF
--- a/node/container.go
+++ b/node/container.go
@@ -502,6 +502,21 @@ func (a *HostAgent) getService(serviceID string) (*service.Service, error) {
 	return svc, nil
 }
 
+func(a *HostAgent) getServicePath(svc *service.Service) (string, error) {
+	var err error
+	s := svc
+	svc_path := fmt.Sprintf("/%s", svc.Name)
+	for s.ParentServiceID != "" {
+		s, err = a.getService(s.ParentServiceID)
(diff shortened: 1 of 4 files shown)


node/container.go
```diff
@@ -502,6 +502,21 @@ func (a *HostAgent) getService(serviceID string) (*service.Service, error) {
 	return svc, nil
 }
 
+func(a *HostAgent) getServicePath(svc *service.Service) (string, error) {
+	var err error
+	s := svc
+	svc_path := fmt.Sprintf("/%s", svc.Name)
+	for s.ParentServiceID != "" {
+		s, err = a.getService(s.ParentServiceID)
```
tgjonestx: This will have negative impacts on performance for customers with 1000s of services.  I have a suggestion on how to fix, but will send it via direct email because it's more involved than I can explain in a code-review comment.   However, the short version is that we need to refactor the implementation of GetEvaluatedService() as well the associated caching, so that it also returns information on the service path.

PR #3479, opened 2017-04-04
Adding Audit flag to service definition

svambati:


domain/servicedefinition/servicedefinition.go
```diff
@@ -136,6 +136,7 @@ type LogConfig struct {
 	Type    string   // Arbitrary string that identifies the "types" of logs that come from this source. This will be
 	Filters []string // A list of filters that must be contained in either the LogFilters or a parent's LogFilter,
 	LogTags []LogTag // Key value pair of tags that are sent to logstash for all entries coming out of this logfile
+	IsAudit bool
```
tgjonestx: nit pick - should include a comment that clarifies what setting this flag to true will do.

PR #3489, opened 2017-04-06
Feature/cc 3480

svambati:
Log Messages Looks like :

`2017-04-06 16:16:24 INFO action=Start kind=Daemon daemon=zentrap source=Shell process=zentrap.py logname=zenoss parameters="--configfile /opt/zenoss/etc/zentrap.conf -c --logfileonly --monitor localhost --useFileDescriptor=10"`
`2017-04-06 16:16:24 INFO action=Start kind=Daemon daemon=zensyslog source=Shell process=zensyslog.py logname=zenoss parameters="--configfile /opt/zenoss/etc/zensyslog.conf -c --logfileonly --monitor localhost --useFileDescriptor=6"`
`2017-04-06 16:16:24 INFO action=Start kind=Daemon daemon=zentrap source=Shell process=zentrap.py logname=zenoss parameters="--configfile /opt/zenoss/etc/zentrap.conf -c --logfileonly --monitor localhost --useFileDescriptor=10"
`


isvcs/container.go
```diff
@@ -352,7 +353,12 @@ func (svc *IService) create() (*docker.Container, error) {
 	// service-specific volumes
 	if svc.Volumes != nil && len(svc.Volumes) > 0 {
 		for src, dest := range svc.Volumes {
-			hostpath := svc.getResourcePath(src)
+			var hostpath string
+                        if src == "servicedLogDir" {
```
tgjonestx: Since `servicedLogDir` is a kind of key or symbolic, I suggest a global variable (defined at the top of this file), named something like `UseServicedLogDir`, and use that new variable in the Volumes definition in logstash.go as well.

facade/logstash.go
```diff
@@ -180,9 +183,34 @@ func getFilters(services []servicedefinition.ServiceDefinition, filterDefs map[s
 	return filters
 }
 
+func getAuditLogSection(services []servicedefinition.ServiceDefinition, auditTypes *[]string) string {
```
tgjonestx: Can you write some unit-test for this?  I can point you at some examples

The tests should cover several cases:
- no services
- a list of services, but no audit files
- a list of services with multiple audit files for the same audit type
- a list of services with multiple audit files for different audit types

utils/directory.go
```diff
@@ -76,5 +77,8 @@ func TempDir(p string) string {
 
 // ServicedLogDir gets the serviced log directory
 func ServicedLogDir() string {
-	return "/var/log/serviced"
+	if os.Getenv("SERVICED_LOG_PATH") != "" {
```
tgjonestx: you don't need this if condition. `config.GetOptions().LogPath` will give you the right path regardless of whether the environment variable was specified or if the default is used. The change you made to `cli/api/options.go` takes care of that for you.

pkg/serviced.default
```diff
@@ -79,6 +79,9 @@
 # Set the BACKUPS path for serviced backups
 # SERVICED_BACKUPS_PATH=/opt/serviced/var/backups
 
+# Set the LOG_PATH for serviced logs
```
tgjonestx: Should probably add some clarification. Something like  "for serviced access and audit logs. Note that regular serviced operational messages are written to journald."

facade/logstash_test.go
```diff
@@ -172,6 +267,46 @@ func TestConstructingFilterString(t *testing.T) {
...
+	}
+}
+func TestMultipleTypesForAuditLogging(t *testing.T){
+	services := make([]servicedefinition.ServiceDefinition, 1)
+	services[0] = getTestingServiceWithAuditEnabled()
+	auditableTypes := []string{}
+	auditLogSection := getAuditLogSection(services, &auditableTypes)
+	fieldTypeCount := strings.Count(auditLogSection, "if [fields][type]")
+	if fieldTypeCount !=2 {
+		t.Error(fmt.Printf("expected 2 for two different LogCoongfig Types, but found %d only", fieldTypeCount))
```
tgjonestx: should be fmt.Sprintf()

facade/logstash_test.go
```diff
@@ -172,6 +267,46 @@ func TestConstructingFilterString(t *testing.T) {
...
+	}
+}
+
+func TestGetAuditLogSectionForServicesNotEnabledAudit(t *testing.T) {
+	services := make([]servicedefinition.ServiceDefinition, 1)
+	services[0] = getTestingService()
+	auditableTypes := []string{}
+	auditLogSection := getAuditLogSection(services, &auditableTypes)
+	if len(auditLogSection) != 0 {
+		t.Error(fmt.Printf("expected am empty auditLogSection , but found %d size", len(auditLogSection)))
```
tgjonestx: should be fmt.Sprintf()

Also, if this test fails in the future, it would speed up debugging if the error message also printed the value of `auditLogSection`

facade/logstash_test.go
```diff
@@ -172,6 +267,46 @@ func TestConstructingFilterString(t *testing.T) {
 	}
 }
 
+func TestGetAuditLogSectionForNoservice(t *testing.T) {
+	services := []servicedefinition.ServiceDefinition{}
+	auditLogTypes := []string{}
+	auditLogSection := getAuditLogSection(services, &auditLogTypes)
+	if len(auditLogSection) > 0 {
+		t.Error(fmt.Sprintf("Audit Log Section should be empty but it is not %d", len(auditLogSection)))
```
tgjonestx: if this test fails in the future, it would speed up debugging if the error message also printed the value of `auditLogSection`

PR #3502, opened 2017-04-19
CC-3505: add host IP, service path and resource pool to filebeat cfg

rderkach:
When running multiple remote hubs in resource pools, it is very
difficult to distinguish the source among the hub log records.
Add IP addresses of hosts, service path and resource pool to
filebeat config file, which will give ability to filter
in Kibana by host ip or resource pool.

[JIRA&DEMO](https://jira.zenoss.com/browse/CC-3505)

### DIFF
--- a/facade/servicepath_cache.go
+++ b/facade/servicepath_cache.go
@@ -27,10 +27,11 @@ type serviceCache struct {
 }
 
 type servicePath struct {
-	serviceID   string
-	tenantID    string
-	parentID    string
-	servicePath string
+	serviceID       string
+	tenantID        string
+	parentID        string
+	servicePath     string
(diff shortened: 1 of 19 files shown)


facade/servicepath_cache.go
```diff
@@ -27,10 +27,11 @@ type serviceCache struct {
...
 type servicePath struct {
-	serviceID   string
-	tenantID    string
-	parentID    string
-	servicePath string
+	serviceID       string
+	tenantID        string
+	parentID        string
+	servicePath     string
+	serviceNamePath string
```
tgjonestx: Please update the unit tests in servicepath_cache_test.go to also verify the contents for serviceNamePath

PR #3508, opened 2017-04-26
Feature/CC-3376 - queryServices(name) no longer filters results

cchristiezen:
Backport for CC-3409

### DIFF
--- a/domain/service/servicedetails_store.go
+++ b/domain/service/servicedetails_store.go
@@ -16,6 +16,7 @@ package service
 import (
 	"errors"
 	"fmt"
+	"regexp"
 	"strings"
 	"time"
 	"unicode"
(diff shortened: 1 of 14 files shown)


domain/service/servicedetails_store.go
```diff
@@ -73,6 +62,32 @@ func (s *storeImpl) GetAllServiceDetails(ctx datastore.Context, since time.Durat
 		}
 
 		parentMap[d.ParentServiceID] = struct{}{}
+
+		if r != nil && !r.MatchString(d.Name) {
+			continue
+		}
+
+		if query.Since > 0 && d.UpdatedAt.Before(since) {
```
jplouis: Why not include the since in the elastic query if present? Would reduce the number of results.

domain/service/servicedetails_store.go
```diff
@@ -73,6 +62,32 @@ func (s *storeImpl) GetAllServiceDetails(ctx datastore.Context, since time.Durat
...
+			continue
+		}
+
+		if query.Tenants && d.ParentServiceID != "" {
```
jplouis: In general, any way to include the parameters into the actual elastic query?

cchristiezen: I started with that approach but encountered an issue with the name parameter. In the v1 endpoint, the name specified can be a regex and it used Go's regex library to evaluate.  The regex in elastic is evaluated differently so I was getting different results between the two for the same name.  I wanted to keep the same functionality as the v1 endpoint so I have to evaluate the regex server side in Go.  Since I had to do it server side and go through all the results anyway, I just added the other checks as well to make it more consistent.

jplouis: Makes sense for regex. However the "since" parameter can filter out a lot if it is being used by the client.  I'm thinking of issues with very large installs. So params that can be used should be used to filter at the elastic level.

PR #3509, opened 2017-04-28
Upgrade to filebeat 5.2.3

tgjonestx:
Fixes CC-3537

Includes some refactor to remove duplicate variables/properties for `/etc/filebeat.conf`.

Support for `idle-flush-time` argument was lost when we upgraded to filebeat last fall (not sure it every really worked anyway), so I removed it to lower the cruft factor. Also, I externalized the filebeat configuration into a template, so now if someone wants to adjust the idle-flush-time, they can modify the default value in the template file.

### DIFF
--- a/container/logstash_test.go
+++ b/container/logstash_test.go
@@ -1,5 +1,4 @@
 // Copyright 2014 The Serviced Authors.
-// Copyright 2014 The Serviced Authors.
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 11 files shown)


container/logstash_test.go
```diff
@@ -1,4 +1,5 @@
 // Copyright 2014 The Serviced Authors.
+// Copyright 2014 The Serviced Authors.
```
mozart27: unneeded dupe.

PR #3534, opened 2017-05-18
Feature/CC-3494 Add facilities for CC audit logging

corntoole:
This PR adds a new package to support audit logging of CC user actions. This new API is used audit log the addition and removal of Resource Pool objects.

### DIFF
--- a/audit/doc.go
+++ b/audit/doc.go
@@ -47,28 +47,24 @@
 	The default implementation that is returned is a wrapped logri Logger.  Logri is a package owned by Zenoss that adds additional
 	functionality to Loggers from the third party package, logrus.
 
-	Contextual information can be passed into the Logger through the Context method.  The Context method will take a Context interface that is defined in
-	this package.   Currently the context only takes a user.  In the future, additional contextual information can be added to this object.  A new context
-	object can be retrieved through the NewContext method.
+	To log an entry , a message along with relavant contextual information must be set on the audit logger through the use of the "Message" method.  This will set the "msg" and "user" fields:
 
-		var ctx = audit.NewContext("Alice")
+		auditLogger = auditLogger.Message(ctx, "Adding Resource Pool")
 
-	The context, like the other fields is optional.  If one is not provied the user will default to "system".  To set the Context call the following.
-
-		auditLogger = auditLogger.Context(ctx)
-
-	A friendly message can be set on the audit logger through the use of the "Message" method.  This will set the "msg" field:
-
-		auditLogger = auditLogger.Message("Adding Resource Pool: " + entity.ID)
+	The Message method will take an implementation of the "datastore.context" interface.  Currently the audit logger only uses user field.  In the future, additional contextual information maybe utilized.
 
 	The "action field is set through the "Action" method.  Some constants are provided in this package to normalized the values.  Examples are "add", "update",
 	"delete", "start", "stop", etc.  If addition actions need to be added, new constants should be added to this package.
 
 		auditLogger = auditLogger.Action(audit.Add)
 
-	The type of entity being modified is set through the "Type" method.  To normalize these values, constants found in the domain package should be used.
+	The interface, "Entity", is added to the "datastore" package to provide a convenient method for encapsulating a domain object's ID and type. The "type" and "ID" of an audit log entry can be set via the "Entity" method. "Entity" is the preferred method for setting an audit log entry's "id" and "type", but if an implementation of the "Entity" is not available, the methods "ID" and "Type" are available.
 
-		auditLogger = auditLogger.Type(domain.ResourcePoolType)
+		auditLogger = auditLogger.Entity(entity)
+
+	The type of entity being modified is set through the "Type" method.  To get normalized type values for a domain object, use the package method "GetType" for a given domain object.
+
+		auditLogger = auditLogger.Type(pool.GetType())
 
 	To set the "id" field, use the "ID" method.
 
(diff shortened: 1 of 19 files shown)


audit/logger.go
```diff
@@ -0,0 +1,140 @@
...
+
+	// Log whether the action succeeded or failed based on the error passed in.
+	Error(err error) error
+}
+
+// NewLogger returns a default implmentation of the audit logger.  The "user" will default
+// to "system" unless otherwise specified in the context.  This wraps a logri Logger,
+// and will write to the location specified in the logger config.
+func NewLogger() Logger {
+	entry := logri.GetLogger("audit").WithField("user", "system")
```
tgjonestx: I don't think we need to the user field here. I think we should let the `Message()` method always get the user from the context, and let the context create the default user. Basically, I don't want the default user to be defined in more than one place - the context object should be the system of record for that.

audit/doc.go
```diff
@@ -0,0 +1,204 @@
...
+	API
+
+	The Logger interface provides a fluent API for creating log entries.  A new Logger can be retrieved by calling the NewLogger method.
+
+		var auditLogger = audit.NewLogger()
+
+	The default implementation that is returned is a wrapped logri Logger.  Logri is a package owned by Zenoss that adds additional
+	functionality to Loggers from the third party package, logrus.
+
+	Contextual information can be passed into the Logger through the Context method.  The Context method will take a Context interface that is defined in
```
tgjonestx: This is no longer true

audit/doc.go
```diff
@@ -0,0 +1,204 @@
...
+	this package.   Currently the context only takes a user.  In the future, additional contextual information can be added to this object.  A new context
+	object can be retrieved through the NewContext method.
+
+		var ctx = audit.NewContext("Alice")
+
+	The context, like the other fields is optional.  If one is not provied the user will default to "system".  To set the Context call the following.
+
+		auditLogger = auditLogger.Context(ctx)
```
tgjonestx: This discussion of the Context() method is out of date too

audit/doc.go
```diff
@@ -0,0 +1,204 @@
...
+
+	A friendly message can be set on the audit logger through the use of the "Message" method.  This will set the "msg" field:
+
+		auditLogger = auditLogger.Message("Adding Resource Pool: " + entity.ID)
```
tgjonestx: Needs a `ctx` argument

audit/doc.go
```diff
@@ -0,0 +1,204 @@
...
+
+	The "action field is set through the "Action" method.  Some constants are provided in this package to normalized the values.  Examples are "add", "update",
+	"delete", "start", "stop", etc.  If addition actions need to be added, new constants should be added to this package.
+
+		auditLogger = auditLogger.Action(audit.Add)
+
+	The type of entity being modified is set through the "Type" method.  To normalize these values, constants found in the domain package should be used.
```
tgjonestx: The discussion should start with the `Entity()` method, and then mention `Type()` and `ID()` as alternatives if the caller can't use `Entity()`

domain/common.go
```diff
@@ -70,3 +70,9 @@ type Prereq struct {
 	Name   string
 	Script string
 }
+
+const (
```
tgjonestx: We shouldn't duplicate these strings (here and in each of the `kind` variables in the various entity-specific subdirectories).  

I suggest adding a GetType() method to each domain package in addition to the GetType() for each domain object. For instance, something like this for ResourcePool:
```
func (p *ResourcePool) GetType() string {
	return GetType()
}


func GetType() string {
	return kind
}
```

audit/doc.go
```diff
@@ -0,0 +1,204 @@
...
+			...
+		}
+
+	In this case, the err variable will still have the same value as the previous call because the error returned by "DoSomething" is just passed through.
+	Success or failure will be logged depending on the value of the error that is returned from the "DoSomething" method.
+
+	The API is designed to be fluent so the method calls can be chained together.  For example, setting the fields and logging success can be done
+	at the same time.  The following sets the context, message, action, id, type, and success:
+
+		auditLogger.Context(ctx).
```
tgjonestx: No more `Context(ctx)`

facade/testutils.go
```diff
@@ -71,6 +72,16 @@ func (ft *FacadeIntegrationTest) SetUpSuite(c *gocheck.C) {
 	ft.CTX = datastore.Get()
 
 	ft.Facade = New()
+	mockLogger := &auditmocks.Logger{}
+	mockLogger.On("Context", mock.AnythingOfType("*datastore.context")).Return(mockLogger)
```
tgjonestx: The `Context` method needs to be removed.

facade/testutils.go
```diff
@@ -71,6 +72,16 @@ func (ft *FacadeIntegrationTest) SetUpSuite(c *gocheck.C) {
...
+	mockLogger.On("Message", mock.AnythingOfType("string")).Return(mockLogger)
```
tgjonestx: This mock is wrong because `Message()` now takes 2 arguments not 1.

audit/logger.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func (l *logger) addField(name string, value string) Logger {
+    if l.entry != nil {
+        l.entry = l.entry.WithField(name, value)
+    } else {
+        l.entry = l.loggeri.WithField(name, value)
+        l.entry.Logger.Hooks = make(map[logrus.Level][]logrus.Hook)
+    }
+    return l
```
tgjonestx: Sorry I missed this when I first suggested, but this could a little DRYer by having `addField` just call `addFields` with a list of one item:
```
    return l.addFields(logrus.Fields{name, value})
```

domain/pool/pool.go
```diff
@@ -153,3 +153,22 @@ func (a *ResourcePool) HasDfsAccess() bool {
...
+// GetID returns its ResourcePool's ID.
+// It return the ID as a string
+func (a *ResourcePool) GetID() string {
+	return a.ID
+}
+
+// GetType return a ResourcePool's Entity type or kind.
+// It returns the Kind as a string.
+func (a *ResourcePool) GetType() string {
+	return kind
```
tgjonestx: One other DRY comment - just `return GetType()`

PR #3535, opened 2017-05-23
Feature CC-3494 pt. 2: CC Audit Logging API

corntoole:
Change methods back to previous behavior of returning new Logger instances, so that the audit API can be used as specified in documentation.

### DIFF
--- a/audit/logger.go
+++ b/audit/logger.go
@@ -81,8 +81,9 @@ func (l *logger) Action(action string) Logger {
 }
 
 func (l *logger) Message(ctx datastore.Context, message string) Logger {
-	l.message = message
-	return l.newLoggerWith("user", ctx.User())
+	result := l.newLoggerWith("user", ctx.User())
+	result.message = message
+	return result
 }
 
 func (l *logger) Type(theType string) Logger {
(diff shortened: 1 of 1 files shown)


audit/logger.go
```diff
@@ -77,32 +77,38 @@ type logger struct {
...
 func (l *logger) Action(action string) Logger {
-	return l.addField("action", action)
+	return l.newLoggerWith("action", action)
 }
 
 func (l *logger) Message(ctx datastore.Context, message string) Logger {
-	l.addField("user", ctx.User())
 	l.message = message
-	return l
+	return l.newLoggerWith("user", ctx.User())
```
tgjonestx: Setting message first may 'break' the notion here that you get a new object. Instead, you can change the return type of `newLoggerWith` to be `*logger`, then you can could do:
```
newLogger := l.newLoggerWith("user", ctx.User())
newLogger.message = message
return newLogger
```

audit/logger.go
```diff
@@ -77,32 +77,38 @@ type logger struct {
...
-	return l.addField("id", id)
+	return l.newLoggerWith("id", id)
 }
 
 func (l *logger) Entity(entity datastore.Entity) Logger {
-	return l.addFields(logrus.Fields{
-		"id":   entity.GetID(),
-		"type": entity.GetType(),
-	})
+	l.addField("id", entity.GetID())
```
tgjonestx: Ditto previous comment

audit/logger.go
```diff
@@ -77,32 +77,38 @@ type logger struct {
...
+	return l.newLoggerWith("type", entity.GetType())
 }
 
 func (l *logger) WithField(name string, value string) Logger {
-	return l.addField(name, value)
+	return l.newLoggerWith(name, value)
```
tgjonestx: Unless I missed something, `WithFields` needs to be changed as well.

audit/logger.go
```diff
@@ -77,32 +77,38 @@ type logger struct {
...
+}
+
+func (l *logger) newLoggerWith(name string, value string) Logger {
```
tgjonestx: Per comment above, change return type to `*logger` so callers in this package can access all of the fields of `logger`, like `message`

PR #3550, opened 2017-06-05
CC-3599: add audit log entries for public endpoint changes

corntoole:


facade/publicendpoint.go
```diff
@@ -145,31 +160,44 @@ func (f *Facade) RemovePublicEndpointPort(ctx datastore.Context, serviceid, endp
...
+	if isEnabled {
+		auditMsg = "Enabling Public Endpoint Port"
+	} else {
+		auditMsg = "Disabling Public Endpoing Port"
+	}
+	alog := f.auditLogger.Message(ctx, auditMsg).Action(audit.Update).ID(serviceid).
+		WithFields(logrus.Fields{
+			"endpointname": endpointName,
+			"portaddr": portAddr,
+			"isenabled": isEnabled,
```
svambati: Since, we are showing appropriate message, I think `isenabled` field is not necessary.

PR #3553, opened 2017-06-07
Leave DM_BASESIZE unspecified if not given in conf.

jpeacock-zenoss:
This change sets DM_BASESIZE (or "dm.basesize") as an unspecified option to allow code to determine whether a default base size value should be used.  Having DM_BASESIZE automatically acquire a value when it's not given in the config file means that other code has to assume that the value is specified and not a default.

Also changed the associated message when a specified DM_BASESIZE is less than the current base size so the message displays values in a multiple of 1024 (e.g. "100 GiB" instead of "107.4 GB").

Fixes CC-3522.

### DIFF
--- a/cli/api/storageoptions.go
+++ b/cli/api/storageoptions.go
@@ -36,7 +36,7 @@ func getDefaultStorageOptions(driverType volume.DriverType, config utils.ConfigR
 		addStorageOption(config, "DM_THINPOOLDEV", "", func(v string) {
 			options = append(options, fmt.Sprintf("dm.thinpooldev=%s", v))
 		})
-		addStorageOption(config, "DM_BASESIZE", "100G", func(v string) {
+		addStorageOption(config, "DM_BASESIZE", "", func(v string) {
 			options = append(options, fmt.Sprintf("dm.basesize=%s", v))
 		})
 		addStorageOption(config, "DM_LOOPDATASIZE", "", func(v string) {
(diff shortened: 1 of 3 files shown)


volume/devicemapper/devmapper/deviceset.go
```diff
@@ -1127,7 +1127,7 @@ func (devices *DeviceSet) checkGrowBaseDeviceFS(info *devInfo) error {
 	}
 
 	if devices.baseFsSize < devices.getBaseDeviceSize() {
-		return fmt.Errorf("devmapper: Base device size cannot be smaller than %s", units.HumanSize(float64(devices.getBaseDeviceSize())))
+		return fmt.Errorf("devmapper: Base device size cannot be smaller than %s", units.BytesSize(float64(devices.getBaseDeviceSize())))
```
iancmcc: Will the string output here match the spelling of the viable config option parsed by Docker?

cli/api/storageoptions.go
```diff
@@ -36,7 +36,7 @@ func getDefaultStorageOptions(driverType volume.DriverType, config utils.ConfigR
 		addStorageOption(config, "DM_THINPOOLDEV", "", func(v string) {
 			options = append(options, fmt.Sprintf("dm.thinpooldev=%s", v))
 		})
-		addStorageOption(config, "DM_BASESIZE", "100G", func(v string) {
+		addStorageOption(config, "DM_BASESIZE", "", func(v string) {
```
iancmcc: The default Docker specifies is 10G. We're overriding their default by doing this. Are default devices now 10G?

jpeacock-zenoss: I thought https://github.com/control-center/serviced/blob/develop/volume/devicemapper/devicemapper.go#L317 would set it to 100G anyway, but rechecked by base size and found `10737418240` which looks similar to `107374182400` if you're not paying attention.

PR #3554, opened 2017-06-08
Have the size displayed as traditional 1024 units.

jpeacock-zenoss:
The go-units library used for handling sizing units, won't use binary (or 1024) sizing with the "MB", "GB", etc. labeling by default.  This is worked around by using the CustomSize function.

### DIFF
--- a/volume/devicemapper/devmapper/deviceset.go
+++ b/volume/devicemapper/devmapper/deviceset.go
@@ -198,6 +198,8 @@ type ThinpoolInitError struct {
 	msg string
 }
 
+var sizeAbbrs = []string{"B", "kB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB"}
+
 func (e ThinpoolInitError) Error() string {
 	return e.msg
 }
@@ -1127,7 +1129,9 @@ func (devices *DeviceSet) checkGrowBaseDeviceFS(info *devInfo) error {
 	}
 
 	if devices.baseFsSize < devices.getBaseDeviceSize() {
-		return fmt.Errorf("devmapper: Base device size cannot be smaller than %s", units.HumanSize(float64(devices.getBaseDeviceSize())))
+		return fmt.Errorf(
+			"devmapper: Base device size cannot be smaller than %s",
+			units.CustomSize("%.4g%s", float64(devices.getBaseDeviceSize()), 1024.0, sizeAbbrs))
 	}
 
 	if devices.baseFsSize == devices.getBaseDeviceSize() {


volume/devicemapper/devmapper/deviceset.go
```diff
@@ -1127,7 +1129,9 @@ func (devices *DeviceSet) checkGrowBaseDeviceFS(info *devInfo) error {
 	}
 
 	if devices.baseFsSize < devices.getBaseDeviceSize() {
-		return fmt.Errorf("devmapper: Base device size cannot be smaller than %s", units.HumanSize(float64(devices.getBaseDeviceSize())))
+		return fmt.Errorf(
+			"devmapper: Base device size cannot be smaller than %s",
+			units.CustomSize("%.4g%s", float64(devices.getBaseDeviceSize()), 1024.0, sizeAbbrs))
```
iancmcc: Can you confirm that when Docker parses this in the config file, it treats it as gibibytes rather than gigabytes?

jpeacock-zenoss: t.go
```
package main

import (
	"fmt"
	"github.com/docker/go-units"
)

var sizeAbbrs = []string{"B", "kB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB"}
var gigs int64 = 1024 * 1024 * 1024

func main() {
	fmt.Println(units.HumanSize(float64(100 * gigs)))
	fmt.Println(units.BytesSize(float64(100 * gigs)))
	if size, err := units.RAMInBytes("100GB"); err != nil {
		fmt.Printf("failed: %s\n", err)
	} else {
		fmt.Println(units.BytesSize(float64(size)))
	}
	if size, err := units.RAMInBytes("100GiB"); err != nil {
		fmt.Printf("failed: %s\n", err)
	} else {
		fmt.Println(units.BytesSize(float64(size)));
	}
	fmt.Println(units.CustomSize("%.4g%s", float64(100 * gigs), 1024.0, sizeAbbrs))
}
```
$ go build t.go
$ ./t
107.4 GB
100 GiB
100 GiB
failed: invalid size: '100GiB'
100GB
```
I think this is what you're asking for?

iancmcc: This, actually: https://github.com/moby/moby/blob/05ad14fc1bc5626895a419a991fcf8932eb91d27/daemon/graphdriver/devmapper/deviceset.go#L2648
and
https://github.com/moby/moby/blob/05ad14fc1bc5626895a419a991fcf8932eb91d27/daemon/graphdriver/devmapper/deviceset.go#L40

It's RAMInBytes, so you're good.

PR #3562, opened 2017-06-20
CC-3564: Improve error handling for assigning virtual IPs

cchristiezen:
Changes to prevent logs from being spammed.  Also, additional fixes for wiring up dependencies and initializing a single pool listener.

### DIFF
--- a/zzk/service/assignment.go
+++ b/zzk/service/assignment.go
@@ -37,7 +37,7 @@ var (
 // AssignmentHandler is used to assign, unassign, and watch virtual IP assignments
 // to hosts
 type AssignmentHandler interface {
-	Assign(poolID, ipAddress, netmask, binding string, cancel <-chan interface{}) error
+	Assign(poolID, ipAddress, netmask, binding string) error
 	Unassign(poolID, ipAddress string) error
 }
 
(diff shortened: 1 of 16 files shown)


zzk/service/mocks/RegisteredHostHandler.go
```diff
@@ -22,19 +22,19 @@ type RegisteredHostHandler struct {
 	mock.Mock
 }
 
-func (_m *RegisteredHostHandler) GetRegisteredHosts(cancel <-chan interface{}) ([]host.Host, error) {
-	ret := _m.Called(cancel)
+func (_m *RegisteredHostHandler) GetRegisteredHosts(pool string) ([]host.Host, error) {
```
tgjonestx: ditto my earlier concern about dropping the `cancel` argument.  I'm not familiar with how this is all _supposed_ to work, so maybe it's OK.

zzk/service/assignment.go
```diff
@@ -136,48 +132,102 @@ func (h *ZKAssignmentHandler) getAssignedHostID(poolID, ipAddress string) (strin
...
+	hostIDs := []string{}
+
+	for _, h := range hosts {
+		hostIDs = append(hostIDs, h.ID+":"+h.PoolID)
+	}
+
+	return strings.Join(hostIDs, ",")
+}
+
 func (h *ZKAssignmentHandler) assignToHost(poolID, ipAddress, netmask, binding string, cancel <-chan interface{}) error {
```
cchristiezen: That is intentional.  In order to fix and clean up the dependencies, the pool listener no longer uses the host registry listener to look for hosts.  It used to block until a host was online, but now it polls every 5s for a host so the channel is no longer needed.

PR #3565, opened 2017-06-23
Add support for --no-children option on log export

tgjonestx:
Fixes CC-3627

The following export logs for all services in Zenoss.remsgr:
```
serviced log export --service Zenoss.resmgr
```
Using the new option added this in the PR, the following will export logs *only* for the Zenoss.resmgr service:
```
serviced log export --service Zenoss.resmgr --no-children
```


tgjonestx:
ian is cool

PR #3570, opened 2017-06-29
reload logstash configuration after service is edited

tgjonestx:
Fixes CC-3646

### DIFF
--- a/dao/elasticsearch/service.go
+++ b/dao/elasticsearch/service.go
@@ -75,13 +75,19 @@ func (this *ControlPlaneDao) CloneService(request dao.ServiceCloneRequest, clone
 
 //
 func (this *ControlPlaneDao) UpdateService(svc service.Service, unused *int) error {
-	if err := this.facade.DFSLock(datastore.Get()).LockWithTimeout("update service", userLockTimeout); err != nil {
+	ctx := datastore.Get()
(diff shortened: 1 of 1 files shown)


dao/elasticsearch/service.go
```diff
@@ -75,13 +75,19 @@ func (this *ControlPlaneDao) CloneService(request dao.ServiceCloneRequest, clone
...
 		return err
 	}
-	defer this.facade.DFSLock(datastore.Get()).Unlock()
+	defer this.facade.DFSLock(ctx).Unlock()
 
-	return this.facade.UpdateService(datastore.Get(), svc)
+	err := this.facade.UpdateService(ctx, svc)
+	if err == nil {
+		// CC-3646 - rebuild logstash config in case the set of auditable log files has changed
+		this.facade.ReloadLogstashConfig(ctx)
```
rphillipsz: Should this be done in the facade, so that it's also done if called the from other places in code that update the service definition?

tgjonestx: Generally speaking, using the facade is the better approach, but the both `UpdateService()` and `updateService()` methods in the facade are re-used for lots of other use-cases like migration which already has a reload at the very end (rather than reloading for each separate service, which is what would happen if I moved this call down into one of those methods).   So this was the place where it was most clearly just for `serviced service edit`

PR #3556, opened 2017-06-12
CC-3615 add remove-ip and set-ip commands

ssOlexBaiko:
Fix provides two additional commands to cli package to remove and set new IP Assignment.
[Jira](https://jira.zenoss.com/browse/CC-3615)

PR #3573, opened 2017-07-03
CC-2377 add remove-ip and set-ip commands

ssOlexBaiko:
port from https://github.com/control-center/serviced/pull/3556
[Jira](https://jira.zenoss.com/browse/CC-2377)


dao/interface.go
```diff
@@ -161,6 +161,12 @@ type ControlPlane interface {
 	// Assign IP addresses to all services at and below the provided service
 	AssignIPs(assignmentRequest addressassignment.AssignmentRequest, unused *int) (err error)
 
+	// Remove the IP assignment of a service's endpoints
```
jplouis: Should not be adding new methods to the DAO package unless necessary

rpc/master/service_server.go
```diff
@@ -152,3 +154,11 @@ func (s *Server) ClearEmergency(serviceID string, count *int) error {
 	*count = c
 	return nil
 }
+
+func (s *Server) RR(args []string) error {
```
tgjonestx: This method needs to be named `RemoveIPs`

rpc/master/service_client.go
```diff
@@ -109,3 +110,15 @@ func (c *Client) ClearEmergency(serviceID string) (int, error) {
...
+// Remove the IP assignment of a service's endpoints
+func (c *Client) RemoveIPs(args []string) (err error) {
+	err = c.call("RemoveIPs", args, new(string))
+	return
+}
+
+// Assigns an IP address to a services that haven't IP Assignment by default
+func (c *Client) SetIPs(r addressassignment.AssignmentRequest) (err error) {
+	err = c.call("SetIPs", r, new(string))
+	return
```
jplouis: Why aren't the errors returned in these 2 methods?

rpc/master/service_server.go
```diff
@@ -152,3 +153,13 @@ func (s *Server) ClearEmergency(serviceID string, count *int) error {
...
+
+func (s *Server) RemoveIPs(args []string, unused *string) (err error) {
+	err = s.f.RemoveIPs(s.context(), args)
+        return
+}
+
+func (s *Server) SetIPs(request addressassignment.AssignmentRequest, unused *string) (err error) {
+	err = s.f.SetIPs(s.context(), request)
+        return
+}
```
jplouis: Why aren't the errors returned in these 2 methods?

rpc/master/service_client.go
```diff
@@ -109,3 +110,15 @@ func (c *Client) ClearEmergency(serviceID string) (int, error) {
 	err := c.call("ClearEmergency", serviceID, &affected)
 	return affected, err
 }
+
+// Remove the IP assignment of a service's endpoints
+func (c *Client) RemoveIPs(args []string) (err error) {
+	err = c.call("RemoveIPs", args, new(string))
```
jplouis: Please change these methods to "return c.call("RemoveIPs", args, new(string))" The implicit return of the error is more verbose in this case.  Same is true for the other methods doing this.

rpc/master/service_server.go
```diff
@@ -152,3 +153,13 @@ func (s *Server) ClearEmergency(serviceID string, count *int) error {
 	*count = c
 	return nil
 }
+
+func (s *Server) RemoveIPs(args []string, unused *string) (err error) {
+	err = s.f.RemoveIPs(s.context(), args)
```
jplouis: These as well

PR #3578, opened 2017-07-05
CC-3174: Multiline Log Parsing Does not Work

olst:


olst:
retest this

facade/logstash.go
```diff
@@ -215,7 +215,7 @@ func getFilters(configs []servicedefinition.LogConfig, filterDefs map[string]str
 		for _, filtName := range config.Filters {
 			//do not write duplicate types, logstash doesn't handle this
 			if !utils.StringInSlice(config.Type, *typeFilter) {
-				filters += fmt.Sprintf("\n  if [type] == \"%s\" {\n    %s\n  }\n",
+				filters += fmt.Sprintf("\n  if \"%s\" in [file] {\n    %s\n  }\n",
```
tgjonestx: Can you explain how this solves the problem?

Typically, the `file` attribute in logstash and the value of `config.Type` are not the same at all.

olst: We have this issue because in logstash.conf `if` statements like `if [type] == "zenmodeler"` don't work. We don't have multiline error parsing issue if I delete all those ^ conditions or if I use conditions like:
`if [file] == "/opt/zenoss/log/zenmodeler.log"`
`if "zenmodeler" in "/opt/zenoss/log/zenmodeler.log"`
`if "zenmodeler" in [file]`
This is my current logstash.conf:
https://gist.github.com/olst/971858d0dcd5c03cbf5d3765f797d3fd
issue:
https://jira.zenoss.com/secure/attachment/42765/42765_zenmodelerUI.png
fix:
https://jira.zenoss.com/secure/attachment/42928/42928_bbc.png

tgjonestx: Your solution assumes that the value of `config.Type` is always found in the file path defined by `[file]`, but that's not true in all cases. It's true sort of by accident for many case, but will not in the most general case. 
For instance, these lines will never produce a positive match:
```
if "zenossaudit" in [file] {
if "zope_eventlog" in [file] {
```
because the file for type=zenossaudit is named `/opt/zenoss/log/audit.log`, and the file for type=zenossaudit is named `/opt/zenoss/log/event.log`

I think you need to use something like `if [file] == "/opt/zenoss/log/zenmodeler.log"`, but you need to look for 2 things in the final output:
* there should be only one "if" stanza for "/opt/zenoss/log/audit.log" (this file appears in multiple services)
* there should be an "if" stanza for "/opt/zenoss/log/event.log"  and ""/opt/zenoss/log/Z2.log" (both of these log files are defined by the Zope service).

olst:
These two things:
 - there should be only one "if" stanza for "/opt/zenoss/log/audit.log" (this file appears in multiple services)
 - there should be an "if" stanza for "/opt/zenoss/log/event.log" and ""/opt/zenoss/log/Z2.log" (both of these log files are defined by the Zope service).
are already covered by these loops:
https://github.com/control-center/serviced/blob/626e68865e51f90e8c6afbcfb75346ee0738ed4c/facade/logstash.go#L214-L215
this my `logstash.conf` with latest changes:
https://gist.github.com/olst/76d39421a036a6c5219840f5e26b940e

tgjonestx:
retest this

PR #3583, opened 2017-07-12
ZEN-26148 zenossdbpack cron job is installed with each serviced instance

romanbakaleyko:


romanbakaleyko:
This should work on a clean install, the question I have here is how to update scripts for upgrades ?

PR #3585, opened 2017-07-14
Fix up some logstash filtering stuff

tgjonestx:
Fixes CC-3669

* Do not write multiple sections to logstash.conf for the same file; e.g. Zope and zenreports both have a Z2.log, but logstash.conf should have a single filter for that file
* Do not write empty sections to the logstash.conf like
```
    if [file] == "/opt/zenoss/log/zenmail.log" {
    
    }
```
* Add `rcvd_datetime` to each message to record when logstash received the message, as opposed to `@timestamp` which should record the time parsed out of the message itself.

### DIFF
--- a/facade/logstash.go
+++ b/facade/logstash.go
@@ -125,19 +125,20 @@ func writeLogstashConfiguration(templates map[string]servicetemplate.ServiceTemp
 	// filters will be a syntactically correct logstash filters section
 	filters := ""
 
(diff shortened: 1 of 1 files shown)


tgjonestx:
retest this

tgjonestx:
retest this.

tgjonestx:
retest this

PR #3588, opened 2017-07-17
Feature/cc 3536

rphillipsz:
https://jira.zenoss.com/browse/CC-3536

Under some circumstances (probably after a restore), serviced will switch tenant devices but will be unable to unmount the previous device from the NFS export mount. When this happens, we end up in a situation where a different device is used by the master vs. the delegates. Typically, this means that the application continues to use the old device, while backups, shell, and run commands are applied to the new device. This can go on for a long time undetected, until the system ends up in a state where there are no recent backups that contain any useful data.
This is an invalid state, and as such serviced should refuse to start services if it is in this state.


rphillipsz:
retest this, please

rphillipsz:
retest this

rphillipsz:
retest this

PR #3597, opened 2017-07-26
Add in support for multiple health checks for ISVCs.

zedieck:
https://jira.zenoss.com/browse/CC-3676

zk.go:
- Created a separate health check for quorum.
- modified the creation of health checks to do so for each ZK member from config file.
- optimized the health check functions to return immediately so multiple health checks don't stall
- cleaned up logging in the health status functions so we get a more clear indication in the logs about what is happening when something fails a health check.

container.go:
- doHealthChecks() now iterates over all healthchecks instead of just hitting the default.
- startupHealthcheck() now takes into account all the healthchecks before starting up not just the default.
- setHeathStatus() now takes the key for the health check so that health check status gets set properly.
- setStoppedHealthStatus() now properly sets stopped state for all health statues instead of just the default health check.

### DIFF
--- a/isvcs/container.go
+++ b/isvcs/container.go
@@ -769,41 +769,45 @@ func (svc *IService) startupHealthcheck() <-chan error {
 	go func() {
 		var result error
 		if len(svc.HealthChecks) > 0 {
-			checkDefinition, found := svc.HealthChecks[DEFAULT_HEALTHCHECK_NAME]
-			if !found {
-				log.WithFields(logrus.Fields{
-					"healthcheck": DEFAULT_HEALTHCHECK_NAME,
-					"isvc":        svc.Name,
-				}).Debug("Default health check not found")
-				err <- nil
-				return
-			}
 
 			startCheck := time.Now()
+			breakCheck := false
+
 			for {
 				currentTime := time.Now()
 				elapsed := time.Since(startCheck)
-				result = svc.runCheckOrTimeout(checkDefinition)
-				svc.setHealthStatus(result, currentTime.Unix())
-				log := log.WithFields(logrus.Fields{
-					"isvc":    svc.Name,
-					"elapsed": elapsed,
-				})
-				if result == nil {
-					log.Info("Internal service checked in healthy")
-					break
-				} else if elapsed.Seconds() > svc.StartupTimeout.Seconds() {
-					log.WithFields(logrus.Fields{
-						"lastresult": result,
-					}).Warn("Unable to verify health of internal service")
+
+				for healthCheckIndex, healthCheckDefinition := range svc.HealthChecks {
+					result = svc.runCheckOrTimeout(healthCheckDefinition)
+					svc.setHealthStatus(result, currentTime.Unix(), healthCheckIndex)
+
+					log := log.WithFields(logrus.Fields{
+						"isvc":    svc.Name,
+						"elapsed": elapsed,
+					})
+
+					if result == nil {
+						log.Info("Internal service checked in healthy")
+						breakCheck = true
+					} else if elapsed.Seconds() > svc.StartupTimeout.Seconds() {
+						log.WithFields(logrus.Fields{
+							"lastresult": result,
+						}).Warn("Unable to verify health of internal service")
+						breakCheck = true
+					}
+				}
+
+				if breakCheck == true {
 					break
 				}
+
 				log.Debug("Waiting for internal service to check in healthy")
+
 				time.Sleep(time.Second)
 			}
 			err <- result
 		} else {
-			svc.setHealthStatus(nil, time.Now().Unix())
+			svc.setHealthStatus(nil, time.Now().Unix(), DEFAULT_HEALTHCHECK_NAME)
 			err <- nil
 		}
 	}()
@@ -831,26 +835,24 @@ func (svc *IService) runCheckOrTimeout(checkDefinition healthCheckDefinition) er
 
 func (svc *IService) doHealthChecks(halt <-chan struct{}) {
 	if len(svc.HealthChecks) == 0 {
-		return
-	}
-
-	var (
-		found           bool
-		checkDefinition healthCheckDefinition
-	)
-
-	log := log.WithFields(logrus.Fields{
-		"isvc": svc.Name,
-	})
-
-	if checkDefinition, found = svc.HealthChecks[DEFAULT_HEALTHCHECK_NAME]; !found {
 		log.WithFields(logrus.Fields{
-			"healthcheck": DEFAULT_HEALTHCHECK_NAME,
-		}).Warn("Default health check not found")
+			"healthcheck": "None",
+		}).Debug("No health checks not found")
 		return
 	}
 
-	timer := time.Tick(checkDefinition.Interval)
+	var tickInterval time.Duration
+
+	// go through the health check s and find the longest duration interval for our ticker.
+	// CURRENT - we have one ticker for the longest of the periods for all health checks.
+	for _, checkDefinition := range svc.HealthChecks {
+		if tickInterval < checkDefinition.Interval {
+			tickInterval = checkDefinition.Interval
+		}
+	}
+
+	timer := time.Tick(tickInterval)
+
 	for {
 		select {
 		case <-halt:
@@ -858,16 +860,18 @@ func (svc *IService) doHealthChecks(halt <-chan struct{}) {
 			return
 
 		case currentTime := <-timer:
-			err := svc.runCheckOrTimeout(checkDefinition)
-			svc.setHealthStatus(err, currentTime.Unix())
-			if err != nil {
-				log.WithError(err).Warn("Health check failed")
+			for healthCheckIndex, checkDefinition := range svc.HealthChecks {
+				err := svc.runCheckOrTimeout(checkDefinition)
+				svc.setHealthStatus(err, currentTime.Unix(), healthCheckIndex)
+				if err != nil {
+					log.WithError(err).Warn("Health check failed")
+				}
 			}
 		}
 	}
 }
 
-func (svc *IService) setHealthStatus(result error, currentTime int64) {
+func (svc *IService) setHealthStatus(result error, currentTime int64, healthCheckIndex string) {
 	if len(svc.healthStatuses) == 0 {
 		return
 	}
@@ -879,7 +883,7 @@ func (svc *IService) setHealthStatus(result error, currentTime int64) {
 		"isvc": svc.Name,
 	})
 
-	if healthStatus, found := svc.healthStatuses[DEFAULT_HEALTHCHECK_NAME]; found {
+	if healthStatus, found := svc.healthStatuses[healthCheckIndex]; found {
 		if result == nil {
 			if healthStatus.Status != "passed" && healthStatus.Status != "unknown" {
 				log.Info("Internal service checked in healthy")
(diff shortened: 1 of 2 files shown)


isvcs/zk.go
```diff
@@ -61,82 +74,97 @@ func initZK() {
...
-		healthCheck: zkHealthCheck,
-		Interval:    DEFAULT_HEALTHCHECK_INTERVAL,
-		Timeout:     DEFAULT_HEALTHCHECK_TIMEOUT,
-	}
+	// setup the health check definitions
+	Zookeeper.HealthChecks 	= make(map[string]healthCheckDefinition)
 
-	Zookeeper.HealthChecks = map[string]healthCheckDefinition{
-		DEFAULT_HEALTHCHECK_NAME: defaultHealthCheck,
+	for _, zkIP := range zkConnectStrings {
```
tgjonestx: This loop is wrong. In cases where zkConnectStrings has 3 or more stanzas, the value used for the isvc is going to be the _last_ entry, but it should be the entry for the localhost.   Thinking a little more about this, I don't believe that the we should make assumptions about the order of IPs in the `SERVICED_ZK` configuration string. For the purposes of running a healthcheck against the *local* ZK instance, we should always use `127.0.0.1:2181` because that's the only value we can count on to find the local ZK instance.

isvcs/container.go
```diff
@@ -769,41 +769,45 @@ func (svc *IService) startupHealthcheck() <-chan error {
...
+				for healthCheckIndex, healthCheckDefinition := range svc.HealthChecks {
+					result = svc.runCheckOrTimeout(healthCheckDefinition)
+					svc.setHealthStatus(result, currentTime.Unix(), healthCheckIndex)
+
+					log := log.WithFields(logrus.Fields{
+						"isvc":    svc.Name,
+						"elapsed": elapsed,
+					})
+
+					if result == nil {
```
tgjonestx: I don't think the `Info` message here is necessary since a similar message is logged below in `setHealthStatus`

isvcs/container.go
```diff
@@ -879,7 +883,7 @@ func (svc *IService) setHealthStatus(result error, currentTime int64) {
 		"isvc": svc.Name,
 	})
 
-	if healthStatus, found := svc.healthStatuses[DEFAULT_HEALTHCHECK_NAME]; found {
+	if healthStatus, found := svc.healthStatuses[healthCheckIndex]; found {
 		if result == nil {
 			if healthStatus.Status != "passed" && healthStatus.Status != "unknown" {
 				log.Info("Internal service checked in healthy")
```
tgjonestx: This Info message needs to identify the healthcheck that passed, and probably should be reworded a bit since it's not the entire "Internal service" that's healthy - just one of possibly several checks that passed

isvcs/zk.go
```diff
@@ -61,82 +74,97 @@ func initZK() {
...
+	return func(halt <-chan struct{}) error {
+		for {
+			select {
+			case <-halt:
+				log.Debug("Stopped health checks for ZooKeeper")
+				return nil
+			default:
+				// Try ruok.
+
+				ruok, err := zkFourLetterWord(connectString, "ruok", time.Second*2)
```
tgjonestx: The connect time out needs to be longer I think

isvcs/zk.go
```diff
@@ -61,82 +74,97 @@ func initZK() {
...
+	return func(halt <-chan struct{}) error {
+
+		for {
+			select {
+			case <-halt:
+				log.Debug("Stopped health checks for ZooKeeper")
+				return nil
+			default:
+				// check the 'stat' command and see what the instance returns to the caller.
+				stat, err := zkFourLetterWord(connectString, "stat", time.Second*2)
```
tgjonestx: Ditto previous comment on connection timeout.

isvcs/zk.go
```diff
@@ -61,82 +74,97 @@ func initZK() {
...
-			if string(ruok) != "imok" {
-				log.WithFields(logrus.Fields{
-					"response": ruok,
-				}).Debug("Improper response to ruok from ZooKeeper")
-				healthy = false
-				time.Sleep(1 * time.Second)
-				continue
-			}
+// This function sets up a ZooKeeper health check function parameterized with the connection string eg: "127.0.0.1:2181".
+func SetZKHealthCheck(connectString string) HealthCheckFunction {
```
tgjonestx: The log messages in this method should identify the name of the healthcheck

isvcs/zk.go
```diff
@@ -61,82 +74,97 @@ func initZK() {
...
+	}
+}
 
-			if !healthy {
-				log.Info("ZooKeeper checked in healthy")
-			} else {
-				log.Debug("ZooKeeper checked in healthy")
+// This function sets up a ZooKeeper quorum check function parameterized with the connection string eg: "127.0.0.1:2181".
+func SetZKQuorumCheck(connectString string) HealthCheckFunction {
+	return func(halt <-chan struct{}) error {
```
tgjonestx: The log messages in this method should identify the name of the healthcheck

isvcs/container.go
```diff
@@ -769,41 +769,45 @@ func (svc *IService) startupHealthcheck() <-chan error {
...
-				})
-				if result == nil {
-					log.Info("Internal service checked in healthy")
-					break
-				} else if elapsed.Seconds() > svc.StartupTimeout.Seconds() {
-					log.WithFields(logrus.Fields{
-						"lastresult": result,
-					}).Warn("Unable to verify health of internal service")
+
+				for healthCheckIndex, healthCheckDefinition := range svc.HealthChecks {
```
zedieck: Change made as requested.

isvcs/container.go
zedieck: Change made as requested.

isvcs/container.go
```diff
@@ -769,41 +769,45 @@ func (svc *IService) startupHealthcheck() <-chan error {
...
 				}
+
 				log.Debug("Waiting for internal service to check in healthy")
+
 				time.Sleep(time.Second)
 			}
 			err <- result
 		} else {
-			svc.setHealthStatus(nil, time.Now().Unix())
+			svc.setHealthStatus(nil, time.Now().Unix(), DEFAULT_HEALTHCHECK_NAME)
```
zedieck: Refactoring completed.  Looks a whole lot better.

isvcs/container.go
```diff
@@ -831,43 +835,43 @@ func (svc *IService) runCheckOrTimeout(checkDefinition healthCheckDefinition) er
...
-	log := log.WithFields(logrus.Fields{
-		"isvc": svc.Name,
-	})
-
-	if checkDefinition, found = svc.HealthChecks[DEFAULT_HEALTHCHECK_NAME]; !found {
-		log.WithFields(logrus.Fields{
-			"healthcheck": DEFAULT_HEALTHCHECK_NAME,
-		}).Warn("Default health check not found")
-		return
+	// go through the health check s and find the longest duration interval for our ticker.
```
zedieck: :+1:  a little to happy with the space bar. :)

isvcs/container.go
```diff
@@ -831,43 +835,43 @@ func (svc *IService) runCheckOrTimeout(checkDefinition healthCheckDefinition) er
...
+	// CURRENT - we have one ticker for the longest of the periods for all health checks.
+	for _, checkDefinition := range svc.HealthChecks {
+		if tickInterval < checkDefinition.Interval {
```
zedieck: Extra logging added.

isvcs/container.go
```diff
@@ -831,43 +835,43 @@ func (svc *IService) runCheckOrTimeout(checkDefinition healthCheckDefinition) er
...
 		case <-halt:
 			log.Debug("Stopped health checks")
 			return
 
 		case currentTime := <-timer:
-			err := svc.runCheckOrTimeout(checkDefinition)
-			svc.setHealthStatus(err, currentTime.Unix())
-			if err != nil {
-				log.WithError(err).Warn("Health check failed")
+			for healthCheckIndex, checkDefinition := range svc.HealthChecks {
```
zedieck: Change completed as requested.

isvcs/container.go
```diff
@@ -831,43 +835,43 @@ func (svc *IService) runCheckOrTimeout(checkDefinition healthCheckDefinition) er
...
+				err := svc.runCheckOrTimeout(checkDefinition)
+				svc.setHealthStatus(err, currentTime.Unix(), healthCheckIndex)
+				if err != nil {
+					log.WithError(err).Warn("Health check failed")
```
zedieck: Change completed as requested.

isvcs/container.go
zedieck: Log message adjusted as requested.

isvcs/container.go
```diff
@@ -896,33 +900,34 @@ func (svc *IService) setHealthStatus(result error, currentTime int64) {
...
+		}).Warn("Health check not found")
 	}
 }
 
 func (svc *IService) setStoppedHealthStatus(stopResult error) {
 	if len(svc.healthStatuses) == 0 {
+		log.WithFields(logrus.Fields{
+			"isvc":        svc.Name,
+			"healthcheck": "None",
+		}).Warn("No health check not found")
```
zedieck: Grammatical error fixed here as well as two other places I found.

isvcs/zk.go
```diff
@@ -19,13 +19,26 @@ import (
...
 	"net"
 	"time"
+	"github.com/control-center/serviced/config"
+	"fmt"
 )
 
 var Zookeeper IServiceDefinition
 var zookeeper *IService
 
+const QUORUM_HEALTHCHECK_NAME = "Participating in Quorum"
```
zedieck: Change made as requested.  When shown in the UI, the "hasQuorum" format fits nicely with the existing "running" check.

isvcs/zk.go
zedieck: Completed as requested.

isvcs/zk.go
zedieck: Moved it back to 10 second time out AND put back in the retry logic.

isvcs/zk.go
zedieck: Completed as requested.

isvcs/zk.go
zedieck: Moved it back to 10 second time out AND put back in the retry logic.

isvcs/zk.go
```diff
@@ -105,30 +100,50 @@ func initZK() {
 // This function sets up a ZooKeeper health check function parameterized with the connection string eg: "127.0.0.1:2181".
 func SetZKHealthCheck(connectString string) HealthCheckFunction {
 	return func(halt <-chan struct{}) error {
+		healthy := true
+		times := -1
 		for {
+			if !healthy && times == 3 {
+				log.WithFields(logrus.Fields{
```
tgjonestx: There's an easier way to do some of this logging, if you want to.

Instead of repeating with the `WithFields` in each call, you can add the following before line 102:
```
    logger := log.WithField("healthcheck", DEFAULT_HEALTHCHECK_NAME)
```
Then here at lines 107-109, you can do something like:
```
        logger.Warn("Unable to validate health of ZooKeeper. Retrying silently")
```

And wash/rinse/repeat for other logging calls in this method

isvcs/zk.go
```diff
@@ -139,27 +154,49 @@ func SetZKHealthCheck(connectString string) HealthCheckFunction {
 // This function sets up a ZooKeeper quorum check function parameterized with the connection string eg: "127.0.0.1:2181".
 func SetZKQuorumCheck(connectString string) HealthCheckFunction {
 	return func(halt <-chan struct{}) error {
+		healthy := true
+		times := -1
 
 		for {
+			if !healthy && times == 3 {
+				log.WithFields(logrus.Fields{
```
tgjonestx: Ditto previous comment about reducing repetitive `WithFields` calls in all of the log messages in this method.

isvcs/container.go
```diff
@@ -883,10 +890,12 @@ func (svc *IService) setHealthStatus(result error, currentTime int64, healthChec
...
 	})
 
-	if healthStatus, found := svc.healthStatuses[healthCheckIndex]; found {
+	if healthStatus, found := svc.healthStatuses[healthCheckName]; found {
 		if result == nil {
 			if healthStatus.Status != "passed" && healthStatus.Status != "unknown" {
-				log.Info("Internal service checked in healthy")
+				log.WithFields(logrus.Fields{
+					"healthcheck": healthCheckName,
+				}).Info("Internal service health check is healthy")
```
zedieck: Fixed that up.

isvcs/zk.go
zedieck: Optimized this, looks a lot cleaner.

isvcs/zk.go
zedieck: Optimized this, looks a lot cleaner.

PR #3604, opened 2017-08-04
CC-3734 ERR Failed to publish events caused by: EOF" error message du…

romanbakaleyko:
…ring RM upgrade

### DIFF
--- a/isvcs/resources/logstash/filebeat.conf.in
+++ b/isvcs/resources/logstash/filebeat.conf.in
@@ -1,5 +1,5 @@
 filebeat:
-  idle_timeout: 5s
+  idle_timeout: 5m
   prospectors: ${PROSPECTORS_SECTION}
 
 path:


romanbakaleyko:
I noticed that the problem appears when during zenpack installation there is no updates to a log file during this 5s timeout. I found that we have a setting for a harvester to close inactive connections after 5m https://github.com/control-center/serviced/blob/develop/container/logstash.go#L72 that is why I chose 5m idle timeout but I am not very familiar with filebeat so probably this is a bad decision and would be better to decrease value for close_inactive setting. If anybody is familiar with filebeat please advice.

rphillipsz:
I don't know what a proper value is for that timeout.  I'm having trouble figuring out what version of filebeats we're using.. https://github.com/logstash-plugins/logstash-input-beats/issues/163 looks similar, and reverting to 3.1.10 (from 3.1.12) resolved the problem for them.  No idea if this is applicable though.. I'm having trouble finding people with context.

zedieck:
For context, here is the def for idle_timeout: https://www.elastic.co/guide/en/beats/filebeat/1.2/configuration-filebeat-options.html#_idle_timeout

Changing this to 5m means after 5m of idle inactivity, the spooler will be flushed even if the spool_size (default 2048) has not been reached.

zedieck:
Roman, does changing idle_timeout to 5m fix your issue?

romanbakaleyko:
>Roman, does changing idle_timeout to 5m fix your issue?
I did several checks and confirm that bumping to 5m resolves the issue.

>We need to determine if the error only happens when the container is being stopped.
No, errors are not related to a container stopping, they appear in the middle of zenpack installation process and appear when there is no output in logs from zenpack-manager for the a time which is bigger than idle_timeoout.

PR #3608, opened 2017-08-10
Implemented the backend for multiple instances for isvcs

zedieck:
https://jira.zenoss.com/browse/CC-3677

This set of changes implements the backend changes for supporting multiple instances of ISVCs which is part of the jira ticket.  We have UI changes that are coming in from another branch.

### DIFF
--- a/cli/cmd/cmd.go
+++ b/cli/cmd/cmd.go
@@ -206,6 +206,8 @@ func (c *ServicedCli) cmdInit(ctx *cli.Context) error {
 		fmt.Fprintf(os.Stderr, "Unable to set logging options: %s\n", err)
 	}
 
+	isvcs.InitAllIsvcs()
+
 	// TODO: Since isvcs options are only used by server (master/agent), these settings
 	//       should be moved to api.ValidateServerOptions
 	if err := setIsvcsEnv(ctx); err != nil {
--- a/cli/cmd/cmd.go
+++ b/cli/cmd/cmd.go
@@ -18,7 +18,6 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
-	"strings"
 	"syscall"
 
 	"github.com/Sirupsen/logrus"
@@ -26,7 +25,6 @@ import (
 	"github.com/control-center/serviced/cli/api"
 	"github.com/control-center/serviced/config"
 	"github.com/control-center/serviced/domain/service"
-	"github.com/control-center/serviced/isvcs"
 	"github.com/control-center/serviced/logging"
 	"github.com/control-center/serviced/servicedversion"
 	"github.com/control-center/serviced/utils"
@@ -206,14 +204,6 @@ func (c *ServicedCli) cmdInit(ctx *cli.Context) error {
 		fmt.Fprintf(os.Stderr, "Unable to set logging options: %s\n", err)
 	}
 
-	isvcs.InitAllIsvcs()
-
-	// TODO: Since isvcs options are only used by server (master/agent), these settings
-	//       should be moved to api.ValidateServerOptions
-	if err := setIsvcsEnv(ctx); err != nil {
-		fmt.Fprintf(os.Stderr, "Unable to set isvcs options: %s\n", err)
-		return err
-	}
 	return nil
 }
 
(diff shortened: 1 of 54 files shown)


cli/cmd/cmd.go
```diff
@@ -206,6 +206,8 @@ func (c *ServicedCli) cmdInit(ctx *cli.Context) error {
 		fmt.Fprintf(os.Stderr, "Unable to set logging options: %s\n", err)
 	}
 
+	isvcs.InitAllIsvcs()
```
tgjonestx: This should move because we don't want/need to initialize ISVCS when running `serviced` as a CLI tool - only when we're starting the actual server process.  This should be moved to `cli/api/daemon.go` - either the `startISVCS()` method or possibly the `run()` method before calling `startISVCS()`

Better yet - can it be called in `isvcs.Init()` instead?

isvcs/manager.go
```diff
@@ -157,9 +159,20 @@ func (m *Manager) GetHealthStatus(name string) (IServiceHealthResult, error) {
...
-		result.HealthStatuses = append(result.HealthStatuses, *value)
+
+	if instIndex == HEALTH_STATUS_INDEX_ALL {
+		for _, statusMap := range svc.healthStatuses {
+			for _, value := range statusMap {
+				result.HealthStatuses = append(result.HealthStatuses, *value)
+			}
+		}
+	} else {
+		for _, value := range svc.healthStatuses[instIndex] {
```
tgjonestx: Before using `instIndex`, there should be some kind of validation check to make sure that `instIndex` is in the valid range of `svc.healthStatuses` (i.e. not too big)

isvcs/es.go
```diff
@@ -77,9 +77,9 @@ func initElasticSearch() {
 		Timeout:     DEFAULT_HEALTHCHECK_TIMEOUT,
 	}
 
-	healthChecks := map[string]healthCheckDefinition{
-		DEFAULT_HEALTHCHECK_NAME: defaultHealthCheck,
-	}
+	healthChecks := make([]map[string]healthCheckDefinition, 1)
+	healthChecks[0] = make(map[string]healthCheckDefinition)
+	healthChecks[0][DEFAULT_HEALTHCHECK_NAME] = defaultHealthCheck
```
tgjonestx: Just FYI - if you wanted to use a static declaration, something like this would also work:
```
healthChecks := []map[string]healthCheckDefinition{
    map[string]healthCheckDefinition{
        DEFAULT_HEALTHCHECK_NAME: defaultHealthCheck,
    }
}
```
Which is actually 1 line longer, so potaytoe, potahtoe

isvcs/es_test.go
```diff
@@ -20,6 +20,7 @@ import (
 )
 
 func TestPurge(t *testing.T) {
+	InitAllIsvcs()
```
tgjonestx: Ditto comment about about possibly calling `InitAllIsvcs()` from within `Init()`

cli/cmd/cmd.go
zedieck: The issue with this request is If I do not call it after `config.LoadOptions(options)`, the configs are not loaded which is done one line up.  If I don't call it before `if err := setIsvcsEnv(ctx); err != nil {`, I get a panic because the objects have not been created.  Lets discuss first thing monday and I can demo for you.

zedieck: Ok I made the move your requested for this but it causes an unrecoverable error on start on both CLI and master service operations.
```
Unable to set isvcs options: Error setting internal service environment:'zookeeper:ZKID=1'  Service 'zookeeper' must be one of []'
Error setting internal service environment:'zookeeper:ZKID=1'  Service 'zookeeper' must be one of []'
```
Lets discuss.

isvcs/manager.go
```diff
@@ -166,13 +166,18 @@ func (m *Manager) GetHealthStatus(name string, instIndex int) (IServiceHealthRes
...
-	} else {
+	} else if instIndex < len(svc.healthStatuses) {
 		for _, value := range svc.healthStatuses[instIndex] {
 			result.HealthStatuses = append(result.HealthStatuses, *value)
 		}
+	} else {
+		log.WithFields(logrus.Fields{
+			"isvc": 		name,
+			"instIndex":	instIndex,
+		}).Warn("Instance index out of range")
```
tgjonestx: Should probably be an `Error()`

PR #3609, opened 2017-08-10
Support upgrades for application logfilters

tgjonestx:
There is a comment in the Jira with a link to the design doc. The basic stages of this PR are:
* Various refactors just to do general code cleanup
* Create a new Store in elastic for LogFilters
* On startup, copy LogFilters from any existing templates into the new Store
* On template-add, save the LogFilters in the new store
* Use the new LogFilter store to populate logstash.conf
* Add LogFilters as part of the ServiceMigrationRequest

The corresponding changes to the service migration SDK are here - https://github.com/control-center/service-migration/pull/66

### DIFF
--- a/facade/logstash.go
+++ b/facade/logstash.go
@@ -25,11 +25,12 @@ import (
 
 	"github.com/control-center/serviced/dao"
 	"github.com/control-center/serviced/datastore"
+	"github.com/control-center/serviced/domain/logfilter"
 	"github.com/control-center/serviced/domain/servicedefinition"
-	"github.com/control-center/serviced/domain/servicetemplate"
 	"github.com/control-center/serviced/domain/service"
 	"github.com/control-center/serviced/utils"
 	log "github.com/Sirupsen/logrus"
+	"github.com/docker/docker/pkg/version"
 )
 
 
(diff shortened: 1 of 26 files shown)


facade/logstash.go
```diff
@@ -181,51 +201,66 @@ func writeLogstashConfiguration(templates map[string]servicetemplate.ServiceTemp
...
+					"servicename": logInfo.Name,
 					"filter": filterName,
-				}).Warn("log filter not found")
+				}).Warn("service log filter not found")
 				continue
 			}
 			//  CC-3669: do not write duplicate filters for the same log file
 			if !utils.StringInSlice(config.Path, *logFiles) {
-				filters += fmt.Sprintf("\n  if [file] == \"%s\" {\n%s\n  }\n",
+				filterSection += fmt.Sprintf("\n  if [file] == \"%s\" {\n%s\n  }\n",
```
zedieck: Not sure if this is an issue or not but it looks like this is the older non regex logic that we fixed in develop to support rabbitmq logging.  (That could be in a different branch.)

tgjonestx: That will be sorted out when merging `bespin` back into `develop` (or rebasing `bespin` off of `develop`). Either way there will be some kind of file conflict that will have to be resolved

PR #3611, opened 2017-08-14
Add UI for multiple ZooKeeper instances

cchristiezen:
A grid was added to the internal ZooKeeper page which contains
information about the different ZooKeeper instances.  Health
checks will show if the ZooKeeper instances are running and if
they are in a quorum.  The grid will also show which instance is
the leader and the number of connections for each instance.


isvcs/zk.go
```diff
@@ -241,13 +307,70 @@ func BuildZookeeperInstance(instanceID int) dao.RunningService{
...
 
 	return instances
 }
+
+type ZooKeeperKey struct {
+	InstanceID int
+	Connection string
+}
+
+func GetZooKeeperKeys() []ZooKeeperKey {
```
tgjonestx: This more or less duplicates the logic at lines 38-49.  Please rework something so all of the logic in this file operates off a single list of ZK instances (or single function that builds a list)

isvcs/zk.go
```diff
@@ -232,7 +233,72 @@ func zkFourLetterWord(server, command string, timeout time.Duration) ([]byte, er
...
+		if key.InstanceID != instanceID {
+			continue
+		}
+
+		return GetZooKeeperServerStats(instanceID, key.Connection)
+	}
+	return ZooKeeperServerStats{}
+}
+
+func GetZooKeeperServerStats(instanceID int, connection string) ZooKeeperServerStats {
```
tgjonestx: We need to split out this logic somewhat like we do for health statuses.

In the current impl, a UI request to the CC backend turns around an issues a series of 4-letter commands to each ZK instance.   Instead, we need a go routine running in the background to continually sample the ZK stats, say every 30 seconds, and storing that info in some kind of internal map/list which the UI reads from.

isvcs/zk.go
```diff
@@ -232,7 +233,72 @@ func zkFourLetterWord(server, command string, timeout time.Duration) ([]byte, er
...
+	logger := log.
+		WithField("instanceid", instanceID).
+		WithField("connection", connection)
+
+	stats := ZooKeeperServerStats{InstanceID: instanceID}
+
+	bytes, err := zkFourLetterWord(connection, "srvr", 1*time.Second)
```
tgjonestx: A one second timeout is too short for a busy system. Please increase to 10 seconds like other 4-letter queries in this file.

isvcs/zk.go
```diff
@@ -232,7 +233,72 @@ func zkFourLetterWord(server, command string, timeout time.Duration) ([]byte, er
...
+
+		if strings.HasPrefix(line, "Mode:") {
+			tokens := strings.SplitN(line, ":", 2)
+			if len(tokens) < 2 {
+				logger.Error("Unable to get mode")
+			}
+			stats.Mode = strings.TrimSpace(tokens[1])
+			logger = logger.WithField("mode", stats.Mode)
+		}
+	}
```
tgjonestx: If `stats` is empty after the loop, we should log either a Warning or Debug message so we might notice if a collection interval was missed.  Let's do a warning for now.

isvcs/zk.go
```diff
@@ -232,7 +233,72 @@ func zkFourLetterWord(server, command string, timeout time.Duration) ([]byte, er
...
+func GetZooKeeperServerStats(instanceID int, connection string) ZooKeeperServerStats {
+	logger := log.
+		WithField("instanceid", instanceID).
+		WithField("connection", connection)
+
+	stats := ZooKeeperServerStats{InstanceID: instanceID}
+
+	bytes, err := zkFourLetterWord(connection, "srvr", 1*time.Second)
+	if err != nil {
+		logger.WithError(err).Error("Unable to get ZooKeeper stats")
```
tgjonestx: Loss of 1 ZK is not necessarily an error, so this should probably be logged as a Warning.

isvcs/zk.go
cchristiezen: Caching and saving to OpenTSDB has been added.

isvcs/zk.go
cchristiezen: Changed to 10.

isvcs/zk.go
cchristiezen: Changed to warning.

isvcs/zk.go
cchristiezen: Warning message added.

isvcs/zk.go
cchristiezen: Refactored so everything used the GetZooKeeperKeys method instead of the looping through the zkConnectString.

isvcs/container.go
```diff
@@ -625,6 +639,7 @@ func (svc *IService) run() {
 						haltStats <- struct{}{}
 						if len(svc.HealthChecks) > 0 {
 							haltHealthChecks <- struct{}{}
+							haltCustomStats <- struct{}{}
```
tgjonestx: I think this should be in a  different if block - something that matches the if block that starts the customStats:
```
if svc.customStats != nil {
    haltCustomStats <- struct{}{}
}
```

isvcs/container.go
```diff
@@ -673,6 +688,7 @@ func (svc *IService) run() {
 					haltStats <- struct{}{}
 					if len(svc.HealthChecks) > 0 {
 						haltHealthChecks <- struct{}{}
+						haltCustomStats <- struct{}{}
```
tgjonestx: ditto last comment

PR #3622, opened 2017-08-24
Add comments to serviced iptables rules so firewall mgmt scripts can ignore them.

ecifreo:
USAF reloads their firewall rules peridocially and needs a way to not interfere with the rules serviced creates.


ecifreo:
retest this please

PR #3627, opened 2017-09-01
Do not recreate dm device the base uuid and filesystem don't match

rphillipsz:
https://jira.zenoss.com/browse/CC-3751
When issuing serviced-storage check, if you forgot to include the devicemapper device (or specified the wrong device), your metadata got deleted.  This adds a few additional checks to ensure we aren't doing devicemapper.CreatePool() if there's an existing device.  I've also created custom types for the ThinpoolInitError error to allow the ThinpoolBaseDeviceVerificationFailedError to fall through so we can give a meaningful suggestion in the serviced-storage tool.

image

### DIFF
--- a/volume/devicemapper/devicemapper.go
+++ b/volume/devicemapper/devicemapper.go
@@ -778,9 +778,11 @@ func (d *DeviceMapperDriver) ensureInitialized() error {
 		dmoptions = append(dmoptions, "dm.fs=ext4")
 		deviceSet, err := devmapper.NewDeviceSet(poolPath, true, dmoptions, nil, nil)
 		if err != nil {
-			if _, thinError := err.(devmapper.ThinpoolInitError); thinError {
+			switch err.(type) {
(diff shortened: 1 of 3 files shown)


mayathegreat:
retest this

volume/devicemapper/devicemapper.go
```diff
@@ -76,6 +76,12 @@ func Init(root string, options []string) (volume.Driver, error) {
 	}
 
 	if err := driver.ensureInitialized(); err != nil {
+		switch err.(type) {
+		case devmapper.ThinpoolBaseDeviceVerificationFailedError:
+			// dm.thinpool argument was not provided but driver is still initialized
+			// so we will not error out
+			return driver, nil
```
jplouis: Seems like this case should fall through and call driver.cleanUpSnapshots() below.

mayathegreat: that's what we're trying to avoid with this change. We don't want the metadata to be modified  if the argument for dm.thinpool was mistakenly left out

PR #3629, opened 2017-09-06
CC-3870 Emergency shutdown replaces config files with originals

ssOlexBaiko:
[Jira](https://jira.zenoss.com/browse/CC-3870)
The fix adds additional calls to `fillServiceConfigs` for both emergency stoping / clearing emergency flag flows for preserving current configs.

### DIFF
--- a/facade/service.go
+++ b/facade/service.go
@@ -1498,6 +1498,10 @@ func (f *Facade) clearEmergencyStopFlag(ctx datastore.Context, tenantID, service
 
 	cleared := 0
 	for _, svc := range svcs {
+		// CC-3871 preserve config changes after clearEmergencyStopFlag call
+		if err = f.fillServiceConfigs(ctx, svc); err != nil {
+			glog.Errorf("%s", err)
+		}
 		svc.EmergencyShutdown = false
 		f.SetServicesCurrentState(ctx, service.SVCCSStopped, svc.ID)
 		err = f.updateService(ctx, tenantID, *svc, false, false)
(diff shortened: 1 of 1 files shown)


facade/service.go
```diff
@@ -1498,6 +1498,10 @@ func (f *Facade) clearEmergencyStopFlag(ctx datastore.Context, tenantID, service
 
 	cleared := 0
 	for _, svc := range svcs {
+		// CC-3871 preserve config changes after clearEmergencyStopFlag call
+		if err = f.fillServiceConfigs(ctx, svc); err != nil {
+			glog.Errorf("%s", err)
```
johnlhamilton: Use plog, not glog.

facade/service.go
```diff
@@ -2554,6 +2558,10 @@ func (f *Facade) GetServicesForScheduling(ctx datastore.Context, ids []string) [
 		if err != nil {
 			glog.Warningf("Could not get service with id %s: %s", id, err)
 		} else {
+			// CC-3871 preserve config changes after Emergency shutdown
+			if err := f.fillServiceConfigs(ctx, &svc); err != nil {
+				glog.Errorf("%s", err)
```
johnlhamilton: Please replace the 2 glogs in this method with plog.WithError(err)....

johnlhamilton:
A few more comments:

1. Is it necessary that we fill out configs?  What does the servicestore do if we just pass a nil value for ConfigFiles?  Does it preserve the old value, or replace them with nil?
2. If it IS necessary to fill out configs, should we just call GetService (which fills out both configs and address assignments).  If we don't, are we also going to end up removing address assignments?
3. Are there any other places where we call getService and then commit the service back to the service store (resulting in the same bug)

facade/service.go
ssOlexBaiko: ok

facade/service.go
ssOlexBaiko: good

ssOlexBaiko:
1. If we just pass a nil for ConfigFiles it will remove all configs from all affected services
2. GetService use `fillOutService` which triggers both `fillServiceAddr` and `fillServiceConfigs` methods. As long as emergency shutdown doesn't unintentionally clear the address assignments we can avoid to use `fillServiceAddr` and it is better for performance
3. I didn't see that workflow anywhere except this case

ssOlexBaiko:
retest this

PR #3637, opened 2017-09-21
Change 'cancel' logic so stop doesn't hang during backups.

ecifreo:
Tests are working, tested with emergency shutdown.


ecifreo:
retest this please

PR #3645, opened 2017-09-26
pull the latest host object into retry loop

ssOlexBaiko:


cli/api/daemon.go
```diff
@@ -843,7 +843,7 @@ func (d *daemon) startAgent() error {
 				err = masterClient.UpdateHost(updatedHost)
 				if err != nil {
 					log.WithError(err).Warn("Unable to update master with delegate host information. Retrying silently")
-					go func(masterClient *master.Client, updatedHost host.Host) {
+					go func(masterClient *master.Client, myHost *host.Host) {
```
johnlhamilton: We need to re-fetch 'myHost' from the master via masterClient.GetHost(myHostID) every time to avoid modifying a stale object.  see line 820 above.

PR #3649, opened 2017-10-03
Update serviced rpm deps with docker-ce-17.09.0

ecifreo:
There is a known issue with Centos 7.4 that we are trying to avoid.


tgjonestx:
Keep in mind that you'll have to update the serviced-iso builds based on these changes and then update zenoss-deploy to use the new ISO images.

PR #3664, opened 2018-01-15
zope workers will consume memory unbounded over time

ISPHOST:


ISPHOST:
Ticket: https://jira.zenoss.com/browse/ZEN-17361

web/ui/static/i18n/en_US.json
```diff
@@ -189,6 +189,7 @@
     "label_protocol": "Protocol",
     "label_public_endpoint": "Public Endpoint",
     "label_public_endpoints": "Public Endpoints",
+    "label_ram_threshold": "Auto restart if reached used % of memory",
```
jpeacock-zenoss: Please use the following text:
> Restart when memory usage exceeds % of requested RAM:

jpeacock-zenoss:
Since the new RAM threshold field is a percentage of the requested RAM field, customers will have to adjust the requested RAM value to avoid services being continually restarted.  Of course, I'm assuming that the 'RAM Requested' field in the UI maps to the `RAMCommitment` field in the service def structure.

web/ui/static/i18n/en_US.json
ISPHOST: Value 0 mean: don't restart.
You can find this if statement in file cli/api/daemon.go and line 1111
if getService.RAMThreshold == 0

cli/api/daemon.go
```diff
@@ -1088,6 +1088,63 @@ func (d *daemon) startAgent() error {
...
+							} else {
+								log.Info("Restarted instance by RAM threshold", insta.ServiceID, a)
+							}
+						}
+					}
+				}
+
+			}
+		}
+	}()
```
jpeacock-zenoss: This should only run on the master.  Several other goroutines in this function have checks for whether the running process is master or delegate and this should do the same.  Since this routine looks at all the services, it should only run in one instance, i.e. master.

cli/api/daemon.go
```diff
@@ -1088,6 +1088,65 @@ func (d *daemon) startAgent() error {
...
+							if int(percent) >= int(getService.RAMThreshold) {
+								a, err := strconv.Atoi(insta.InstanceID)
+								if err != nil {
+									log.Errorln(err)
+									continue
+								}
+								if err := d.facade.StopServiceInstance(d.dsContext, insta.ServiceID, a); err != nil {
+									log.Errorln(err)
+								} else {
+									log.Info("Restarted instance by RAM threshold", insta.ServiceID, a)
```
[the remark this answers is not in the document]
ISPHOST: Done, I've updated log message

ISPHOST:
RETEST

ISPHOST:
retest

ISPHOST:
retest this please

PR #3666, opened 2018-01-17
Update debian package build, fix broken deps.

ecifreo:
Also, have the debian package do systemd set up, just like the rpm one, since Ubuntu xenial supports systemd.


ecifreo:
Please retest this

PR #3675, opened 2018-03-05
Fixed CC-4060 switched to daily by default

kickingsawdust:
Changing default to daily https://jira.zenoss.com/browse/CC-4060


kickingsawdust:
I've rethought this and per discussion with support going to leave this to default as weekly

pjohnston:
Yep, as discussed we're gonna leave it as weekly default.

PR #3676, opened 2018-03-05
Adds "addme" method to serviced

ed-:
A delegate host can now add itself to a main host
with a new add-private / "addme" command, which also
registers the host using an agreed-upon keypair stored at
/opt/serviced/etc/common.key

### DIFF
--- a/cli/api/host.go
+++ b/cli/api/host.go
@@ -172,7 +172,7 @@ func (a *api) AddHost(config HostConfig) (*host.Host, []byte, error) {
 	}
 }
 
-// Adds a new host and uses a common key to register it.
+// Adds a new host and uses a common key to register it. Returns the host and the master's public key.
 func (a *api) AddHostPrivate(config HostConfig) (*host.Host, []byte, error) {
 	// if a nat is configured then we connect rpc to the nat, otherwise
 	// connect to the host address.
--- a/cli/api/host.go
+++ b/cli/api/host.go
@@ -23,7 +23,6 @@ import (
 	"github.com/control-center/serviced/rpc/agent"
 	"github.com/control-center/serviced/rpc/master"
 	"github.com/control-center/serviced/utils"
-	//"github.com/control-center/serviced/logging"
 )
 
 // HostConfig is the deserialized object from the command-line
@@ -44,12 +43,6 @@ type AuthHost struct {
 	host.Host
 	Authenticated bool
 }
-
-/*
-var (
-	log = logging.PackageLogger()
-)
-*/
 
 func getAuthInfo(client master.ClientInterface, hosts []host.Host) ([]AuthHost, error) {
 	hostIDs := []string{}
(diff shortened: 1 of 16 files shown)


cli/cmd/host.go
```diff
@@ -286,6 +317,76 @@ func (c *ServicedCli) cmdHostAdd(ctx *cli.Context) {
 	c.outputDelegateKey(host, nat, privateKey, keyfileName, registerHost)
 }
 
+// serviced host add-private HOST:PORT POOLID [--memory SIZE|%] [--nat-address HOST:PORT]
+func (c *ServicedCli) cmdHostAddPrivate(ctx *cli.Context) {
```
jplouis: For this method if there is an error, please exit with a non-zero code.

cli/api/host.go
```diff
@@ -172,6 +172,49 @@ func (a *api) AddHost(config HostConfig) (*host.Host, []byte, error) {
 	}
 }
 
+// Adds a new host and uses a common key to register it.
+func (a *api) AddHostPrivate(config HostConfig) (*host.Host, []byte, error) {
```
jplouis: Document the return values

cli/cmd/key.go
```diff
@@ -169,3 +171,8 @@ func (c *ServicedCli) outputDelegateKey(host *host.Host, nat utils.URL, keyData
 	}
 	fmt.Println(host.ID)
 }
+
+func (c *ServicedCli) outputCommonKey(host *host.Host, nat utils.URL, keyData []byte) {
```
jplouis: Document the method

facade/host.go
```diff
@@ -61,6 +61,21 @@ func (f *Facade) AddHost(ctx datastore.Context, entity *host.Host) ([]byte, erro
 	return key, alog.Error(err)
 }
 
+// AddHost registers a host with serviced. Returns the host's _public_ key.
+// Returns an error if host already exists or if the host's IP is a virtual IP.
+func (f *Facade) AddHostPrivate(ctx datastore.Context, entity *host.Host) ([]byte, error) {
+	defer ctx.Metrics().Stop(ctx.Metrics().Start("Facade.AddHostPrivate"))
+	alog := f.auditLogger.Message(ctx, "Adding Host").Action(audit.Add).Entity(entity)
```
jplouis: Make the audit log message unique "Adding Host with common key" or something like that

facade/host.go
```diff
@@ -115,6 +130,54 @@ func (f *Facade) addHost(ctx datastore.Context, entity *host.Host) ([]byte, erro
...
+	// verify that there are no virtual IPs with the given host IP(s)
+	for _, ip := range entity.IPs {
+		if exists, err := f.HasIP(ctx, pool.ID, ip.IPAddress); err != nil {
+			return nil, fmt.Errorf("error verifying ip %s exists: %v", ip.IPAddress, err)
+		} else if exists {
+			return nil, fmt.Errorf("pool already has a virtual ip %s", ip.IPAddress)
+		}
+	}
+
+	/*
```
jplouis: remove unused code

rpc/master/hosts_client.go
```diff
@@ -56,6 +56,15 @@ func (c *Client) AddHost(host host.Host) ([]byte, error) {
 	return response, nil
 }
 
+//AddHost adds and registers a host with an agreed-upon shared secret keypair
```
jplouis: Document return value

rpc/rpcutils/authcodec.go
```diff
@@ -28,7 +28,13 @@ import (
 
 var (
 	// RPC Calls that do not require authentication or who handle authentication separately:
-	NonAuthenticatingCalls = []string{"Master.AuthenticateHost", "Agent.BuildHost", "ControlCenterAgent.Ping"}
+	NonAuthenticatingCalls = []string{
+		"Master.AuthenticateHost",
+		"Agent.BuildHost",
+		"ControlCenterAgent.Ping",
+		"Agent.AddHostPrivate",
```
jplouis: Why do we need agent and master added to this list?

cli/cmd/host.go
ed-: Done via `os.Exit(1)` in the body

PR #3677, opened 2018-03-22
Zookeeper doesn't correctly detect that a docker container doesn't exist

ssOlexBaiko:
[Jira](https://jira.zenoss.com/browse/CC-3926)

### DIFF
--- a/zzk/service/hoststate.go
+++ b/zzk/service/hoststate.go
@@ -315,22 +315,6 @@ func (l *HostStateListener) setInstanceState(containerExit <-chan time.Time, ssd
 				logger.WithError(err).Error("Could not set state for started container")
 				return nil, nil, false
 			}
-
-			// CC-3926
-			if err := l.handler.FindContainer(serviceID, instanceID); err != nil {
-				logger.WithError(err).Error("Container not found")
-
-				if err := UpdateState(l.conn, req, func(s *State) bool {
-					s.ServiceState = *ssdat
-					if s.DesiredState == service.SVCRun {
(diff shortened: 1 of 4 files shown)


zzk/service/hoststate.go
```diff
@@ -219,8 +219,16 @@ func (l *HostStateListener) Spawn(cancel <-chan interface{}, stateID string) {
 			return
 		}
 
+		// set up a listener on container state
```
smousa: What does this comment mean?  It looks like you are just getting the container state.  You aren't listening to it.

zzk/service/hoststate.go
```diff
@@ -219,8 +219,16 @@ func (l *HostStateListener) Spawn(cancel <-chan interface{}, stateID string) {
...
+		cstate, err := GetState(l.conn, req)
```
smousa: Why are you getting the state here?  This whole function's purpose is to look up the state object?  And besides, the problem in the ticket explains that the stats.go code is finding a state object for which there is no docker container.  So essentially this block is a no-op

zzk/service/hoststate.go
```diff
@@ -332,7 +340,12 @@ func (l *HostStateListener) setInstanceState(containerExit <-chan time.Time, ssd
 			}
 
 			logger.Debug("Resumed paused container")
-		}
+		} else {
+			if cstate.Status != "started" && cstate.Status != "starting" {
+                                l.cleanUpContainers([]string{stateID}, true)
```
smousa: The behaviour described in the if block manages already handles this information

PR #3688, opened 2018-05-24
ZING-330: use Auth0 for CC in ZING

morr-zenoss:
Work to add Auth0 authentication to CC for ZING (ZING-330)
Added new config parameterss to serviced. These parameterss hold the information we need to use Auth0 to validate logins. A serviced endpoint that serves `static/globals.js` is used to pass those parameters to the frontend. It will populate the file from serviced. On the frontend, if those parameters are present and populated, the login page will go to auth0 for authentication. Otherwise, the legacy (pam-based) login will work. This is to keep things backward-compatible and avoid needing to branch serviced code for ZING. 
On the backend, I've added validation for the jws tokens that auth0 serves. Currently, we attempt that validation first, then fall through to our session-based tokens, and then to pam validation. I will submit a JIRA shortly to rework that logic - @jplouis  wants the backend to not do pam validation if auth0 is configured.

To test this, you'll need to add a couple of callbacks on the auth0 side for your machine. To do this, go to https://manage.auth0.com/#/applications and select the 'ZING CZ (CC)' application (or use this [direct link](https://manage.auth0.com/#/applications/xQF6jCIx6ZynvlvzT8ZWWrbOswcgCwH9/settings)). Once there, scroll down to 'Allowed Callback URLs' and add an entry (please add, and don't replace) for the machine you're testing on. The entries should look like the one(s) there, but replace the ip with the ip of your machine. Do the same in the 'Allowed Logout URLs' field. When you're done, scroll down and click 'Save Changes'.  To test auth0 login, add the following entries to your serviced config file:
```
SERVICED_AUTH0_DOMAIN=zenoss-dev.auth0.com
SERVICED_AUTH0_AUDIENCE=https://cc.zing.com
SERVICED_AUTH0_GROUP=All Zenoss Employees
SERVICED_AUTH0_SCOPE=openid profile read:messages
SERVICED_AUTH0_CLIENT_ID=xQF6jCIx6ZynvlvzT8ZWWrbOswcgCwH9
```
Restart serviced, and you should be able to log in with auth0 (assuming you're in the All Zenoss Employees group on gmail). To test legacy functionality, comment out  or delete the `SERVICED_AUTH0_*` entries and restart serviced.


morr-zenoss:
I'll work on failing tests after lunch...

web/ui/static/auth0callback.js
```diff
@@ -0,0 +1,24 @@
...
+   if (err) {
+       console.error("Unable to authenticate: " + err);
+       webAuth.authorize();
+   } else if (result && result.idToken && result.accessToken) {
+       window.sessionStorage.setItem("auth0AccessToken", result.accessToken);
+       window.sessionStorage.setItem("auth0IDToken", result.idToken);
+       console.log('window.location.origin = ' + window.location.origin);
+       window.location = window.location.origin;
+   }
+});
```
spopezen: probably want to clean up a bit of this logging

web/ui/src/common/authService.js
```diff
@@ -24,6 +24,10 @@
              */
             setLoggedIn: setLoggedIn,
 
+            auth0login: function () {
+                angularAuth0.authorize();
+            },
```
spopezen: I'm not sure if this automatically tries to reauth before trying to create a new auth0 session or not, but we need to do that

web/ui/package.json
```diff
@@ -33,5 +36,10 @@
     "merge-stream": "1.0.0",
     "run-sequence": "1.2.2",
     "which": "1.2.11"
+  },
+  "dependencies": {
+    "angular-auth0": "^3.0.0",
+    "auth0": "^2.9.1",
+    "auth0-js": "^9.4.2"
```
rDr4g0n: It would normally be correct to add these to `dependencies`, but we are effectively vendoring all of our deps, so they should not appear in this list.

web/ui/package.json
```diff
@@ -5,6 +5,9 @@
   "repository": "https://github.com/control-center/serviced",
   "private": true,
   "devDependencies": {
+    "angular-auth0": "^3.0.0",
+    "auth0": "^2.9.1",
+    "auth0-js": "^9.4.2",
```
rDr4g0n: If we were adding these deps to `package.json`, they would not belong in `devDependencies` because they are actually runtime deps, not build time, so they should be removed.

jplouis:
Were you able to deploy a multi CC cluster and was RM able to use the CC API?

PR #3699, opened 2018-07-18
update serviced to take either array of strings or string for audience

morr-zenoss:
https://jira.zenoss.com/browse/ZING-709

First part of ZING-709. We had to update the audience string checks in auth0.go. The tokens we had been working with (identity tokens, I believe) had audience specified as an array of strings, but the tokens coming from the api-key-server are single strings only.

This PR will build a version of serviced that will work with the ZING-709 changes in api-key-server and api-key-proxy. To complete the ticket, we'll need to add the api-key-proxy to serviced as an isvc.

### DIFF
--- a/auth/auth0.go
+++ b/auth/auth0.go
@@ -25,22 +25,24 @@ type jwtAuth0Claims struct {
 	Subject   string      `json:"sub,omitempty"`
 }
 
+// Parse the interface{} that populates Audience - we get string from some certs, and an array of strings (which parses as an array of interfaces) from others.
 func (t *jwtAuth0Claims) CheckAudience(expected string) bool {
-	if audstr, ok := t.Audience.(string); ok {
-		return audstr == expected
+	// String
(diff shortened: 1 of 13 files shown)


morr-zenoss:
retest please

morr-zenoss:
## Latest changes: 
* add isvc to run api-key-proxy
* Add healthchecks for isvc
* Add serviced config parameter `SERVICED_START_KEY_API_PROXY` to specify whether or not to run api key proxy isvc
* Add logic to enable/disable api key proxy isvc based on parameter

## Coordinating changes: 
The Isvc definition in isvcs/api_key_proxy.go uses a variable `API_KEY_PROXY_TAG` which is specified in [isvs/isvcs.go](https://github.com/control-center/serviced/blob/a5d366fd5ab780831a07878827e21280273f70d5/isvcs/isvc.go#L44). This variable specifies the docker tag of the image which contains the proxy. We'll need to update it when the api key proxy image is released with a version number.

PR #3712, opened 2018-10-01
CC-4105 enforce md5 auth for zookeeper quorum

ssOlexBaiko:
[Jira](https://jira.zenoss.com/browse/CC-4105)


isvcs/isvc.go
```diff
@@ -37,7 +37,7 @@ const (
 	IMAGE_REPO         = "zenoss/serviced-isvcs"
 	IMAGE_TAG          = "v61"
 	ZK_IMAGE_REPO      = "zenoss/isvcs-zookeeper"
-	ZK_IMAGE_TAG       = "v10"
+	ZK_IMAGE_TAG       = "develop"
```
smousa: Are you sure this is right?

ssOlexBaiko: PR with the fix is going into `develop` branch
https://github.com/control-center/isvcs-zookeeper/pull/10
As soon as new version of isvcs-zookeeper will be released `ZK_IMAGE_TAG` will be changed

smousa: I think you should merge the zookeeper change first and bump the version

PR #3715, opened 2018-12-31
CC-3929 Risks associated with browser credential storage highlighted by customer security team

Merteg:


web/ui/src/Login/login.html
```diff
@@ -1,5 +1,6 @@
 <div class="container login-container">
-  <form id="login" class="form-signin {{extra_class}}" ng-submit="login()" form-autofill-fix>
+  <form id="login" autocomplete="off" class="form-signin {{extra_class}}" ng-submit="login()" form-autofill-fix>
+    <input type="password" autocomplete="off" style="display:none;" disabled>
     <img class="loginLogo" src="static/img/theMcKrakenLogo.png" style="padding: 0 0 35px;">
     <input type="text" ng-model="username" class="form-control" placeholder="Username" autofocus required>
     <input type="password" ng-model="password" class="form-control" placeholder="Password" required>
```
jpeacock-zenoss: I see *two* password input fields.  Does making the first password field invisible and configured for no autocomplete effect the second password field?

Merteg: In new versions of browsers, the autocomplete attribute doesn't do too much ([forum](https://bugzilla.mozilla.org/show_bug.cgi?id=956906#c100)), but when we have an invisible password input before other inputs, the browser auto-fills it instead of main inputs. This fix doesn't switch off autocomplete in all cases, but it helps in some of them.

PR #3716, opened 2019-01-10
Feature/zing 1635 config edit

ed-:


cli/cmd/service_test.go
```diff
@@ -706,6 +715,97 @@ func ExampleServicedCLI_CmdServiceEdit_err() {
 	// service not found
 }
 
+func ExampleServicedCLI_CmdServiceConfigList_usage() {
+  InitServiceAPITest("serviced", "service", "config", "list")
+	// Output:
```
tgjonestx: tab inconsistency lines [720..732].

Actually, there are other tabbing issues on other lines below

ed-: This is what I get for switching away from vim. Addressing.

PR #3717, opened 2019-01-15
Feature/zing 1635 tune

ed-:


cli/cmd/cmd.go
```diff
@@ -164,6 +164,9 @@ func New(driver api.API, config utils.ConfigReader, logControl logging.LogContro
 		cli.StringFlag{"auth0-scope", defaultOps.Auth0Scope, "Scope to request in Auth0"},
 		cli.StringFlag{"keyproxy-json-server", defaultOps.KeyProxyJsonServer, "URL for API key server (cc auth token endpoint)"},
 		cli.StringFlag{"keyproxy-listen-port", defaultOps.KeyProxyListenPort, "Port for API key proxy to listen on"},
+		cli.IntFlag{Name: "instances", Usage: "Service instance count"},
```
tgjonestx: AFAIK, the flags in this section are global to all commands. We don't want these flags to be global to all.  I think what you added to `cli/cmd/service.go` should be sufficient

cli/cmd/service.go
```diff
@@ -1957,3 +1977,85 @@ func (c *ServicedCli) cmdServiceClearEmergency(ctx *cli.Context) {
...
+			return
+		}
+		if uint64(oldThreshold) != newThreshold {
+			service.RAMThreshold = uint(newThreshold)
+			modified = true
+		}
+	}
+
+	if modified {
+		jsonService, err := json.MarshalIndent(service, " ", "  ")
```
tgjonestx: This seems a little round-about to me. We already have a `Service` object, and then we turn it into JSON, to pass to `c.driver.UpdateService()` which turns the JSON back into a `Service` object  I think it would be cleaner to avoid the JSON back-and-forth.

You could add something like `UpdateServiceObj()` to `cli/api/service.go`, call it here and refactor to `UpdateService()` in `cli/api/service.go` to call it as well

ed-: Can do. I was going for low wake; I'll add the new function.

cli/cmd/cmd.go
ed-: Works for me.

PR #3720, opened 2019-01-22
Feature/zing 1635 tune

pjohnston:


cli/cmd/service.go
```diff
@@ -2140,6 +2140,11 @@ func (c *ServicedCli) cmdServiceTune(ctx *cli.Context) {
 		return
 	}
 
+	if ctx.IsSet("help") {
```
ed-: Can you verify that this is needed? Do we already handle `--help` elsewhere?

cli/cmd/service.go
```diff
@@ -2193,16 +2203,20 @@ func (c *ServicedCli) cmdServiceTune(ctx *cli.Context) {
 	}
 
 	if modified {
-		if service, err := c.driver.UpdateServiceObj(*service); err != nil {
```
ed-: UpdateServiceObj was created specifically to avoid needing to parse the service into JSON and then reparse it back into a service.

cli/cmd/service.go
pjohnston: Right you are, I'll remove it.

cli/cmd/service.go
pjohnston: Yah this was just git foo. I have removed the changes

PR #3724, opened 2019-01-30
CC-4190: Backups won't start if previous backup failed

olst:
closing this PR and opening a new one
https://github.com/control-center/serviced/pull/3734


facade/service.go
```diff
@@ -2056,7 +2056,19 @@ func (f *Facade) WaitService(ctx datastore.Context, dstate service.DesiredState,
 				return result.Err
 			}
 		case <-timeoutC:
-			return fmt.Errorf("timeout")
+			if dstate.String() == "pause" {
```
johnlhamilton: don't compare strings, just do if dstate == SVCPause

node/agent.go
```diff
@@ -211,13 +211,36 @@ func attachAndRun(dockerID, command string) error {
...
+	}()
+
+	select {
+		case result := <- done: {
+			if result.err != nil {
+				glog.Errorf("Could not pause container %s: %s", dockerID, result.err)
+				return result.err
+			}
+		}
+		// replace hardcoded timeout with f.dfs.Timeout() - 10 sec
```
johnlhamilton: You should pass in the timeout to the call to attachAndRun.

node/agent.go
```diff
@@ -211,13 +211,36 @@ func attachAndRun(dockerID, command string) error {
...
+		case <- time.After(time.Duration(290 * time.Second)): {
+			kill := "quiesce=$(ps aux | grep '.*quiesce.*sh.*' | grep -v grep | awk '{print $2}');" +
```
johnlhamilton: Since this is the generic attachAndRun that could be used for anything, you should use the passed-in command to build the kill command.

node/container.go
```diff
@@ -337,8 +337,10 @@ func (a *HostAgent) PauseContainer(serviceID string, instanceID int) error {
 	// pause the running container
 	a.setInstanceState(serviceID, instanceID, service.StatePausing)
 	if err := attachAndRun(ctrName, svc.Snapshot.Pause); err != nil {
-		logger.WithError(err).Debug("Could not pause running container")
-		return err
+		logger.WithError(err).Warn("Could not pause running container")
+		a.setInstanceState(serviceID, instanceID, service.StateRunning)
```
johnlhamilton: after the pause fails, before setting instance state back to running, try an attachAndRun with the unpause command (unquiesce?).

node/agent.go
```diff
@@ -59,6 +59,8 @@ import (
 const (
 	dockerEndpoint     = "unix:///var/run/docker.sock"
 	circularBufferSize = 1000
+	killQuiesce        = "quiesce=$(ps aux | grep '.*quiesce.*sh.*' | grep -v grep | awk '{print $2}');" +
```
johnlhamilton: I still don't like this, because:
1. It only works if the quiesce/pause command contains the string "quiesce".  The command is defined in the service definition, so it could be anything.
2. attachAndRun could be used for other commands besides pausing, so we shouldn't kill "quiesce" anytime ANY command times out.

Can we succesfully grep the process out of ps using the passed-in `command` string?  If not, we need to figure something else out, like getting the pid when we run the command.

facade/service.go
```diff
@@ -2056,6 +2056,18 @@ func (f *Facade) WaitService(ctx datastore.Context, dstate service.DesiredState,
 				return result.Err
 			}
 		case <-timeoutC:
+			if dstate == service.SVCPause {
```
johnlhamilton: How about we change the error message to "Error waiting for services" and go ahead and do it on any timeout (not just Pause)

node/container.go
```diff
@@ -336,9 +350,39 @@ func (a *HostAgent) PauseContainer(serviceID string, instanceID int) error {
...
+		output, err := attachAndRun(ctrName, cmd)
+		done <- out{string(output), err}
+	}()
+
+	select {
+		case result := <- done: {
+			if result.err != nil {
+				logger.WithError(result.err).Warn("Could not run command: %s; %s", svc.Snapshot.Pause, ctrName)
+				// block to trigger timeout in facade/service.go
+				select{}
```
johnlhamilton: Don't block, just return the error.  As long as we never call setInstanceState(...., service.StatePaused), the master will timeout waiting (which is what we want).  If you block forever here, we'll leak a goroutine.

johnlhamilton: Also, we should try the unpause script as well here, and return the state to Running as you do below.

node/container.go
```diff
@@ -336,9 +350,39 @@ func (a *HostAgent) PauseContainer(serviceID string, instanceID int) error {
...
+			}
+		}
+		case <- time.After(commandTimeout): {
+			logger.Warn("Could not pause running container. Command timeout: %s", svc.Snapshot.Pause)
+			// kill pausing process
+			if _, err := attachAndRun(ctrName, kill); err != nil {
+				logger.WithError(err).Warn("Error while killing pausing process.")
+				a.setInstanceState(serviceID, instanceID, service.StateRunning)
+				select{}
```
johnlhamilton: don't block, return a timeout error

node/container.go
```diff
@@ -336,9 +350,39 @@ func (a *HostAgent) PauseContainer(serviceID string, instanceID int) error {
...
+			}
+			// try running resume script
+			if _, err := attachAndRun(ctrName, svc.Snapshot.Resume); err != nil {
+				logger.WithError(err).Warn("Could not resume container")
+			}
+			a.setInstanceState(serviceID, instanceID, service.StateRunning)
+			select{}
```
johnlhamilton: don't block, return a timeout error

node/container.go
```diff
@@ -336,9 +350,39 @@ func (a *HostAgent) PauseContainer(serviceID string, instanceID int) error {
...
 	// pause the running container
 	a.setInstanceState(serviceID, instanceID, service.StatePausing)
-	if err := attachAndRun(ctrName, svc.Snapshot.Pause); err != nil {
-		logger.WithError(err).Debug("Could not pause running container")
-		return err
+
+	go func() {
+		cmd := ""
+		if svc.Snapshot.Pause != "" {
+			cmd = svc.Snapshot.Pause + fmt.Sprintf(" & echo $! > %s", pidFile)
```
johnlhamilton: Can you confirm that this command will still return an error if the underlying quiesce.sh returns a non-zero exit code?

zzk/service/hoststate.go
```diff
@@ -397,8 +397,10 @@ func (l *HostStateListener) setInstanceState(containerExit <-chan time.Time, ssd
 			// container is attached and not paused, so pause the container
 			if err := l.handler.PauseContainer(serviceID, instanceID); err != nil {
 				logger.WithError(err).Error("Could not pause container")
-				l.cleanUpContainers([]string{stateID}, true)
-				return nil, nil, false
+				l.handler.ResumeContainer(serviceID, instanceID)
+				select{}
```
olst: If I return error here, backup won't interrupt, so I'm not sure how to stop backup on errors.
Also, the main issue for me is that I can't stop/restart service service after pausing failure. Service is just can't be stopped.
It stuck in `Stopping` state when I hit Stop button.
I tried updating service state in zk here, but it doesn't help.
I don't know what to do to stop service after pausing failure, also I don't know how to interrupt backup on error from this place without select{} or sleep.

olst:
closing this PR and opening a new one
#3734

PR #3726, opened 2019-01-30
CC-4179 Replace link to local help with link to zsd-help.zenoss.com

Merteg:


pjohnston:
I noticed the integration tests failed with the "foo" image issue. I'd fixed that on the build server earlier so just kicked the integration test off again

PR #3728, opened 2019-01-31
CC-4183: Fixed serviced exit code on command line error

pjohnston:


pjohnston:
We definitely need to make a change, it's such a PITA, all of services and the Ops team find working with the command line awful without proper exit codes. I highly doubt many customers have written anything that would be effected, but safe to assume someone has. We could release note and make them review any command line scripting? Look forward to your thoughts. I assume StevePCs comment is valid, yes?

pjohnston:
I opened CC-4187 to document the concerns here about the way this has been implemented and ExitError etc. I also noted on CC-4183 that it should absolutely be release noted.

PR #3734, opened 2019-03-04
CC-4190: Backup fails due to service pause fail

olst:


node/container.go
```diff
@@ -304,6 +305,15 @@ func (a *HostAgent) ResumeContainer(serviceID string, instanceID int) error {
 
 // PauseContainer pauses a running container
 func (a *HostAgent) PauseContainer(serviceID string, instanceID int) error {
+	commandTimeout := time.Duration(
+		config.GetOptions().MaxDFSTimeout - 10) * time.Second
```
johnlhamilton: this timeout should be like 1s longer than the MaxDFSTimeout.  If it is shorter than the master's timeout, then it will timeout before the master, and will immediately retry the pause and have to timeout again before it can get back to the running state.  If instead, we let the master timeout first, the master will set the desired state to running before this one times out.  Then this one will timeout, see the new desired state, and immediately go back to running.

olst: All problems are due to quiesce script which is stuck. After backup fail it remains running. If we make timeout 1 sec greater than MaxDFSTimeout, then this kill will not work:
https://github.com/control-center/serviced/blob/896cc85d5ea2a632958bbfed5cdc3712b0c88eba/node/container.go#L367
When this kill works a new quiesce is run. After that we have master's timeout, but in the same time we have a new process of the quiesce script which should be killed here when we meet master's timeout:
https://github.com/control-center/serviced/blob/896cc85d5ea2a632958bbfed5cdc3712b0c88eba/facade/service.go#L2065

olst: We should kill quiesce process on master's timeout.

olst: Actually, we can refuse this kill:
https://github.com/control-center/serviced/blob/896cc85d5ea2a632958bbfed5cdc3712b0c88eba/node/container.go#L367
but this kill fixes all things:
https://github.com/control-center/serviced/blob/896cc85d5ea2a632958bbfed5cdc3712b0c88eba/facade/service.go#L2065

johnlhamilton: Did you test this on multi-host?  I ask because utils.AttachAndRun will run docker exec on the host, and it will only succeed if the container you are attaching to is running on that host (the master, in this case).  In most multi-host environments, the containers are all running on delegates.

Below is the pause-for-snapshot process, in the event of a timeout:
On the master (facade package):
--------------------
M1. Master sets desired state to "paused" in zookeeper
M2. Master begins waiting for current state to be "paused"
M3. Master times out
M4. Master changes desired state to "running" in zookeeper
M5. Master returns timeout error, and snapshot fails

On the delegate (node package):
----------------------
D1. Delegate sees desired state change to "paused"
D2. Delegate changes current state to "pausing"
D3. Delegate runs quiesce script and waits for it to return
At this point, without your fix, the delegate waits here forever, stuck in pausing.  With your fix:
D4. Delegate times out
D5. Delegate kills quiesce script
D6. Delegate checks zookeeper for desired state
D7. If desired state is "paused", GOTO D1.
D8. Desired state is "running"
D9. Delegate ensures container is running
D10. Delegate sets current state to "running"
D11. Delegate resumes watching desired state in zookeeper for changes

With your fix on the delegate side, the delegate will always eventually get unstuck.  How long that takes depends on the race between step D6 and step M4.  If D6 happens first, then we have to wait another full timeout before it gets unstuck.  If M4 happens first, then we will get unstuck after the first timeout.  By making the delegate's timeout longer, we increase the chance that the master will timeout first, and desired state will already be "running" when the delegate times out.

So, just to be clear: We should not attempt an AttachAndRun in WaitService (facade/service.go) because:
1. It won't work if the container isn't running on the master: AttachAndRun does NOT ssh to the host first.
2. It is unnecessary: the delegate will get unstuck eventually.  We can make it faster by trying to ensure the master times out first.
3. It is outside the intended purpose of WaitService: This is used any time we need to wait on a state transition change (including "run" and "stop").

olst: Multihost system, service is on delegate: https://youtu.be/kCnymOCZA1k
Multihost system, service is on master: https://youtu.be/8-TbS9DzVUY

facade/service.go
```diff
@@ -2056,7 +2056,16 @@ func (f *Facade) WaitService(ctx datastore.Context, dstate service.DesiredState,
 				return result.Err
 			}
 		case <-timeoutC:
-			return fmt.Errorf("timeout")
+				errMessage := "Error waiting for services"
```
johnlhamilton: "Timeout waiting for services" would be a little more descriptive

olst:
retest this

webhookbotuser:
Build finished.

PR #3737, opened 2019-04-08
CC-4208: Fix --ramThreshold option

pjohnston:


cli/cmd/service.go
```diff
@@ -2521,15 +2522,26 @@ func (c *ServicedCli) cmdServiceTune(ctx *cli.Context) {
...
+			fmt.Fprintln(os.Stderr, fmt.Errorf("ramThreshold '%s' does not end with %%", newThreshold))
+			c.exit(1)
+			return
+		}
+
+		percent := newThreshold[:len(newThreshold)-1]
+		val, err := strconv.ParseUint(percent, 10, 64)
 		if err != nil {
-			fmt.Fprintln(os.Stderr, err)
+			fmt.Fprintln(os.Stderr, fmt.Errorf("ramThreshold '%s' must be an integer", newThreshold))
```
StevePC: perhaps if the setting must be provided in a specific form (ie. integer with a % after) all error messages should model the expected input?

PR #3744, opened 2019-04-22
ZEN-31553 Hourly serviced snapshots creating absurd amount of images

ISPHOST:


dfs/delete.go
```diff
@@ -51,6 +51,10 @@ func (dfs *DistributedFilesystem) deleteImages(tenantID, label string) error {
 			glog.Errorf("Could not remove image %s for %s under label %s: %s", image.String(), tenantID, label, err)
 			return err
 		}
+		if err := dfs.docker.RemoveImage("localhost:5000/"+image.String()); err != nil {
```
tgjonestx: Instead of hardcoding localhost:500 you should reuse the value from configuration.

`options.DockerRegistry` in `cli/api/daemon.go` contains the configured value for the docker registry.

Change the interface `Registry` to have a new method like `GetAddress()` and change `RegistryListener` to implement that method, then you can call it from DistributedFilesystem

ISPHOST: done

PR #3747, opened 2019-05-16
ZEN-31701 Update MariaDB to more current version

romanbakaleyko:


script/eval.go
```diff
@@ -60,16 +60,39 @@ func evalSnapshot(r *runner, n node) error {
 func evalUSE(r *runner, n node) error {
 
 	imageName := n.args[0]
-	replaceImgs := make([]string, len(n.args)-1)
-	//if len(n.args) > 1 && n.args[1] != "" {
+	replaceImgs := make([]string, 0)
```
jplouis: I know this code is poorly commented, but add a comment to the effect of:

If "service" is in the SVC_USE command the image will be replaced only on the service (and its children) listed after the "service" keyword.  eg.  `SVC_USE [PREVIOUS_IMAGE_NAMES] service <SVC_PATH>`

script/eval.go
```diff
@@ -60,16 +60,39 @@ func evalSnapshot(r *runner, n node) error {
 func evalUSE(r *runner, n node) error {
```
jplouis: make sure to add unit test for the new usages of SVC_USE

romanbakaleyko: in the script package I don't see tests for eval.go, regarding functionality in parser.go and nodes.go I am not sure what else I can add there. For ServiceUse facade method I don't see tests as well.

script/eval.go
jplouis: Please add comments on the expected format of the updated command

script/eval.go
```diff
@@ -57,21 +57,56 @@ func evalSnapshot(r *runner, n node) error {
...
+
+	tenantID, found := r.env["TENANT_ID"]
+	if !found {
+		return fmt.Errorf("no service tenant id specified for %s", USE)
+	}
+
 	if len(n.args) > 1 {
-		replaceImgs = n.args[1:]
+		for i, arg := range n.args[1:] {
+			if arg == "service" {
```
jplouis: I forgot to save this comment yesterday.  What happens when the keyword "service" is in the wrong location?\
eg. SVC_USE <new image> service <image to replace>  Zenoss.resmgr/Infrastructure/mariadb-model

Would like a test case for that.

script/eval.go
jplouis: There are tests in parser_test.go. https://github.com/control-center/serviced/blob/develop/script/parser_test.go#L36

It uses a test file to make sure the parsing goes as expected. There are also negative test cases at https://github.com/control-center/serviced/blob/develop/script/parser_test.go#L119

script/eval.go
romanbakaleyko: updated parser,  added check whether we have image to replace in the command and check if we have path_to_service specified.

script/eval.go
romanbakaleyko: added tests

PR #3749, opened 2019-05-24
CC-4217: Services should rebalance on rolling restarts.

olst:


olst:
retest this

olst:
retest this

olst:
retest this

olst:
retest this

olst:
From the last two comments from Paul and Jason here:
https://jira.zenoss.com/browse/CC-4217?focusedCommentId=150684&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-150684.
If I'm not missing something, we decided to refuse from `--rebalance` flag;
make `serviced service restart <service> --rebalance` behavior as default;
add `--skip-children` flag to allow restarting (rebalancing) only the parent service.
For example, if I wanted to restart only zproxy, I'd do
`serviced service restart Zenoss.resmgr --skip-children`

For example, here how I tested Zope rebalancing on restart and zproxy restart:
https://youtu.be/faELkdZAcVY

StevePC:
The Jira (and comments) specifically state that rolling restarts should cause services to rebalance, not that rolling restarts shouldn't be the default.  It also states that the default should be rolling restarts that also rebalance services.  The second comment was simply a warning that --rebalance was overloaded and used for another purpose.
- adding in --skip-children is definitely required, as --rebalance should not have been used for that purpose.
- the default behaviour should be to perform rolling restarts
- rolling restarts should rebalance services

I can't think of a use case for specifically not rebalancing services, however they are restarted.

StevePC:
It would make most sense to have 2 flags. 
 --skip-children (as you've implemented)
and 
 --fast-restart (stops all instances, then starts them.  maybe --full-restart?)
and any restart of instances/services shoudl choose where they run based on the available scheduling methods

PR #3751, opened 2019-06-13
CC-4232 Control Center Zenpack needs better iservices monitoring for OpenTSDB and ElasticSearch

Merteg:
To CC added two new files: `esstats.go` and `tsdbstats.go` The code in those files query ElasticSearch and OpenTSDB API to get JVM stats, process it and pushes it to the OpenTSDB. Updated service definition for the `isvc-opentsdb`, `elasticsearch-serviced` and `elasticsearch-logstash`. Added graphs for garbage collection count, garbage collection time and threads count. For `isvc-opentsdb` we monitor `opentsdb` and `metric-consumer` application that we have in the `opentsdb` container.


dfs/docker/docker.go
```diff
@@ -52,6 +52,7 @@ func IsImageNotFound(err error) bool {
 			regexp.MustCompile("(?i)could not find image"),
 			regexp.MustCompile("(?i)no such id"),
 			regexp.MustCompile("(?i)no such image"),
+                        regexp.MustCompile("(?i)manifest for .* not found"),
```
jplouis: What is this for?

Merteg: After docker was upgraded on the Jenkins node, we have got failed test here: https://github.com/control-center/serviced/blob/b990200b46df1a3b8f9791271abe6f21a0409bbe/dfs/docker/docker_test.go#L325, because of new kind of error message that isn't handled by this function https://github.com/control-center/serviced/blob/b990200b46df1a3b8f9791271abe6f21a0409bbe/dfs/docker/docker.go#L44

PR #3781, opened 2019-11-11
CC-4290: Serviced-controller Fails to Bind to Ports During Container Startup.

Deer-WarLord:


Deer-WarLord:
The explanation of the issue is in [the comment](https://jira.zenoss.com/browse/CC-4290?focusedCommentId=161480&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-161480) to the Jira ticket.

auth/token.go
```diff
@@ -256,12 +259,7 @@ func expired() bool {
 }
 
 func updateToken(token string, expires time.Time, filename string) {
-	select {
-	case <-WaitForDelegateKeys(nil):
-	case <-time.After(1 * time.Second):
-		log.WithField("timeout", "1s").Error("No delegate keys were available to parse the token within the timeout")
-		return
-	}
+	<-WaitForDelegateKeys(nil)
```
Deer-WarLord: Because we're waiting for a non-empty token in the controller function, there's no need to break the function of awaiting the delegate's key by using time.After

auth/localkeys.go
```diff
@@ -308,12 +309,15 @@ func ClearKeys() {
 func WaitForDelegateKeys(cancel <-chan interface{}) <-chan struct{} {
 	ch := make(chan struct{})
 	go func() {
-		for delegateKeys.localPrivate == nil || delegateKeys.masterPublic == nil {
-			select {
-			case <-dKeyCond.Wait():
```
Deer-WarLord: NewChannelCond object is often used in localkeys.go and token.go files and for me, it's hard to say in what exact code place we're not releasing this object for being able to close the channel and started establishing the RPC connections in the controller. Because of that I just simplified the logic. Now the goroutine only checking the existing keys every second until they wouldn't be load and then close the channel.

PR #3783, opened 2019-11-22
Use bash when running services as a non-root ruser

jpeacock-zenoss:
When serviced launches a service using a non-root user (given by the RunAs field), explicitly use the bash login shell to execute the Startup command.

Helps fix ZEN-32408.

Using a bash login shell to run services (as user zenoss) provides an environment that most Zenoss applications expect to be running in.

Also fixes zeneventserver's behavior in a zendev environment (it wouldn't start).

### DIFF
--- a/commons/subprocess/instance.go
+++ b/commons/subprocess/instance.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"os"
 	"os/exec"
+	"strconv"
 	"sync"
 	"syscall"
 	"time"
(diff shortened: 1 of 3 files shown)


StevePC:
perhaps using /bin/sh -i or /bin/sh --login would be simpler, as that will load the profile and rc files when it executes.

jpeacock-zenoss:
please test this.

jpeacock-zenoss:
please retest this.

jpeacock-zenoss:
please test this.

PR #3806, opened 2020-07-21
CC-4315 Upgrade go-dockerclient to fix pulling images from gcr.io

DeKapito:


DeKapito:
Updated the go-dockerclient library to the following commit (fixes the problem with parsing credentials from docker config): https://github.com/fsouza/go-dockerclient/commit/f2ca7eb917de4f08359a9c099a50a3de941359cd

PR #3808, opened 2020-08-03
CC-4354: serviced README.md file is out of date

Deer-WarLord:


README.md
```diff
@@ -44,22 +56,22 @@ serviced host add HOST:PORT RESOURCE_POOL
 
 Dev Environment
 ---------------
-Serviced is written in go. To install go, download go v1.4 from http://golang.org.
-Untar the distribution to /usr/local/go. If you use a different location for go, you
+Serviced is written in go. To install go, download `go v1.14.4` from http://golang.org.
+Untar the distribution to `/usr/local/go`. If you use a different location for go, you
 must set GOROOT. See the http://www.golang.org for more information. Ensure that
-$GOROOT/bin is in you path.
+`$GOROOT/bin` is in you path.
```
jpeacock-zenoss, line 62: There is an alternative installation process on Ubuntu: the snap package.
```bash
$ sudo snap install --classic --channel=1.14/stable go
```

README.md
```diff
@@ -7,31 +7,43 @@ in a uniform way.
...
+3.  Elasticsearch 7.x uses a `mmapfs` directory by default to store its indices. 
+    The default operating system limits on mmap counts is likely to be too low, 
+    which may result in out of memory exceptions. 
+    On Linux, the limits should be increased by running the following command as root:
+    ```bash
+    sysctl -w vm.max_map_count=262144
+    ```
+   
+       To set this value permanently, update the `vm.max_map_count` setting in `/etc/sysctl.conf`. 
+       To verify after rebooting, run `sysctl vm.max_map_count.`
```
jpeacock-zenoss, line 36: Adding a file to `/etc/sysctl.d/` will also work.
```bash
$ sudo echo "vm.max_map_count=262144" > /etc/sysctl.d/10-vm.conf
```

Deer-WarLord: It's an official suggestion from the elastic site. I assume this more general approach. 
https://www.elastic.co/guide/en/elasticsearch/reference/current/vm-max-map-count.html

README.md
Deer-WarLord: Yes but we use not only ubuntu

README.md
jpeacock-zenoss: My suggestion could be more general.  The /etc/sysctl.d directory allows multiple users/packages from making changes to the same file.  The /etc/sysctl.d directory also supports priority.

Since what you wrote is copied from the elasticsearch website, I'll let it stand as is.

README.md
jpeacock-zenoss: Correct, but since our most common development environment is Ubuntu, I think it's worth mentioning that developers on Ubuntu can take advantage of the snap package.

Deer-WarLord: Agree, added

README.md
```diff
@@ -7,59 +7,75 @@ in a uniform way.
...
-1. Follow the instruction at http://www.docker.io/ , to install
+1.  Follow the instruction at http://www.docker.io/ , to install
    it on every host that serviced will run on. Ensure docker is running.
 
-2. Install a generated package
+2.  Install a generated package
+      ```bash 
+          sudo apt-key adv --keyserver keys.gnupg.net --recv-keys AA5A1AD7
+    ￼     REPO=http://get.zenoss.io/apt/ubuntu
+          sudo sh -c 'echo "deb [ arch=amd64 ] '${REPO}' trusty universe" > /etc/apt/sources.list.d/zenoss.list'
```
jpeacock-zenoss, line 20: There's an odd character on line 19.

Since `REPO` is used only once, I suggest replacing the `${REPO}` with `http://get.zenoss.io/apt/ubuntu` and remove the `REPO` variable entirely.

This shell snippet should be copy/paste-able to the command line and removing that variable would enable copy/pasting.

PR #3815, opened 2020-08-21
CC-4345: Apache Zookeeper nodes are accessible without ACL enabled.

Deer-WarLord:


coordinator/client/zookeeper/driver.go
```diff
@@ -131,8 +134,37 @@ func (driver *Driver) GetConnection(dsn, basePath string) (client.Connection, er
 			}
 		}
 	}()
+
+	options := config.GetOptions()
+	user := options.ZkAclUser
+	passwd := options.ZkAclPasswd
+	var acl []zklib.ACL
+
+	if user == "" || passwd == "" {
```
jplouis, line 143: The if/else has duplicated code. Could you simplify it to something like this?
```
	acl := zklib.WorldACL(zklib.PermAll)

  	if user == "" || passwd == "" {
  		user = os.Getenv("SERVICED_ZOOKEEPER_ACL_USER")
 		passwd = os.Getenv("SERVICED_ZOOKEEPER_ACL_PASSWD")
        }
        if user != "" && passwd != "" {
 		acl = zklib.DigestACL(zklib.PermAll, user, passwd)
 		if err := conn.AddAuth("digest", []byte(fmt.Sprintf("%s:%s", user, passwd))); err != nil {
 			plog.Errorf("AddAuth returned error %+v", err)
 			return nil, err
 		}
  	} 
```

Deer-WarLord: sure, fixed

PR #3832, opened 2021-03-12
CC-4419: 'serviced.rpm' upgrade to serviced version >= 1.8.0 fails with errors on delegate

Deer-WarLord:
Fixes CC-4419

All steps for export and import elasticsearch configuration
skipping for delegate hosts.


pkg/rpm/postupgrade
```diff
@@ -150,63 +150,81 @@ fi
 echo "Creating a backup of elasticsearch-serviced data in $HOST_ISVCS_DIR"
```
DeKapito, line 150: I assume this should be deleted because I see the same two lines in the next if-condition

DeKapito (changes requested):

DeKapito (approved):

PR #3899, opened 2023-07-17
CC-4534: Control Center Leaks Connections During Snapshot Deletion

Deer-WarLord:


Deer-WarLord:
Leakeg happens somewhere during the response, at the time we have an error we don't close the response body, and the connection is still alive with infinity timeout. The current approach will close the response body in any case that will prevent connection leaking. Also, to ensure that we have only one es client it was wrapped with a singletone.

vsaliieva (approved):

PR #3915, opened 2024-11-15
CC-4565: handle exception in case of bytes type instead of str

Deer-WarLord:


pkg/rpm/preupgrade
```diff
@@ -222,7 +225,7 @@ def calc_tpool_stats(stats):
     1073741824.0
     """
     data = {}
-    stats = [_f for _f in stats.split(' ') if _f]
+    stats = [_f for _f in stats.split(b' ') if _f]
```
jpeacock-zenoss: Is the `stats` parameter always going to be bytes?  If it's not, this will raise a TypeError.

jpeacock-zenoss (changes requested):

jpeacock-zenoss (approved):

PR #3916, opened 2024-11-19
CC-4567: Fix parse version flow for serviced and docker

Deer-WarLord:
python 3 
```
>>> version
b'Version:    1.11.0'
>>> version = [s for s in version.split(b' ') if not s == '']
>>> version
[b'Version:', b'', b'', b'', b'1.11.0']
```

python 2

```
>>> version = subprocess.check_output("serviced version | grep '^Version' | head -1 || true", shell=True).strip()
>>> version = [s for s in version.split(b' ') if not s == '']
>>> len(version) > 1
True
>>> version
['Version:', '1.11.0']
```

### DIFF
--- a/pkg/rpm/preupgrade
+++ b/pkg/rpm/preupgrade
@@ -346,7 +346,7 @@ def get_docker_version():
     version = subprocess.check_output("docker version | grep '^ Version' | head -1 || true", shell=True).strip()
     version = [s for s in version.split(b' ') if not s == '']
     if len(version) > 1:
-        version = version[1].strip()
+        version = version[-1].strip()
     else:
         version = version[0].strip()
     if len(version) == 0:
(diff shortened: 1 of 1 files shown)


pkg/rpm/preupgrade
```diff
@@ -346,15 +346,15 @@ def get_docker_version():
     version = subprocess.check_output("docker version | grep '^ Version' | head -1 || true", shell=True).strip()
     version = [s for s in version.split(b' ') if not s == '']
```
jpeacock-zenoss: This should be:
```
version = [s for s in version.split(b' ') if not s == b'']
```

jpeacock-zenoss (changes requested):

pkg/rpm/preupgrade
```diff
@@ -346,15 +346,15 @@ def get_docker_version():
...
     if len(version) > 1:
-        version = version[1].strip()
+        version = version[-1].strip()
     else:
         version = version[0].strip()
     if len(version) == 0:
         # older version of docker.
         version = subprocess.check_output("docker version | grep 'Server Version' | head -1 || true", shell=True).strip()
         version = [s for s in version.split(b' ') if not s == '']
```
jpeacock-zenoss: The `if not s ==` clause should be fixed here too.

jpeacock-zenoss (approved):

PR #3917, opened 2024-11-19
CC-4567: Fix write to file

Deer-WarLord:


pkg/rpm/preupgrade
```diff
@@ -399,7 +399,10 @@ def show_help():
 serviced_version = get_serviced_version()
 if serviced_version:
     with open(SERVICED_VERSION_FILE, 'w+') as file:
-        file.write(serviced_version)
+        try:
+            file.write(serviced_version)
+        except TypeError:
+            file.write(serviced_version.decode('utf-8'))
```
jpeacock-zenoss: I think you can fix this like so:
```
with open(SERVICED_VERSION_FILE, 'wb+') as file:
    file.write(serviced_version)
```
Note the `b` in the mode argument to `open`.

We already know that `get_serviced_version()` will return binary data.

I think there are a few other `open` statements in the scripts that could benefit from this.

jpeacock-zenoss (changes requested):

pkg/rpm/preupgrade
Deer-WarLord: No, python 2 returns string. 
```
>>> version = subprocess.check_output("serviced version | grep '^Version' | head -1 || true", shell=True).strip()
>>>
>>> version = version.rsplit(b' ', 1)[-1]
>>> version
'1.11.0'
```

jpeacock-zenoss: Bytes and strings are basically the same thing in Python 2.  However, Python 2.7 added support for binary syntax and modes so that the same code would behave the same in both Python 2.7 and 3.

jpeacock-zenoss (approved):

PR #3939, opened 2026-03-05
CC-4573: NAT IP address not available via serviced host list

Deer-WarLord:
# Fix NAT IP Configuration in AddHostPrivate & Add set-nat Command

## Summary
This PR fixes an issue where the NAT IP configuration was lost during the `AddHostPrivate` workflow and introduces a new CLI command `set-nat` to modify the NAT IP of existing hosts.

### 1. Fix in `AddHostPrivate` (`cli/api/host.go`)
**Issue:**
When adding a host via `AddHostPrivate` (used by `serviced host add-private`), the CLI makes a `BuildHost` RPC call to the agent to construct the `Host` object. Since the agent is unaware of the NAT configuration passed to the CLI, it returns a `Host` object with an empty `NatIP`. Consequently, the host was being registered with the master without the intended NAT IP.

**Fix:**
We now explicitly check if a NAT configuration is present in the CLI arguments. If so, we patch the `Host` object returned by the agent with the correct `NatIP` *before* registering it with the master.

```go
// cli/api/host.go

// ... after getting host 'h' from agent ...
if len(config.Nat.Host) > 0 {
    h.NatIP = config.Nat.Host
}
// ... proceed to register 'h' with master ...
```

### 2. New CLI Command: `set-nat`
A new command has been added to update the NAT IP for an existing host.

**Usage:**
```bash
serviced host set-nat HOSTID NATIP
```

**Example:**
```bash
serviced host set-nat host-1 192.168.100.20
```

### Unit Tests
A new unit test `TestAddHostPrivate_NatIP` has been added to `cli/api/host_test.go`.
- **What it tests:** It mocks the Agent and Master clients to verify that when `AddHostPrivate` is called with a NAT config, the `Host` object sent to the Master's `AddHostPrivate` method correctly contains the `NatIP`, ensuring the fix works as expected.

### Integration Tests
We have added comprehensive acceptance testing scenarios using mock agents to verify NAT IP functionality.

**Documentation:**
Please refer to **[acceptance/NAT_IP_TESTING.md](https://github.com/control-center/serviced/blob/fix/CC-4573/acceptance/NAT_IP_TESTING.md)** for detailed instructions on how to run these tests.

**Covered Scenarios:**
- **Scenario A:** Verifying hosts return pre-configured NAT IPs.
- **Scenario B:** Updating a host's NAT IP using `serviced host set-nat`.
- **Scenario C:** Adding a host without NAT, then setting it post-registration.
- **Scenario D:** Clearing a host's NAT IP.


StevePC (approved):

Deer-WarLord:
test this

